PR #5, opened 2026-05-13
build: organize frontend crates workspace

nv-anants:
Add a root Cargo workspace, shared dependency metadata, pinned Rust toolchain, CI validation, and dependency hygiene config. Move the demo lockfile to the workspace root and add the fastokens test fixture under tokenizers/tests so workspace tests pass without monorepo-local paths.

### DIFF
--- a/examples/dynamo-demo-server/src/handlers/completions.rs
+++ b/examples/dynamo-demo-server/src/handlers/completions.rs
@@ -8,7 +8,10 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use axum::Json;
 use axum::extract::State;
 use axum::http::StatusCode;
-use axum::response::{IntoResponse, Response, sse::{Event, KeepAlive, Sse}};
+use axum::response::{
+    IntoResponse, Response,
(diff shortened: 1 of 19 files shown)


examples/dynamo-demo-server/src/handlers/completions.rs
```diff
@@ -8,7 +8,10 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use axum::Json;
 use axum::extract::State;
 use axum::http::StatusCode;
-use axum::response::{IntoResponse, Response, sse::{Event, KeepAlive, Sse}};
+use axum::response::{
```
ishandhanani, line 14: Why these lints? Should we do this in dynamo as well?

nv-anants: is demo-server part of dynamo repo as well? i thought it was just in this one

ishandhanani: oh whoops. makes sense

PR #23, opened 2026-06-03
chore: bump dynamo-renderer to 1.3.0

nv-anants:
ref: OPS-6936


saturley-hall (approved):
Good catches in the releasing.md

PR #25, opened 2026-06-03
chore(conformance): vendor dynamo tests/parity fixtures into frontend-crates

keivenchang:
## Overview

Vendors the tool-calling + reasoning parity fixture corpus (209 + 24 YAML) from `ai-dynamo/dynamo`'s `tests/parity/` into a new top-level `conformance/parity/`, and teaches `sync-from-dynamo.sh` to keep it synced. Fixtures only — no runner, no parser.

## Details

- The parity fixtures live at dynamo's repo root (`tests/parity/`), outside `lib/`, so the existing per-crate sync never saw them. This is the missing piece for the crate to self-test against the same cases dynamo uses, without reaching into a dynamo checkout.
- Adds a parity block to `scripts/sync-from-dynamo.sh` that mirrors only the `fixtures/` subdirs — the `.py` harness, `parity_table.html.j2`, and the rendered `PARITY.html` stay out. `sync-check` now drift-checks the fixtures like code.
- Vendored from `ai-dynamo/dynamo@e3fb79e`; a follow-up dry-run reports zero drift.
- The Rust conformance harness that consumes these lands in a later PR.

Note: `sync-check` will show red from pre-existing crate drift (dynamo's `lib/parsers` is ahead of this repo, unrelated to this PR); the parity portion added here reports clean.

[DIS-2159](https://linear.app/nvidia/issue/DIS-2159)


keivenchang:
Superseded by #26. GitHub auto-closed this when the branch was renamed to carry DIS-2200; the same commits (rebased onto current main) are in #26.

PR #39, opened 2026-06-06
test(parsers-v2): regenerate TC stream-v2 fixtures for all families

keivenchang:
## Overview

Regenerates the tool-calling stream-v2 conformance fixtures for all 18 non-harmony families: per-chunk vLLM + SGLang streaming captures for every batch-derived case, replacing the original hand-chunked 4-case set.

- Each family's streamv2 cases now mirror its batch taxonomy (same sub-cases the batch tab shows).
- Marker-aware ~1-3 token chunking: special tokens (`<tool_call>`, `[TOOL_CALLS]`, `<|...|>`) stay intact, matching how a real tokenizer emits them — char-shredding a marker would make special-token parsers leak in ways the engine never would.
- Dynamo stream is marked TODO where parser v2 has no stream parser for the family yet; vLLM/SGLang per-chunk output is the target to match.
- Drops the old `TOOLCALLING.stream.*` source chunk-defs.

Generated by `conformance/utils/fill_streamv2.py` (lands in the stacked branch). YAML-only; the generator/viewer changes are in the child PR.

[DIS-2159](https://linear.app/nvidia/issue/DIS-2159)


rmccorm4 (approved):
Curious in future if we can represent the same kind of test coverage with significnatly less lines of YAML, but can be a future refactor.

keivenchang:
@rmccorm4 
You can definitely reduce the # of lines (via JSON) but no matter the format, they will all contain a very very dense and unique information of cases, expectations, and actual outputs. The amount of information (3 * P * C * L) will be the same, where 3 = Dynamo/vllm/sglang, P = #parsers, C = Cases/columns, L = lines/chunks.

PR #56, opened 2026-06-16
ci: pass --git-token to release-plz release

nv-anants:
The publish step calls the release-plz binary directly, but unlike release-plz/action the CLI does not read GITHUB_TOKEN from the env and needs the forge token via --git-token. The 'release' command builds a forge client unconditionally (even with git_release_enable=false), so the missing token aborted publishing with 'git release not configured. Did you specify git-token and forge?'. This regressed in #22 when the release step moved off release-plz/action, leaving crates published versions stuck (e.g. dynamo-parsers bumped to 1.4.0 but never released).


pvijayakrish:
lgtm!

PR #60, opened 2026-06-16
fix(parsers): stop mistral parser leaking [TOOL_CALLS] into content

KrishnanPrash:
#### Overview

The mistral tool-call parser leaked `[TOOL_CALLS]` / `[/TOOL_CALLS]` wrapper markers into user-visible `message.content` whenever the JSON body failed every parse path: malformed args inside a closed wrapper, or an orphan `[/TOOL_CALLS]` close with no opener. Tool-call markup must never reach `content`. The `mistral()` config never opted into `strip_markup_on_recovery` (the marker-stripping mechanism `nemotron_deci` already uses), so the recovery fall-through re-emitted the raw text verbatim. This enables it, plus a small mistral-scoped streaming-end-token extension, so the affected parity cells stop leaking.

#### Details

- `parsers/src/tool_calling/config.rs`: set `strip_markup_on_recovery: true` on `mistral()`.
- `parsers/src/tool_calling/json/base_json_parser.rs`: also run the marker-strip when the payload already carries a complete `[/TOOL_CALLS]` end token (not only on `allow_eof_recovery`), scoped to mistral via its `[TOOL_CALLS]` start token. This covers the streaming early-exit unjail case, which fires on a complete end marker but passes `allow_eof_recovery=false`. Scoped so `nemotron_deci` (the other `strip_markup_on_recovery` user) stays byte-identical.
- `conformance/toolcalling/fixtures/mistral/`: refresh `expected.dynamo` for the de-leaked batch cases (`4.a`/`4.b` to empty, `5.b` recovers `get_weather(NYC)`); classify the vLLM body-leak and SGLang no-space (`[TOOL_CALLS] [` vs `[TOOL_CALLS][`) divergences with a `reason:`. Stream fixtures (`stream.1`/`stream.4`) updated for the rendered table.

#### Before / After

| Case | Before (leaked) | After |
|---|---|---|
| `batch.4.a` | `calls=[] normal_text='[TOOL_CALLS]not a json array[/TOOL_CALLS]'` | `calls=[] normal_text=''` |
| `batch.4.b` | `calls=[] normal_text='[TOOL_CALLS][{"name":"get_weather","arguments":{"location":"NYC"][/TOOL_CALLS]'` | `calls=[] normal_text=''` |
| `batch.5.b` | `calls=[] normal_text='[{"name":"get_weather","arguments":{"location":"NYC"}}][/TOOL_CALLS]'` | `calls=[get_weather(NYC)] normal_text=''` |

#### Cross-impl

vLLM's mistral parser splits only on the `[TOOL_CALLS]` opener and has no `[/TOOL_CALLS]` concept, so on a malformed body it returns the raw post-opener text (leaking the close marker); with only an orphan close it treats the whole string as content. SGLang's `MistralDetector` requires `[TOOL_CALLS] [` with a space, but the canonical `mistral_common` wire format emits `[TOOL_CALLS][` with no space, so it returns no call on genuine Mistral output. Both are documented per-case.

The streaming jail that the end-token extension targets lives in dynamo's `lib/llm` and did not migrate to this crate; v1 conformance here is batch-only, so the batch leak rows are the validated surface (`parity_toolcalling` 606/606).

### DIFF
--- a/conformance/toolcalling/fixtures/mistral/TOOLCALLING.batch.4.yaml
+++ b/conformance/toolcalling/fixtures/mistral/TOOLCALLING.batch.4.yaml
@@ -17,16 +17,14 @@ cases:
           location:
             type: string
     expected:
-      dynamo:
-        reason: "Dynamo's mistral parser leaves [TOOL_CALLS], [/TOOL_CALLS] in normal_text on malformed input (impl-defined recovery)"
+      dynamo: &d_4_a
         calls: []
-        normal_text: '[TOOL_CALLS]not a json array[/TOOL_CALLS]'
+        normal_text: ''
       vllm:
         calls: []
         normal_text: not a json array[/TOOL_CALLS]
-      sglang:
-        calls: []
-        normal_text: ''
+        reason: "vLLM's mistral parser splits only on the [TOOL_CALLS] opener and has no [/TOOL_CALLS] concept; on a malformed body its pre-v11 path returns the raw post-opener text as content, leaking the [/TOOL_CALLS] close marker. Dynamo strips the wrapper markers on malformed recovery so nothing leaks (TOOLCALLING.batch.4 impl-defined recovery)."
+      sglang: *d_4_a
   TOOLCALLING.batch.4.b:
     description: Missing close brace in JSON
     ref: dynamo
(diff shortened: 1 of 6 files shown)


rmccorm4:
Confirmed the mistral row is green in generated report

rmccorm4 (approved):

keivenchang (approved):
Thank you.

PR #61, opened 2026-06-16
fix(parsers): stop qwen25 tool-call parser from leaking <tool_call> markup

KrishnanPrash:
#### Overview

The qwen25 tool-call parser leaked raw `<tool_call>` / `</tool_call>` markup into user-visible `message.content` on several malformed or partial inputs. Six cases on the qwen25 parity row were flagged as leaks. qwen25 shares Hermes' `<tool_call>...</tool_call>` wire format, but unlike Hermes it must never surface tool-call markup to the user. This change makes qwen25 discard or strip that markup so it never reaches content, and recover a no-argument call instead of dropping it. The behavior is opted into by qwen25 only; every other family that runs through the shared JSON parser is byte-for-byte unchanged.

#### Details

- `parsers/src/tool_calling/config.rs`: add `allow_name_only_call` and `discard_unparseable_wrapper` flags and a `qwen25()` config (Hermes wire format plus both opt-ins) instead of aliasing the hermes config.
- `parsers/src/tool_calling/json/base_json_parser.rs`: add a name-only call shape (`{"name":"f"}` with no `arguments`/`parameters` key), tried last and gated behind `allow_name_only_call`. When `discard_unparseable_wrapper` is set, a complete `<tool_call>...</tool_call>` wrapper that fails to parse is discarded instead of leaking the raw text, and an orphan `</tool_call>` end marker with no opener is stripped.
- `parsers/src/tool_calling/parsers.rs`: qwen25 maps to `ToolCallConfig::qwen25()` instead of `ToolCallConfig::hermes()`.
- `conformance/toolcalling/fixtures/qwen25/`: refresh `expected.dynamo` for the changed batch cases (`4.a/4.b/4.c` discarded, `5.b` marker stripped, `6.c` recovers `get_time({})`) and record SGLang outputs with a `reason:` on each documented divergence. Stream fixture (`stream.4`) updated for the rendered table.

#### Before / After

```
batch.4.a  <tool_call>not even json</tool_call>
  before: content leaks    after: content="" (discarded)
batch.5.b  {"name":...}</tool_call>   (orphan close)
  before: leaks marker      after: marker stripped
batch.6.c  <tool_call>{"name":"get_time"}</tool_call>
  before: calls=[], leak     after: calls=[get_time({})]
```

#### Cross-impl

vLLM is n/a because the parity harness has no qwen25-to-hermes mapping (vLLM serves Qwen2.5 tool calls through its hermes parser per its own docs; adding that mapping is a sensible follow-up). Each SGLang divergence carries a reason in the fixture (its `Qwen25Detector` keys on a `<tool_call>\n` prefix and so detects nothing on the compact wrapper), so the qwen25 row shows zero leaks and zero unclassified divergences.

Note: dynamo's original change also touched `lib/llm/.../jail.rs` for the streaming finalize / trailing-content path. That has no home in frontend-crates (the jail is not in this crate; v1 conformance is batch-only; v2 streaming lives in `parsers_v2/`). All validated batch leak fixes here come from the parser-crate changes; the jail/streaming portion is dynamo-`lib/llm`-only and is intentionally not part of this PR.

### DIFF
--- a/conformance/toolcalling/fixtures/qwen25/TOOLCALLING.batch.4.yaml
+++ b/conformance/toolcalling/fixtures/qwen25/TOOLCALLING.batch.4.yaml
@@ -17,12 +17,18 @@ cases:
           location:
             type: string
     expected:
-      dynamo: &id001
+      dynamo:
         calls: []
-        normal_text: <tool_call>not even json</tool_call>
+        normal_text: ''
       vllm:
         unavailable: vLLM has no parser for family='qwen25'
-      sglang: *id001
+      sglang:
+        reason: 'Malformed body in a complete <tool_call> wrapper: Dynamo discards
+          the unparseable jail region (no leak); SGLang''s Qwen25Detector keys on a
+          "<tool_call>\n" prefix, so on this compact wrapper it detects nothing and
+          returns the raw text. Non-parity by design.'
+        calls: []
+        normal_text: <tool_call>not even json</tool_call>
   TOOLCALLING.batch.4.b:
     description: Missing close brace in JSON body
     ref: dynamo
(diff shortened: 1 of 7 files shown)


conformance/toolcalling/fixtures/qwen25/TOOLCALLING.batch.4.yaml
```diff
@@ -32,23 +38,35 @@ cases:
     expected:
       vllm:
         unavailable: vLLM has no parser for family='qwen25'
```
rmccorm4, line 40: For future follow-up, vllm uses hermes parser for qwen2.5, so I still question whether we need 2 distinct parsers on dynamo side or not

rmccorm4:
Confirming qwen25 row went green in conformance table after this PR

rmccorm4 (approved):

rmccorm4 (approved):
Not urgent - but would look to follow up on this: https://github.com/ai-dynamo/frontend-crates/pull/61#discussion_r3424777376

PR #66, opened 2026-06-22
feat(parsers-v2): emit DSv4 DSML tool name eagerly in streaming

keivenchang:
#### Overview:

This PR makes Dynamo's DeepSeek-V4 DSML streaming parser emit the tool **name as soon as the invoke header arrives**, instead of holding the whole call back until it finishes. That matches how vLLM and SGLang already stream (name first, arguments after), so consumers see the tool name earlier and Dynamo's stream lines up with the other engines.

#### What changed (and why):

- One behavior change in `parsers_v2/src/tool_calling/dsml.rs`: when `` `<｜DSML｜invoke name="...">` `` closes, emit a `name`-only delta immediately; emit the `arguments` as a second delta when `` `</｜DSML｜invoke>` `` closes. The two deltas coalesce by `tool_index` back into the exact same call as before — only the on-the-wire timing/splitting changed.
- Why: name-first streaming is what SGLang/vLLM do, and emitting the name early lets callers show "calling `get_weather`…" without waiting for the arguments to finish.
- The 10 `deepseek_v4` stream fixtures were regenerated to record the new emit sequence. One edge case is now an intentional stream-vs-batch divergence (allowlisted in the parity test): if a call truncates after the header but before close, the eager `name` has already been emitted (with empty args) while the strict batch parser drops the incomplete call — `` `batch.5.c` `` / `` `batch.5.e` ``.

#### Before / after (from the conformance chart — `deepseek_v4` → `TOOLCALLING.streamv2.1`, happy path):

Per-chunk emit of the Dynamo Rust stream parser, with SGLang shown for reference (note "after" now mirrors SGLang):

| input chunk | SGLang | Dynamo **before** | Dynamo **after** |
|---|---|---|---|
| `` `<｜DSML｜invoke` `` | — | — | — |
| `` ` name="get_weather">` `` | `name=get_weather` | — (nothing yet) | `name=get_weather, arguments=""` |
| `` ` <｜DSML｜parameter …>` `` | — | — | — |
| `` `NYC</｜DSML｜parameter> </｜DSML｜invoke>` `` | `arguments={"location":"NYC"}` | `name=get_weather, arguments={"location":"NYC"}` | `arguments={"location":"NYC"}` |

Before: Dynamo emitted nothing until the call fully closed, then sent name + arguments in one delta. After: it emits the name on the header chunk (like SGLang), then the arguments on the closing chunk.

#### Verification:

`cargo test -p dynamo-parsers-v2 -p dynamo-conformance-fixtures-v2` -> 27 passed, 0 failed (includes the batch-via-stream parity test). Conformance chart regenerated.

`parsers_v2/src/tool_calling/dsml.rs`, then `conformance/toolcalling/fixtures-stream-v2/deepseek_v4/TOOLCALLING.streamv2.1.yaml`

Related: [DIS-2237](https://linear.app/nvidia/issue/DIS-2237) — DSML parser work. This is the eager-streaming refinement, distinct from the wrapper-strict change tracked there.

/coderabbit profile chill

### DIFF
--- a/conformance/tests/parity_toolcalling_batch_via_stream.rs
+++ b/conformance/tests/parity_toolcalling_batch_via_stream.rs
@@ -66,8 +66,24 @@ fn toolcalling_batch_via_stream_parity() {
     // Batch samples where the streaming parser deliberately differs from the
     // strict batch parser. Removing an entry asserts that stream and batch now
     // agree on that sample.
-    let known_divergences: std::collections::BTreeSet<&str> =
-        ["deepseek_v4:TOOLCALLING.batch.5.g"].into_iter().collect();
+    //
+    // The DSv4 stream parser emits the function `name` eagerly the moment the
+    // `<｜DSML｜invoke name="...">` header closes (SGLang-style streaming latency).
+    // When the invoke then truncates before `</｜DSML｜invoke>`, that name has
+    // already escaped with empty arguments, while the strict batch parser drops
+    // the incomplete call entirely — so stream and batch diverge by design:
+    //   5.c: invoke header closed, truncated mid-parameter -> stream emits
+    //        ("get_weather", "") ; batch emits nothing.
+    //   5.e: first call complete, second call's header closed then truncated
+    //        mid-arg -> stream emits the extra ("get_weather", "") ; batch drops it.
+    //   5.g: bare invoke after prose, recovered eagerly (pre-existing).
+    let known_divergences: std::collections::BTreeSet<&str> = [
+        "deepseek_v4:TOOLCALLING.batch.5.c",
+        "deepseek_v4:TOOLCALLING.batch.5.e",
+        "deepseek_v4:TOOLCALLING.batch.5.g",
+    ]
+    .into_iter()
+    .collect();
 
     let mut total = 0usize;
     let mut consistent = 0usize;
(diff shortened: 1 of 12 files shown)


KrishnanPrash (approved):
Nice find 🙌 

One question: From my understanding, our parity test coalesces the chunks with `coalesce_calls()` and verifies the output.

Should we have an e2e test in dynamo repo that proves that the real streaming consumer handles two frames (name+empty args and then args-only) in a way we expect?

keivenchang:
Thanks! Two parts:

1. At the v2 level (this repo) the raw frame sequence is already locked, not just the coalesced result: `parity_toolcalling_stream` asserts `expected.dynamo` per chunk on both the token and text paths (plus the assembled coalesce). So the parser emitting name-first then args-only (same `tool_index`) is covered frame-by-frame by the fixtures.

2. The gap you're pointing at is downstream of the parser: the v2 streaming consumer that maps `ToolCallDelta` frames into OpenAI SSE `delta.tool_calls[]` and forwards them to the client. That lives in dynamo, not here, so yes — a dynamo e2e is the right place to prove it. It can be GPU-free: drive canned DSML text through the v2 parser + that consumer and assert the SSE frames are exactly:

   ```
   frame 1: delta.tool_calls[0] = {index:0, name:"get_weather", arguments:""}
   frame 2: delta.tool_calls[0] = {index:0, arguments:"{\"location\":\"NYC\"}"}
   => OpenAI client aggregates to get_weather({"location":"NYC"})
   ```

Opened a dynamo follow-up for that, tied to the v2 streaming wiring: ai-dynamo/dynamo#10856

PR #69, opened 2026-06-22
feat: jamba never-leaks tool-call markup on malformed input

KrishnanPrash:
#### Overview:
AI21 Jamba reuses the generic JSON base parser but had opted into none of the never-leak recovery flags, so malformed or unterminated `<tool_calls>` input left the raw envelope in `normal_text`. In the Dynamo per-engine view this showed up as live leaks on batch cases 4.a, 4.b, 5.b and 5.c (and the streaming-jail equivalents 4.b/4.c). This change enables the same discard/strip recovery that hermes, qwen25 and nemotron_deci already use, so jamba never surfaces tool-call markup to the user, and classifies the remaining vLLM-only divergences with a `reason:` so the parity matrix has no unexplained `research` cells.

#### Details:
- `config.rs` `jamba()`: enable `discard_unparseable_wrapper` and `strip_markup_on_recovery`, matching the hermes/qwen25 and nemotron_deci never-leak configs.
- `batch.4` / `batch.5`: flip `expected.dynamo` for the four leak cases to a clean discard (4.a, 4.b, 5.c) or a recovered call (5.b, where the orphan close tag is stripped and the bare array parses), and document vLLM's leak with a `reason:`.
- `batch.4.c`/`4.d`, `batch.5.a`/`5.d`/`5.e`, `batch.6.c`: add a `reason:` for the cases where Dynamo already discards or recovers and only vLLM leaks.
- `batch.2.c`, `batch.8.a`/`8.c`: add a `reason:` for the cosmetic trailing-whitespace difference (Dynamo trims the space adjacent to the fence, vLLM keeps it; calls are identical).
- `batch.8.d`: document the multiple-disjoint-fence limitation (the framed parser spans from the first opener to the last closer, so narration between two fences breaks the JSON parse). Multi-fence support is a base-parser change tracked as a follow-up.
- `stream.1` through `stream.4`: add a `reason:` for the vLLM streaming detector artifacts (it drops arguments to `{}` and leaks partial wrapper chunks). `stream.4.b`/`4.c` assert the suppressed streaming-jail output (empty for the truncated body, close-markers stripped for the orphan case) aligned by ai-dynamo/dynamo#10847.

#### Before / After:
```
batch.4.a expected.dynamo
  before: {calls: [], normal_text: "<tool_calls>not a json array</tool_calls>"}
  after : {calls: [], normal_text: ""}
batch.4.b expected.dynamo
  before: {calls: [], normal_text: "<tool_calls>[{\"name\": \"get_weather\", \"arguments\": {\"location\": \"NYC\"]</tool_calls>"}
  after : {calls: [], normal_text: ""}
batch.5.b expected.dynamo
  before: {calls: [], normal_text: "[{\"name\": \"get_weather\", \"arguments\": {\"location\": \"NYC\"}}]</tool_calls>"}
  after : {calls: [get_weather(location=NYC)], normal_text: ""}
batch.5.c expected.dynamo
  before: {calls: [], normal_text: "<tool_calls>[{\"name\": \"get_weather\", \"arguments\": {\"loc"}
  after : {calls: [], normal_text: ""}
stream.4.b expected.dynamo
  before: {calls: [], normal_text: "<tool_calls>[{\"name\": \"get_weather\", \"arguments\": {\"loc"}
  after : {calls: [], normal_text: ""}
stream.4.c expected.dynamo
  before: {calls: [], normal_text: "[{...NYC}}]</tool_calls></tool_calls></tool_calls>"}
  after : {calls: [], normal_text: "[{\"name\": \"get_weather\", \"arguments\": {\"location\": \"NYC\"}}]"}
```

#### Cross-impl:
```
4.a    Dynamo: leak -> discard ("")            vLLM: still leaks (documented)
4.b    Dynamo: leak -> discard ("")            vLLM: still leaks (documented)
5.b    Dynamo: leak -> recover call            vLLM: still leaks (documented)
5.c    Dynamo: leak -> discard ("")            vLLM: still leaks (documented)
s4.b   Dynamo: leak -> suppressed ("")         vLLM: partial call, empty args (documented)
s4.c   Dynamo: leak -> close markers stripped  vLLM: still leaks markers (documented)
       SGLang: n/a (no jamba detector) for all cases
```

Note: the `stream.4.b`/`4.c` fixtures assert the streaming-jail behavior from ai-dynamo/dynamo#10847, so this PR should merge after that one. Jamba stream fixtures are declarative (not parser-executed here), so CI does not gate on the ordering, but the dashboard would briefly show green ahead of the jail fix otherwise.

### DIFF
--- a/conformance/toolcalling/fixtures/jamba/TOOLCALLING.batch.2.yaml
+++ b/conformance/toolcalling/fixtures/jamba/TOOLCALLING.batch.2.yaml
@@ -56,6 +56,7 @@ cases:
             timezone: EST
         normal_text: 'Both:'
       vllm:
+        reason: "Cosmetic whitespace only: Dynamo trims the single space between the leading narration and the <tool_calls> fence; vLLM preserves it. Calls are identical."
         calls:
         - name: get_weather
           arguments:
--- a/conformance/toolcalling/fixtures/jamba/TOOLCALLING.batch.4.yaml
+++ b/conformance/toolcalling/fixtures/jamba/TOOLCALLING.batch.4.yaml
@@ -17,11 +17,13 @@ cases:
           location:
             type: string
     expected:
-      dynamo: &id001
-        reason: "Dynamo's jamba parser leaves <tool_calls> envelope in normal_text on malformed input (impl-defined recovery)"
+      dynamo:
+        calls: []
+        normal_text: ''
+      vllm:
+        reason: "vLLM's jamba detector leaves the raw <tool_calls> envelope in normal_text when the wrapped payload is not a valid JSON array. Dynamo discards the unparseable wrapper (discard_unparseable_wrapper) so no tool-call markup reaches the user."
         calls: []
         normal_text: <tool_calls>not a json array</tool_calls>
-      vllm: *id001
       sglang:
         unavailable: SGLang has no detector for family='jamba'
   TOOLCALLING.batch.4.b:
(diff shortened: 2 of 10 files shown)


rmccorm4 (approved):
LGTM, thanks for the following up on these.

PR #72, opened 2026-06-23
fix(parsers): drop tool calls truncated mid-parameter-value

keivenchang:
#### Overview:

This PR drops tool calls truncated mid-parameter-value instead of emitting a guessed argument. It is the complete v1 fix (the shared XML-parser guard plus the batch fixtures that guard forces), ported from the now-closed dynamo#10733 after `lib/parsers` was cut over into the `dynamo-parsers` crate by dynamo#10423. Rebased onto current `main`.

#### Details:

- `parse_tool_call_block` now drops a recovered function whose trailing `<parameter=...>` value runs to raw end-of-input with no `</parameter>` / `</function>` / `</tool_call>` boundary. The lenient `$`-fallback recovery regex would otherwise capture the partial value and emit a guessed argument (e.g. `New York` for a truncated `New York City`). A value bounded by any close marker (only the wrapper missing) is still recovered.
- The guard is shared across all XML-config families, so it forces both `qwen3_coder` and `nemotron_nano` batch `5.c` (-> `[]`) and `5.e` (-> `[get_weather(Boston)]`) to be rebaselined together; the `5.e` `reason:` is updated (vLLM still over-recovers). minimax_m2 is strict-match, so the guard is inert there.
- Independent of #53 (the v2 streaming parser) — no file overlap. qwen3_coder's batch path is still v1, so its v1 batch fixture is rebaselined here, not in #53; the two PRs can land in any order.

#### Verification:

`dynamo-parsers` xml unit tests (114) and the conformance batch / batch-via-stream / stream tool-calling parity suites pass on the rebased base; `cargo fmt` clean.

`parsers/src/tool_calling/xml/parser.rs` (the truncation guard in `parse_tool_call_block`)

#### Related Issues:

Relates to DIS-2238. Supersedes the closed dynamo#10733.

/coderabbit profile chill

### DIFF
--- a/parsers/src/tool_calling/xml/parser.rs
+++ b/parsers/src/tool_calling/xml/parser.rs
@@ -363,6 +363,30 @@ fn parse_tool_call_block(
             continue;
         }
 
+        // Truncation guard: when a function block is recovered without its
+        // `</function>` close (EOF / max_tokens cut) and its trailing
+        // `<parameter=...>` value runs to raw end-of-input with no following
+        // close marker, the value was cut off mid-stream. The lenient recovery
+        // regex would capture the partial value via its `$` fallback, but an
+        // unterminated value can't be trusted (e.g. "New York" may be a truncated
+        // "New York City"), so drop the whole call rather than emit a guessed
+        // argument. A value bounded by *any* close marker — `</parameter>`,
+        // `</function>` (function terminated), or `</tool_call>` — is complete and
+        // still recovered.
+        let function_terminated = func_cap
+            .get(0)
+            .is_some_and(|m| m.as_str().contains(config.function_end_token.as_str()));
+        if !function_terminated {
+            if let Some(open_idx) = function_body.rfind(config.parameter_start_token.as_str()) {
+                let after_value = &function_body[open_idx..];
+                let value_bounded = after_value.contains(config.parameter_end_token.as_str())
+                    || after_value.contains(config.tool_call_end_token.as_str());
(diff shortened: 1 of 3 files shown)


parsers/src/tool_calling/xml/parser.rs
```diff
@@ -363,6 +363,30 @@ fn parse_tool_call_block(
...
+        // regex would capture the partial value via its `$` fallback, but an
+        // unterminated value can't be trusted (e.g. "New York" may be a truncated
+        // "New York City"), so drop the whole call rather than emit a guessed
+        // argument. A value bounded by *any* close marker — `</parameter>`,
+        // `</function>` (function terminated), or `</tool_call>` — is complete and
+        // still recovered.
+        let function_terminated = func_cap
+            .get(0)
+            .is_some_and(|m| m.as_str().contains(config.function_end_token.as_str()));
+        if !function_terminated {
```
rmccorm4: needs clippy fix for failing rust ci

rmccorm4 (approved):
Pre-approving, assuming https://github.com/ai-dynamo/frontend-crates/pull/72#discussion_r3457179113 will be addressed

parsers/src/tool_calling/xml/parser.rs
KrishnanPrash: Fixed clippy

keivenchang:
Thank you @KrishnanPrash for fixing clippy!

PR #73, opened 2026-06-23
test: add integration test for DSv4 two-frame streaming consumer

keivenchang:
Linear: [DIS-2237](https://linear.app/nvidia/issue/DIS-2237/dsml-require-outer-function-callstool-calls-wrapper-spec-strict)

#### Overview:

This PR adds an integration test proving a streaming consumer reconstructs the complete DeepSeek-V4 DSML tool call from the v2 parser's two-frame output (name-first delta, then args-only delta). Follow-up to the eager-streaming change in #66.

It is **not** an end-to-end test: it has no serving stack, HTTP boundary, or real client. It is a parser-plus-consumer integration/contract test, all in-process. The true HTTP e2e is tracked in ai-dynamo/dynamo#10856.

#### Relationship to existing fixtures:

The two-frame **emit shape** asserted here is already locked by `parity_toolcalling_stream` via `deepseek_v4/TOOLCALLING.streamv2.1` — this test reuses that exact happy-path chunking and the same `{name, ""}` then `{args}` frames, so that portion is a deliberate **duplication** of the fixture.

What is **new** and covered by no fixture test is the **consumer-side reassembly**. The fixtures stop at "here are the deltas the parser emits"; this test takes the next step — it maps those frames onto the real `dynamo-protocols` OpenAI wire types and merges them with a standard client-side aggregation (NOT the parser's `coalesce_calls()`) — proving the two frames reassemble into one complete, well-formed call with valid JSON arguments. That parser-output -> wire-type -> client-merge contract is the unique coverage this PR adds.

#### Details:

- Drives the real v2 `DeepSeekV4ToolStreamParser` over canned DSML chunks, asserts the two-frame shape, then maps the frames onto the real `dynamo-protocols` OpenAI wire types (`ChatCompletionMessageToolCallChunk`) and aggregates them the way an OpenAI client does — deliberately *not* the parser's `coalesce_calls()`, so it exercises the standard wire-protocol merge a downstream client performs. The "consumer" is the in-process delta-merge helper in the test file.
- Adds `dynamo-protocols` to the conformance crate's dev-dependencies.

What it asserts:

```
frame 1: delta.tool_calls[0] = {index:0, name:"get_weather", arguments:""}
frame 2: delta.tool_calls[0] = {index:0, arguments:"{\"location\":\"NYC\"}"}
=> aggregated to one call: get_weather({"location":"NYC"})
```

#### Verification:

`cargo test -p dynamo-conformance-fixtures-v2 --test integration_streaming_consumer_dsv4` -> 1 passed; full conformance + parsers_v2 suite green.

`conformance/tests/integration_streaming_consumer_dsv4.rs`

Related: partially addresses the spirit of ai-dynamo/dynamo#10856 — the full HTTP e2e there remains blocked on wiring the v2 parser into dynamo's streaming path. Builds on #66 (eager name-first streaming).

/coderabbit profile chill

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -860,6 +860,7 @@ version = "0.1.0"
 dependencies = [
  "dynamo-parsers",
  "dynamo-parsers-v2",
+ "dynamo-protocols",
  "serde",
  "serde_json",
  "serde_yaml",
--- a/conformance/Cargo.toml
+++ b/conformance/Cargo.toml
@@ -11,6 +11,7 @@ publish = false
 [dev-dependencies]
 dynamo-parsers = { workspace = true }
 dynamo-parsers-v2 = { path = "../parsers_v2", version = "0.1.10" }
+dynamo-protocols = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 serde_yaml = "0.9"
--- a/conformance/tests/integration_streaming_consumer_dsv4.rs
+++ b/conformance/tests/integration_streaming_consumer_dsv4.rs
@@ -1,0 +1,142 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Integration check that a streaming consumer reconstructs the complete
+//! DeepSeek-V4 DSML tool call from the v2 parser's two-frame output.
+//!
+//! The v2 stream parser emits one tool call as two OpenAI streaming deltas: a
+//! name-only delta the moment the `<｜DSML｜invoke name="...">` header closes,
+//! then an arguments-only delta when `</｜DSML｜invoke>` closes (SGLang/vLLM-style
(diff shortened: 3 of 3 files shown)


keivenchang:
Closing as redundant.

`parity_toolcalling_stream` already covers what this test asserts. Its step-3 "Assembled" check (`assemble_emitted`) drives the DSv4 v2 parser over the same `deepseek_v4/TOOLCALLING.streamv2.1` chunks, groups deltas by index, concatenates name/argument fragments, and JSON-parses the result — proving the two-frame (name-first, then args-only) output reassembles into `get_weather({"location":"NYC"})`.

The only thing this PR added on top was routing the frames through the `dynamo-protocols` OpenAI wire types to also assert first-frame id/type stamping. That wire-type mapping is too thin to warrant a standalone integration test, and the reassembly logic it exercises is a duplicate of the fixture harness. If the OpenAI wire-type round-trip needs explicit coverage later, it belongs as an assertion in the fixture harness, not a parallel test.

The real gap — a true HTTP e2e against dynamo's serving stack — remains tracked in ai-dynamo/dynamo#10856.

PR #79, opened 2026-06-23
fix(parsers-v2): drop DSv4 tool calls truncated mid-call to match v1 batch

keivenchang:
#### Overview:

This PR fixes the DSv4 streaming impl to conform to the correct batch behavior: tool calls truncated mid-call are dropped (matching the v1 batch parser) instead of emitted with empty arguments.

#### Details:

- `dsml.rs`: buffer each invoke until `</｜DSML｜invoke>` and emit `name` + `arguments` together; a call that truncates before its close is dropped (was: `name` emitted eagerly at the header, leaving `get_weather('')`).
- Regenerated deepseek_v4 streamv2 + batch-on-stream fixtures — 5.c -> `[]`, 5.e -> `[get_weather(Boston)]`, now matching `expected.dynamo`.
- Removed the now-stale `deepseek_v4:5.c`/`5.e` entries from the `parity_toolcalling_batch_via_stream` known-divergence allowlist (kept `5.g`, a separate bare-invoke recovery divergence).

#### Verification:

`validate_fixtures` clean; dsml unit tests + `parity_toolcalling_stream` + `parity_toolcalling_batch_via_stream` pass.

#### Related issue:

[DIS-2267](https://linear.app/nvidia/issue/DIS-2267)

`parsers_v2/src/tool_calling/dsml.rs`, then `conformance/tests/parity_toolcalling_batch_via_stream.rs`

/coderabbit profile chill

### DIFF
--- a/conformance/tests/parity_toolcalling_batch_via_stream.rs
+++ b/conformance/tests/parity_toolcalling_batch_via_stream.rs
@@ -67,23 +67,14 @@ fn toolcalling_batch_via_stream_parity() {
     // strict batch parser. Removing an entry asserts that stream and batch now
     // agree on that sample.
     //
-    // The DSv4 stream parser emits the function `name` eagerly the moment the
-    // `<｜DSML｜invoke name="...">` header closes (SGLang-style streaming latency).
-    // When the invoke then truncates before `</｜DSML｜invoke>`, that name has
-    // already escaped with empty arguments, while the strict batch parser drops
-    // the incomplete call entirely — so stream and batch diverge by design:
-    //   5.c: invoke header closed, truncated mid-parameter -> stream emits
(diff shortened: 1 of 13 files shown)


conformance/tests/parity_toolcalling_batch_via_stream.rs
rmccorm4, line 1: needs cargo fmt fix

rmccorm4 (approved):
LGTM other than cargo fmt failure

conformance/tests/parity_toolcalling_batch_via_stream.rs
keivenchang: fixed in `647c13b` — ran `cargo fmt --all` (I'd only fmt'd the `parsers_v2` package and missed the conformance test). `rust` check is green now.

PR #81, opened 2026-06-24
fix(parsers): preserve surrounding text where the spec is silent, drop where the reference decoder mandates a terminal block

keivenchang:
> [!WARNING]
> **PUNTING on the behavior change — this is not ready to land as a behavior change.**
>
> The drop-vs-preserve-surrounding-text change here only lands in the **batch** path. Shipping it for real requires updating **both batch and stream together**, which is out of scope for this PR:
>
> - All 48 fixture changes are `TOOLCALLING.batch.*` — no `.stream.*` fixtures are touched.
> - `content_is_prefix_only` is consumed only in the batch parsers (`base_json_parser.rs`, `dsml/parser.rs`, `xml/parser.rs`, `kimi_k2_parser.rs`), plus `deepseek_v4`'s v2 stream (`parsers_v2/tool_calling/dsml.rs`). For `kimi_k2`, `mistral`, and `nemotron_nano` the stream path is **not** updated — so batch would drop trailing/inter-call text while stream keeps it, and the same model would answer differently depending on `stream=true`/`false`.
> - Making the four terminal families consistent means wiring `content_is_prefix_only` through the v1 jail stream path (and v2 where applicable) and recapturing the `.stream.*` fixtures too. That is the follow-up; until batch and stream move together, the behavior change stays parked.

---

Linear: [DIS-2271](https://linear.app/nvidia/issue/DIS-2271/add-v2-streaming-tool-call-parsers-for-gemma-4-glm-kimi-k2-minimax-m2)

#### Examples — what changes for the user

When a model writes plain text around a tool call, do we keep it in the reply or drop it? It depends on what the model's own spec says.

**We DROP the surrounding text** (the model's spec says nothing belongs after a tool call):
- `deepseek_v4`, sandwich: `I will check the weather.` → tool call → `Let me know if you need more.` ⇒ we return `I will check the weather. ` and drop the trailing sentence.
- `kimi_k2`, two calls with narration between ⇒ we return only `First check Dallas.` and drop the wedged text.

**We KEEP the surrounding text** (spec is silent — dropping would throw away what the model wrote):
- `glm47`, same sandwich ⇒ we return `I will check the weather.  Let me know if you need more.`
- `gemma4`, two calls with narration between ⇒ we keep `I will check the weather.  Then check LA weather.`

Same cases on both sides: the model's spec decides, not the shape of the input.

#### Overview:

This PR makes v1 tool-call parsers keep natural-language text around a tool call when the model's spec is silent, and drop it for the four families whose published decoder defines the reply as only the text before the tool calls (`deepseek_v4`, `kimi_k2`, `mistral`, `nemotron_nano`).

#### Details:

- Adds a shared `content_is_prefix_only` flag (default off) across the json/xml/dsml/kimi parser configs; the four terminal families opt in. `nemotron_nano` split into its own config (it had been sharing `qwen3_coder`, which must keep preserving).
- The v2 `deepseek_v4` streaming parser latches suppression after the first block so streaming matches the batch behavior.
- Every narration case (`2.b/2.c`, `8.b/c/d`, batch + stream) carries a `dynamo_note` citing the per-model spec basis, shown in the conformance tooltip.

#### Verification:

`cargo test -p dynamo-conformance-fixtures-v2` (3 parity binaries), `dynamo-parsers` (624), `dynamo-parsers-v2` (32) all green; preserve-families show zero non-note value changes.

`parsers/src/tool_calling/config.rs`, then `parsers/src/tool_calling/xml/parser.rs`

/coderabbit profile chill

### DIFF
--- a/conformance/toolcalling/fixtures-batch-on-stream-v2/deepseek_v4/TOOLCALLING.batch.4.yaml
+++ b/conformance/toolcalling/fixtures-batch-on-stream-v2/deepseek_v4/TOOLCALLING.batch.4.yaml
@@ -30,7 +30,8 @@ cases:
     dynamo_rust:
       calls:
       - name: get_weather
-        arguments: {}
+        arguments:
+          location: NYC
       normal_text: ''
     vllm_rust:
       error: 'tool parser parsing failed: '
--- a/conformance/toolcalling/fixtures-stream-v2/deepseek_v3/TOOLCALLING.streamv2.2.yaml
+++ b/conformance/toolcalling/fixtures-stream-v2/deepseek_v3/TOOLCALLING.streamv2.2.yaml
@@ -194,6 +194,7 @@ cases:
         sglang_python: []
   TOOLCALLING.streamv2.2.c:
     description: 'With surrounding narration'
+    dynamo_note: 'DeepSeek-V3/V3.1/V3.2 publishes only the tool-call encoding grammar (a chat-template renderer, not a decoder) but is silent on natural-language text that follows the tool-call block. The model can emit trailing or inter-call prose, and nothing in the spec dictates whether a parser keeps or drops it; both readings are defensible. Because dropping silently discards genuine user-facing assistant text, Dynamo keeps it — it strips only the tool-call markup and preserves the surrounding narration verbatim. No published rule is violated, because the spec does not address this case. Source: https://huggingface.co/deepseek-ai/DeepSeek-V3.1/blob/main/README.md'
     ref: 'derived from TOOLCALLING.batch.2.c'
     tools:
     - name: get_weather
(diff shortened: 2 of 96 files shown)


rmccorm4:
> Recovery — argument value missing its </arg_value> close tag (glm47 4.d):

Do we give the user any kind of indication that we have modified the response stream in order to recover? This will help debug unexpected behavior of model emitting ABC but client receiving ABCD

rmccorm4:
May punt on this one until after code freeze based on offline discussion

keivenchang:
Closing this. It's 29 commits behind main and conflicting, built on the pre-[#95]/pre-DIS-2296 parser layout (v1 jail has since moved into dynamo-parsers and v1/v2/v2-py were regrouped under `parsers/`). The headline `content_is_prefix_only` terminal-block behavior is being punted — it only lands in batch and needs batch+stream updated together, or the same model answers differently by `stream=true/false` (see the warning note on the desc).

Not worth rebasing as one PR. The still-relevant residue (recovery fixes that never landed on main, plus the per-model terminal-block spec research) is being captured in a TODO for a fresh PR on current main.

PR #83, opened 2026-06-25
feat: Add MiniMax M3 tool-calling, reasoning, and conformance coverage

indrajit96:
## Description

This adds MiniMax M3 support to the frontend parser crates and conformance matrix.

MiniMax M3 uses a namespace-token XML-ish tool-call format:

```text
]<]minimax[>[<tool_call>
]<]minimax[>[<invoke name="get_weather">
]<]minimax[>[<location>NYC]<]minimax[>[</location>
]<]minimax[>[</invoke>
]<]minimax[>[</tool_call>
```

It also uses `<mm:think>...</mm:think>` reasoning markers, where the opener can
be prompt-prefilled by the chat template.

## Changes

- Added `minimax_m3` / `minimax-m3` tool-call parser registration.
- Added `MiniMaxM3ParserConfig` and `ParserConfig::MiniMaxM3`.
- Added a MiniMax M3 parser for:
  - namespace-prefixed tool-call blocks
  - `<invoke name="...">` calls
  - parameter tags where tag names are argument names
  - duplicate parameter tags as arrays
  - schema-based scalar coercion
  - nested MiniMax XML object/array payloads
  - EOF recovery for complete invokes with a missing outer close
- Added `minimax_m3` / `minimax-m3` reasoning parser registration.
- Added opt-in streaming dangling-end recovery for MiniMax M3 reasoning, so
  `reason</mm:think>answer` is parsed as reasoning plus answer when the opener
  was prompt-prefilled.
- Added MiniMax M3 conformance coverage:
  - tool-calling batch fixtures
  - tool-calling stream-v2 inventory fixtures
  - reasoning batch fixtures
  - reasoning stream fixtures
  - Top-N table/registry metadata
- Wired MiniMax M3 into the vLLM/SGLang conformance wrappers, with pinned peer
  engines marked unavailable where they do not yet expose M3 parsers.

## Notes

- TC stream-v2 coverage for MiniMax M3 is table/inventory coverage only. It marks
  `dynamo_rust`, `vllm_rust`, `vllm_python`, and `sglang_python` unavailable for
  now.
- This PR does not add a real `parsers_v2` MiniMax M3 streaming parser.

### Rust parser coverage

Targeted MiniMax M3 parser tests:

```bash
cargo test --locked -p dynamo-parsers minimax_m3 -- --nocapture
```

This covers:

- MiniMax M3 reasoning parser registration for both `minimax_m3` and
  `minimax-m3`.
- Batch reasoning extraction for visible `<mm:think>...</mm:think>` spans.
- Streaming reasoning extraction across split chunks.
- Prompt-prefilled reasoning opener behavior:
  `reason</mm:think>answer`.
- Dangling end-marker recovery in batch and streaming paths.
- Close-marker-only streaming case where the close marker is stripped and does
  not leak into normal text.
- MiniMax M3 tool-call parsing for the namespace-token XML format.
- EOF recovery for a complete `<invoke>` when the outer `tool_call` close is
  missing.
- Streaming jail start detection regression:
  normal text ending in `]` or `]<` does not trigger MiniMax M3 tool-call mode,
  while `]<]` and the full opener still do.

### Live E2E validation

In addition to the repo-local parser/conformance tests above, these
frontend-crates changes were built into the downstream Dynamo vLLM runtime image
and exercised in a live K8s deployment. Used `vllm/vllm-openai:minimax-m3` vllm base container.

### DIFF
--- a/conformance/toolcalling/fixtures/minimax_m3/TOOLCALLING.batch.2.yaml
+++ b/conformance/toolcalling/fixtures/minimax_m3/TOOLCALLING.batch.2.yaml
@@ -1,0 +1,115 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+family: minimax_m3
+model_label: MiniMax M3
+mode: batch
+cases:
+  TOOLCALLING.batch.2.a:
+    description: Two invokes inside one tool_call wrapper
+    ref: MiniMax M3 namespace-token XML tool-call grammar
+    model_text: |-
+      ]<]minimax[>[<tool_call>
+      ]<]minimax[>[<invoke name="get_weather">
+      ]<]minimax[>[<location>NYC]<]minimax[>[</location>
+      ]<]minimax[>[</invoke>
+      ]<]minimax[>[<invoke name="get_time">
+      ]<]minimax[>[<timezone>EST]<]minimax[>[</timezone>
+      ]<]minimax[>[</invoke>
+      ]<]minimax[>[</tool_call>
+    tools:
+    - &weather_tool
+      name: get_weather
+      parameters:
+        type: object
+        properties:
+          location:
+            type: string
+    - &time_tool
+      name: get_time
+      parameters:
+        type: object
+        properties:
+          timezone:
+            type: string
+    expected:
+      dynamo: &weather_time_calls
+        calls:
+        - name: get_weather
+          arguments:
+            location: NYC
+        - name: get_time
+          arguments:
+            timezone: EST
+        normal_text: ''
+      vllm: &m3_vllm_unavailable
+        unavailable: vLLM Python 0.23.0 in this conformance harness does not include the MiniMax M3 tool parser yet.
+      sglang: &m3_sglang_unavailable
+        unavailable: SGLang 0.5.12.post1 in this conformance harness does not include the MiniMax M3 detector yet.
+  TOOLCALLING.batch.2.b:
+    description: Two invokes adjacent inside one tool_call wrapper
+    ref: MiniMax M3 namespace-token XML tool-call grammar
+    model_text: |-
+      ]<]minimax[>[<tool_call>
+      ]<]minimax[>[<invoke name="get_weather">]<]minimax[>[<location>NYC]<]minimax[>[</location>]<]minimax[>[</invoke>]<]minimax[>[<invoke name="get_time">]<]minimax[>[<timezone>EST]<]minimax[>[</timezone>]<]minimax[>[</invoke>
+      ]<]minimax[>[</tool_call>
+    tools:
+    - *weather_tool
+    - *time_tool
+    expected:
+      dynamo: *weather_time_calls
+      vllm: *m3_vllm_unavailable
+      sglang: *m3_sglang_unavailable
+  TOOLCALLING.batch.2.c:
+    description: Parallel invokes with surrounding narration
+    ref: MiniMax M3 namespace-token XML tool-call grammar
+    model_text: |-
+      Both: ]<]minimax[>[<tool_call>
+      ]<]minimax[>[<invoke name="get_weather">
+      ]<]minimax[>[<location>NYC]<]minimax[>[</location>
+      ]<]minimax[>[</invoke>
+      ]<]minimax[>[<invoke name="get_time">
+      ]<]minimax[>[<timezone>EST]<]minimax[>[</timezone>
+      ]<]minimax[>[</invoke>
+      ]<]minimax[>[</tool_call> Done.
+    tools:
+    - *weather_tool
+    - *time_tool
+    expected:
+      dynamo:
+        calls:
+        - name: get_weather
+          arguments:
+            location: NYC
+        - name: get_time
+          arguments:
+            timezone: EST
+        normal_text: 'Both: '
+      vllm: *m3_vllm_unavailable
+      sglang: *m3_sglang_unavailable
+  TOOLCALLING.batch.2.d:
+    description: Same function name twice inside one tool_call wrapper
+    ref: MiniMax M3 namespace-token XML tool-call grammar
+    model_text: |-
+      ]<]minimax[>[<tool_call>
+      ]<]minimax[>[<invoke name="get_weather">
+      ]<]minimax[>[<location>NYC]<]minimax[>[</location>
+      ]<]minimax[>[</invoke>
+      ]<]minimax[>[<invoke name="get_weather">
+      ]<]minimax[>[<location>LA]<]minimax[>[</location>
+      ]<]minimax[>[</invoke>
+      ]<]minimax[>[</tool_call>
+    tools:
+    - *weather_tool
+    expected:
+      dynamo:
+        calls:
+        - name: get_weather
+          arguments:
+            location: NYC
+        - name: get_weather
+          arguments:
+            location: LA
+        normal_text: ''
+      vllm: *m3_vllm_unavailable
+      sglang: *m3_sglang_unavailable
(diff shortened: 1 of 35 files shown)


parsers/src/tool_calling/xml/minimax_m3_parser.rs
```diff
@@ -0,0 +1,605 @@
...
+    if raw.contains("]<]minimax[>[<") {
+        parse_nested_minimax_xml(raw, schema.cloned())
+    } else {
+        convert_scalar_value(raw, schema)
+    }
+}
+
+// Parses nested parameter bodies such as arrays of `<item>` objects into JSON values.
+fn parse_nested_minimax_xml(raw: &str, schema: Option<Value>) -> Value {
+    let chunks: Vec<&str> = raw.split("]<]minimax[>[").collect();
```
keivenchang: [Low -- DRY nit] These two hardcode the `]<]minimax[>[` literal, while the rest of the file derives every marker from `config.namespace_token` (see lines 77-97). It's the correct token so not a bug, but it's a divergent copy of a constant the config already holds: if `namespace_token` were ever set to a non-default, the outer parse would still work but this nested object/array path would silently fall back to scalar. Suggest routing both through `config.namespace_token`.

parsers/src/tool_calling/parsers.rs
```diff
@@ -52,6 +54,10 @@ pub fn get_tool_parser_map() -> &'static HashMap<&'static str, ToolCallConfig> {
         map.insert("qwen3_coder", ToolCallConfig::qwen3_coder());
         map.insert("jamba", ToolCallConfig::jamba());
         map.insert("minimax_m2", ToolCallConfig::minimax_m2());
+        map.insert("minimax_m3", ToolCallConfig::minimax_m3());
+        map.insert("minimax-m3", ToolCallConfig::minimax_m3());
+        map.insert("minimax_m3_nom", ToolCallConfig::minimax_m3());
+        map.insert("minimax-m3-nom", ToolCallConfig::minimax_m3());
```
keivenchang, line 63: [Question] Both aliases map to the identical `ToolCallConfig::minimax_m3()` -- functionally indistinguishable from `minimax_m3`. What does `_nom` denote, and why a separate alias if there's no behavioral difference? If it's a placeholder for a future variant, a comment would help.

conformance/toolcalling/fixtures/minimax_m3/TOOLCALLING.batch.yaml
```diff
@@ -0,0 +1,222 @@
...
+          location:
+            type: string
+    expected:
+      dynamo: &weather_call
+        calls:
+        - name: get_weather
+          arguments:
+            location: NYC
+        normal_text: ''
+      vllm:
```
keivenchang, line 32: [Note -- not blocking] vLLM and SGLang are `unavailable` across these fixtures, so the expected outputs are self-authored against the Dynamo parser. The wire *format* is now externally verified against the upstream chat template, but cross-engine behavioral parity is not -- the live E2E run is the only independent behavioral check. Worth calling out so it isn't read as cross-engine-confirmed.

conformance/toolcalling/fixtures/minimax_m3/TOOLCALLING.batch.yaml
```diff
@@ -0,0 +1,222 @@
...
+    - *weather_tool
+    expected:
+      dynamo:
+        calls: []
+        normal_text: Hello, how can I help you today?
+      vllm:
+        unavailable: vLLM Python 0.23.0 in this conformance harness does not include the MiniMax M3 tool parser yet.
+      sglang:
+        unavailable: SGLang 0.5.12.post1 in this conformance harness does not include the MiniMax M3 detector yet.
+  TOOLCALLING.batch.5.a:
```
keivenchang, line 49: [Coverage gap] Only `5.a` is shipped. I ran the other truncation shapes through the actual M3 parser -- all are well-formed M3 inputs with deterministic output, so they're addable (`NS` = `]<]minimax[>[`):

| Case | Verified Dynamo result |
|---|---|
| 5.b missing open (`NS<invoke ...>...NS</invoke>NS</tool_call>`) | batch + recover: **calls=[]**, whole text -> `normal_text` |
| 5.c trunc mid-value | batch + recover: **calls=[]**, `normal_text=""` |
| 5.d multi, no outer close | batch: **[]** / recover: **[a, b]** |
| 5.e multi, last trunc | batch: **[]** / recover: **[a]** (partial tail dropped) |
| 5.f bare then wrapped | recover: **[wrapped]** only; bare call **leaks into `normal_text`** |
| 5.g prose + bare + orphan close | recover: **calls=[]**, everything -> `normal_text` |

Notable: **5.b, 5.f, 5.g diverge from sibling `minimax_m2`** -- m2 recovers the inner invoke when the outer open is missing (it admits a bare body), but M3 hard-requires the `NS<tool_call>` fence, so it returns no call and leaks the text. That's a deliberate consequence of M3's syntax, but worth either adding 5.b-5.g as fixtures to pin it, or noting why M3 intentionally differs from m2 here. (Per the taxonomy, 5.b-5.e apply to any paired-fence parser; 5.f-5.g only to bare-body families, so for M3 those last two pin a divergence rather than a recovery.)

keivenchang (commented):
Reviewed the M3 work. Locally at `56447bd`: 628 parser tests pass (12 new), all 3 parity suites pass, clippy clean.

`]<]minimax[>[` is verified real -- MiniMax M3 namespace token 200058; checked the upstream `chat_template.jinja` (params encoded as tag names), so the `config.rs` M2-difference comment is accurate. The shared `BasicReasoningParser` opt-in `recover_dangling_end` is the right pattern -- extends the parent, no fork, no regression.

Pushed `14eb3bc`: the HTML highlighter was rendering `]<]minimax[>[` as a red orphan tag before every structural tag; now muted (`.tt-ns`) in static + streaming.

Inline notes below. Nothing blocking.

conformance/toolcalling/fixtures/minimax_m3/TOOLCALLING.batch.2.yaml
rmccorm4, line 1: It seems like M3 is added to `TC Batch` table:

Image

but not to `TC Stream` table:

Image

can we add it?

rmccorm4 (approved):
LGTM after open comments addressed

conformance/toolcalling/fixtures/minimax_m3/TOOLCALLING.batch.yaml
indrajit96: Thanks for running these!
Ran them locally and updated the table

conformance/toolcalling/fixtures/minimax_m3/TOOLCALLING.batch.2.yaml
indrajit96: Thanks done!

parsers/src/tool_calling/parsers.rs
indrajit96: Yes they are are supposed to be the same.
`_nom` here is only a deployment/backward-compatibility parser-name alias for configs that may already use that string. It was shared as part of Day0 atrifacts and launch commands.
I have updated the PR with a comment

parsers/src/tool_calling/xml/minimax_m3_parser.rs
indrajit96: Good catch. 
Fixed by threading `MiniMaxM3ParserConfig` into the nested-value parser path. `parse_parameter_value` now checks `parameter_start(config)`, and `parse_nested_minimax_xml` splits on `config.namespace_token` instead of hardcoding `]<]minimax[>[`.
I also added a regression test for this with a custom namespace

conformance/toolcalling/fixtures/minimax_m3/TOOLCALLING.batch.yaml
indrajit96: Done.
It's a note as when you hover over M3 in all the tables with text
```
Peer availability: no vLLM peer parser, no SGLang peer parser
```

indrajit96:
Merging, as all opens addressed.

PR #86, opened 2026-06-27
feat(conformance): add Dynamo v1 jail as a stream impl (TC v1 jail perspective) [WIP]

keivenchang:
Linear: [DIS-2279](https://linear.app/nvidia/issue/DIS-2279/add-v1-jail-streaming-conformance-tc-v1-tab-golden-fixtures-dynamo)

> **Draft / WIP.** Independent of #80 — the v1 jail and the streamv2 corpus both predate it.

#### Overview:

The v1 streaming **jail** (`lib/llm/.../jail.rs`, dynamo) is the current production streaming tool-call path, but no conformance surface exercises it — which let ai-dynamo/dynamo#10570 silently change recovery/no-args cases. This adds the v1 jail as a conformance impl with its own tab and a before/after view that surfaces those regressions.

#### What's here:

- **New tab: "TC v1 (stream data on stream-parser)"** — the v1 jail run over the v2 stream fixtures, sitting next to the TC v1 batch tab. Only the **Dynamo v1 Jail** parser is selectable here; the v2 stream tabs no longer list it.
- **"Compare with golden (#10570)" checkbox** — toggles the jail perspective's coloring between its normal **v1-vs-v2** divergence view (default) and a **golden-file** view (current vs the recorded golden). In golden mode every cell that ran is green except where the assembled #10570 output diverges from golden, which goes red — **8 regressions** (hermes/qwen25 `6.c`, mistral `5.b`, nemotron_nano/qwen3_coder `5.c`+`5.e`, pythonic `8.d`). URL-persisted via `?jaildiff=1`.
- **Two recorded golden sets in the fixtures** — `dynamo_jail` (golden = Dynamo main) and `dynamo_jail_after` (#10570), merged per-chunk into the 252 `streamv2` fixtures.
- **`merge_jail_perchunk.py`** — rewritten as a quote-aware **line inserter**: it groups physical lines into logical YAML lines (multi-line single-quoted `delta_text` scalars stay intact) and splices only the jail entries, so the diff is **added lines only** — no `yaml.safe_dump` reflow of the hand-formatted fixtures.
- **`capture_dynamo_jail.py`** — capture worker mirroring the vLLM-Rust source-path model: runs the in-crate `record_jail_streamv2` recorder from a dynamo checkout via `--dynamo-source` over the streamv2 corpus.
- **`fixtures.py` / `generate_conformance_table.py`** — assemble `dynamo_jail_after`, set a per-case `jail_regressed` (call-level diff vs golden), emit `data-jaildiff` + the golden-file CSS, and build the jail-only tab. Driven from `impls.py` (the `dynamo_jail` `ImplSpec`).

#### Validated:

Full gate green — 42 conformance pytests, Rust stream parity 121/121 (the new `dynamo_jail`/`dynamo_jail_after` keys deserialize cleanly), `validate_fixtures` 0 problems. Render confirms before = `green 611 / red 0`, after (golden mode) = `green 605 / red 6` (8 logical regressions; nemotron_nano shares the qwen3_coder row).

#### Remaining:

- Live `jail == golden` gate lands in **dynamo** (separate PR; `process_content` is crate-internal there).

/coderabbit profile chill


keivenchang:
Closing — superseded by the DIS-2296 jail migration (jail moved into frontend-crates via #92/#94/#101). This PR unique value was the golden before/after jail-regression tab, but its foundations are now stale: (1) its capture uses `capture_dynamo_jail.py --dynamo-source <dynamo>`, obsolete now that the native in-repo `capture_dynamo_jail_stream.py` + `record_dynamo_jail_stream.rs` are on main; (2) its whole framing is golden-vs-#10570, and #10570 was held pending DIS-2296, which has now landed; (3) 251 pre-migration fixtures/tooling. The jail-regression-guard idea is still worth having and stays tracked under DIS-2279 — cleaner to re-cut small on current main using the native recorder and a current baseline (reusing this PR tab/view code as reference) than to finish this. Branch preserved locally at backup/dis2279-pre-rebase-0710.

PR #91, opened 2026-06-30
ci: label external-contribution PRs and skip conformance comment on forks

nv-anants:
Add an external-contribution job to the Lint PR workflow that labels fork PRs on open.

Also guard the conformance-table comment step so it only runs for same-repo PRs; fork PRs receive a read-only token under pull_request, so the gh pr comment call was failing with 'Resource not accessible by integration'. The render and artifact-upload steps still run for forks.

closes: OPS-7444

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
           retention-days: 14
 
       - name: Comment on PR
-        if: github.event_name == 'pull_request'
+        if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name }}
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
--- a/.github/workflows/lint-pr-title.yaml
+++ b/.github/workflows/lint-pr-title.yaml
@@ -21,3 +21,23 @@ jobs:
         with:
           task_types: '["feat", "fix", "docs", "test", "ci", "refactor", "perf", "chore", "revert", "style", "build"]'
           add_label: 'true'
+
+  external-contribution:
+    name: Label external contributions
+    # Fork PRs only, on open.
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
rmccorm4, line 1: > Also guard the conformance-table comment step so it only runs for same-repo PRs; fork PRs receive a read-only token under pull_request, so the gh pr comment call was failing with 'Resource not accessible by integration'. The render and artifact-upload steps still run for forks.

What does this part mean exactly? The CI and the upload of the table can still run on an external fork contribution, but only the github comment linking to it will be unable to run?

rmccorm4 (approved):
Thanks Anant 🙏 One question on the github comment step

.github/workflows/ci.yml
nv-anants: yes, exactly. Fork PRs dont get write permissions to add comments, unless we use `pull_request_target` as trigger which should only be used for metadata actions like labeling (used in other lint pr workflow file)

saturley-hall: isn't the easier option here to use a `push` action on `pull-request/<id>` branches? That would utilize `copy-pr-bot` and allow commenting on from-non-canonical-repo-prs

saturley-hall: I mean you would still want to gate on only PRs and not pushes to main or workflow dispatches.

nv-anants: we dont have pull-request/* branches here yet, everything runs on public runners

nv-anants: CI is pretty light weight so prefer to keep things accessible for everyone and avoid adding copy-pr-bot (for now)

saturley-hall: Understood. While we do have copy-pr-bot enabled here that is not necessary yet

saturley-hall (approved):

PR #94, opened 2026-07-03
perf: Make tool-call jail completion incremental

jthomson04:
## Summary

- track tool-call jail completion incrementally instead of rescanning the full accumulated buffer for every streamed delta
- validate parsers only at lexical completion candidates, cache only successful results, and keep `find_tool_call_end_position` authoritative for emitted boundaries
- replace per-call wrapped-JSON regex construction with slice scanning while preserving parser output, recovery behavior, and public interfaces
- add focused regressions without modifying existing tests or parity fixtures

This ports the optimization from ai-dynamo/dynamo#10570 to the crate that now owns the jail after ai-dynamo/dynamo#11112. The previous path repeatedly scanned and parsed a growing buffer, producing quadratic behavior for large streamed tool arguments.

Markerless JSON completion is derived from parser configuration rather than parser-name allowlists. Invalid balanced candidates remain eligible for later revalidation, and parser results are reused only after successful validation.

## Validation

All of the following passed locally:

```text
cargo fmt --all -- --check
cargo check --workspace --all-targets --locked
cargo test --workspace --all-targets --locked
cargo test --workspace --doc --locked
cargo clippy --workspace --all-targets --all-features --locked -- -D warnings

cargo test --locked -p dynamo-parsers
cargo test --locked -p dynamo-conformance-fixtures-v2 --test parity_toolcalling
cargo test --locked -p dynamo-conformance-fixtures-v2 --test parity_toolcalling_batch_via_stream
cargo test --locked -p dynamo-conformance-fixtures-v2 --test parity_toolcalling_stream
```

The downstream Dynamo integration suite also passed against this local crate candidate: 127 tests passed across the six jail/tool-call targets, with one pre-existing ignored test. Dynamo formatting and all-targets clippy passed as well.

The new regression target covers invalid first balanced JSON, argument-less Hermes/Qwen calls, split Mistral close markers, markerless Llama/Phi calls, Kimi finalize recovery, overlapping markers, explicit null arguments, and UTF-8 boundaries.

## Benchmark

Hermes tool calls streamed in 16-byte chunks, median time:

| Payload | `dynamo-parsers` 3.1.0 | Candidate | Speedup |
| ---: | ---: | ---: | ---: |
| 128 B | 380 us | 7 us | 54.3x |
| 1 KiB | 1,707 us | 22 us | 77.6x |
| 8 KiB | 14,113 us | 135 us | 104.5x |
| 64 KiB | 227,887 us | 1,084 us | 210.2x |

### DIFF
--- a/parsers/v1/src/tool_calling/json/base_json_parser.rs
+++ b/parsers/v1/src/tool_calling/json/base_json_parser.rs
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::borrow::Cow;
+
 use serde_json::value::RawValue;
 use uuid::Uuid;
 
@@ -8,7 +10,9 @@ use super::super::ToolDefinition;
 use super::config::JsonParserConfig;
 use super::response::{CalledFunction, ToolCallResponse, ToolCallType};
 
-// Same as CalledFunction with named parameters.
+/// Public compatibility shape for callers that deserialize a tool call using
+/// the legacy `parameters` key. The parser itself uses an internal union so it
+/// can accept both argument-key variants in a single pass.
 //
 // `parameters` / `arguments` are deserialized as `Box<RawValue>` so the
 // original byte span — including key order, whitespace, and number
@@ -24,35 +28,20 @@ pub struct CalledFunctionParameters {
     pub parameters: Box<RawValue>,
 }
 
+/// Public compatibility shape for callers that deserialize a tool call using
+/// the `arguments` key. The parser itself uses an internal union.
 #[derive(Debug, serde::Deserialize)]
 pub struct CalledFunctionArguments {
     pub name: String,
     pub arguments: Box<RawValue>,
 }
 
-// A tool call may omit both `arguments` and `parameters` when the function
-// takes no inputs (e.g. `{"name": "get_time"}`). vLLM's and SGLang's Hermes
-// detectors treat that as an empty-argument call; this name-only shape is
-// tried as a last resort (after the `parameters` / `arguments` shapes) so the
-// call is recovered with `{}` args rather than dropped and its wrapper leaked.
+/// Public compatibility shape for a call containing only a function name.
+/// Parsers whose configuration permits name-only calls convert it to `{}`
+/// arguments through the internal union representation.
 #[derive(Debug, serde::Deserialize)]
 pub struct CalledFunctionNameOnly {
     pub name: String,
-}
-
-#[derive(Debug)]
-enum JsonPayload<'a> {
-    Borrowed(&'a str),
-    Owned(String),
-}
-
-impl JsonPayload<'_> {
-    fn as_str(&self) -> &str {
-        match self {
-            Self::Borrowed(value) => value,
-            Self::Owned(value) => value,
-        }
-    }
 }
 
 #[derive(Debug, serde::Deserialize)]
@@ -77,7 +66,7 @@ fn extract_tool_call_content<'a>(
     input: &'a str,
     start_token: &str,
     end_token: &str,
-) -> Option<JsonPayload<'a>> {
+) -> Option<Cow<'a, str>> {
     if start_token.is_empty() || end_token.is_empty() {
         return None;
     }
@@ -100,8 +89,8 @@ fn extract_tool_call_content<'a>(
 
     match matches.len() {
         0 => None,
-        1 => Some(JsonPayload::Borrowed(matches[0])),
-        _ => Some(JsonPayload::Owned(format!("[{}]", matches.join(",")))),
+        1 => Some(Cow::Borrowed(matches[0])),
+        _ => Some(Cow::Owned(format!("[{}]", matches.join(",")))),
     }
 }
 
@@ -112,11 +101,11 @@ fn extract_tool_call_content<'a>(
 fn extract_tool_call_content_eof_recovery<'a>(
     input: &'a str,
     start_token: &str,
-) -> Option<JsonPayload<'a>> {
+) -> Option<Cow<'a, str>> {
     let start_pos = input.find(start_token)?;
     let tail = input[start_pos + start_token.len()..].trim();
     if tail.starts_with('{') || tail.starts_with('[') {
-        Some(JsonPayload::Borrowed(tail))
+        Some(Cow::Borrowed(tail))
     } else {
         None
     }
@@ -353,10 +342,9 @@ fn try_parse_normal_text(input: &str, start_token: &str) -> String {
 /// let result = try_tool_call_parse_json(input)?;
 /// assert!(result.is_some());
 /// ```
-/// Parse `payload` into tool calls, trying the three canonical JSON shapes in
-/// order: an array of calls, a single `{name, arguments}`, then a single
-/// `{name, parameters}`. Within an array, each element is tried as
-/// `arguments` then `parameters`.
+/// Parse `payload` into tool calls through one internal representation that
+/// accepts both `arguments` and `parameters`. Arrays retain arguments-first
+/// precedence; single calls retain parameters-first precedence.
 ///
 /// Returns:
 /// - `Ok(Some(calls))` when `payload` matched one of the shapes. The vec may be
@@ -455,7 +443,7 @@ pub fn try_tool_call_parse_basic_json(
 
     // Iterate over all start and end tokens and try to extract the content between them
     // Assumption : One message will not contain different tags for tool calls. Iteration over tags is to support different tags by default for multiple models
-    let mut json = JsonPayload::Borrowed(trimmed);
+    let mut json = Cow::Borrowed(trimmed);
     let mut normal_text = trimmed.to_string();
     let mut found_start_token_with_no_valid_json = false;
 
@@ -473,7 +461,7 @@ pub fn try_tool_call_parse_basic_json(
             let extracted_json = trimmed[idx..].trim();
             if !extracted_json.is_empty() {
                 normal_text = extracted_normal;
-                json = JsonPayload::Borrowed(extracted_json);
+                json = Cow::Borrowed(extracted_json);
             }
         }
     } else {
@@ -487,7 +475,7 @@ pub fn try_tool_call_parse_basic_json(
                 match (start_token.is_empty(), end_token.is_empty()) {
                     (false, true) => {
                         // Single token case
-                        let result = handle_single_token_tool_calls(json.as_str(), start_token);
+                        let result = handle_single_token_tool_calls(json.as_ref(), start_token);
                         if let Some(content) = result {
                             // handle_single_token_tool_calls returns either:
                             //   Some("[{...}, ...]") — one or more extracted calls
@@ -500,7 +488,7 @@ pub fn try_tool_call_parse_basic_json(
                                 found_start_token_with_no_valid_json = true;
                             }
 
-                            json = JsonPayload::Owned(content);
+                            json = Cow::Owned(content);
                             // For single token case, use the normal text we extracted earlier
                             normal_text = new_normal_text;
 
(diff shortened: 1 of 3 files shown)


parsers/src/tool_calling/jail/mod.rs
```diff
@@ -150,6 +150,114 @@ struct ChoiceJailState {
...
+impl JsonCompletionProgress {
+    fn reset(&mut self) {
+        *self = Self::default();
+    }
+
+    /// Scan only newly appended bytes and return the last completed top-level
+    /// JSON boundary when new input makes validation useful. The boundary is
+    /// lexical state, not parser acceptance: every appended suffix can trigger
+    /// fresh validation, so a malformed first candidate never pins the jail.
+    fn new_complete_end_from(&mut self, content: &str, base_offset: usize) -> Option<usize> {
```
jthomson04: **Confirmed regression (differential repro vs main):** this incremental lexer never resynchronizes, so a quote- or brace-unbalanced first candidate permanently poisons it (`in_string`/`depth > 0` forever). No boundary is ever reported again, and the per-delta reparse the old `should_exit_jail_early` performed — which would have recovered — no longer runs.

Repro on `llama3_json` with chunks `<|python_tag|>{"` → `<|python_tag|>{"name":"get_time","arguments":{}}` → ` all done!` (also with `<|python_tag|>{{{` as the poisoning chunk): main emits the tool call mid-stream and streams the trailing text; this branch holds the jail to EOF and drops `" all done!"` entirely. The parser itself resyncs at each start token (`input.split(start_token)`), so the lexer needs the same resync — this is exactly the invalid-first-candidate recovery case the PR description says it wanted to preserve.

parsers/src/tool_calling/jail/mod.rs
```diff
@@ -1026,87 +1111,203 @@ impl JailedStream {
...
+                        .await
+                    {
+                        return JailCompletion::Complete(completed);
+                    }
+                }
+
+                // Parsers without markers or JSON framing have no generic
+                // lexical completion signal. Preserve their existing streaming
+                // behavior by validating appended input as before.
+                if self.jail_end_sequences.is_empty()
```
jthomson04: **Confirmed regression (differential repro vs main):** deepseek bare (unwrapped) tool calls never complete mid-stream. For `deepseek_v3`/`deepseek_v3_1` bare inner calls — no `<｜tool▁calls▁begin｜>` wrapper, a form both parsers explicitly recover — `jail_end_sequences` only contains the section-end token that never arrives, `json_completion_base` is `None` (no empty end token), and this every-chunk fallback is gated out because `jail_end_sequences` is non-empty.

Repro: chunk 1 = `<｜tool▁call▁begin｜>get_weather<｜tool▁sep｜>{"location":"Paris"}<｜tool▁call▁end｜>`, chunk 2 = ` And here is some trailing prose.` On main the call is emitted after chunk 1 and the prose streams as content; here the jail buffers to EOF and the trailing prose is dropped entirely (the bare-call parse keeps only text before the call marker as `normal_text`).

parsers/src/tool_calling/jail/mod.rs
```diff
@@ -1026,87 +1111,203 @@ impl JailedStream {
...
-                            } else {
-                                (false, accumulated_content.len())
-                            }
-                        } else {
-                            (false, accumulated_content.len())
+                            return JailCompletion::Complete(completed);
                         }
-                    } else {
-                        (false, accumulated_content.len())
+                        return JailCompletion::Incomplete;
```
jthomson04: **Plausible (not independently reproduced):** `find_incremental_end_marker` advances `next_end_search_start` past a found marker even when `completed_from_valid_marker_parse` returns `None` (e.g. `find_tool_call_end_position` isn't ready yet at the moment the marker arrives). The old code retried the full completion check on every chunk; now that marker's completion decision can never fire again unless a second end sequence shows up, so the stream buffers to EOF.

parsers/src/tool_calling/jail/mod.rs
```diff
@@ -1026,87 +1111,203 @@ impl JailedStream {
...
+            .jail_end_sequences
+            .iter()
+            .filter(|seq| !seq.is_empty())
+            .filter_map(|seq| {
+                accumulated_content[search_start..].find(seq).map(|pos| {
+                    let start = search_start + pos;
+                    (start, start + seq.len())
+                })
+            })
+            .min_by(|(start_a, end_a), (start_b, end_b)| {
```
jthomson04: **Plausible (not independently reproduced):** marker selection order changed. This picks the earliest-starting match in the buffer; the old code took the first sequence in `jail_end_sequences` list order. With overlapping custom sequences (e.g. `</tool_call>` and `</tool`), the split point moves to the short marker's occurrence and can leak part of the real region as visible content. The new test only covers markers at the same start position.

parsers/src/tool_calling/jail/mod.rs
```diff
@@ -1026,87 +1111,203 @@ impl JailedStream {
...
+        let parser_name = self.tool_call_parser.as_deref()?;
+        let parser_map = get_tool_parser_map();
+        let Some(ParserConfig::Json(config)) = parser_map
+            .get(parser_name)
+            .map(|config| &config.parser_config)
+        else {
+            return None;
+        };
+
+        if !config.tool_call_end_tokens.iter().any(String::is_empty) {
```
jthomson04: **Design concern:** the markerless-JSON gate reads the global parser config while marker detection honors builder-set `jail_end_sequences` — two sources of truth for one decision. A caller overriding `.jail_end_sequences(...)` still gets JSON early-exit under the global config's rules. And the capability itself is an implicit empty-string sentinel in `tool_call_end_tokens`: tidying away mistral's odd-looking `""` entry would silently disable incremental completion with no type or test error. An explicit capability flag on `ParserConfig` would be sturdier.

parsers/src/tool_calling/jail/mod.rs
```diff
@@ -150,6 +150,114 @@ struct ChoiceJailState {
...
+                        self.last_complete_end = Some(pos + ch.len_utf8());
+                        self.started = false;
+                    }
+                }
+                _ => {}
+            }
+        }
+
+        self.scanned_len = content.len();
+        if self.last_complete_end.is_some() && self.last_validation_len < content.len() {
```
jthomson04: **Efficiency — one still-quadratic state remains:** once a balanced-but-invalid candidate exists, this `last_validation_len < content.len()` gate re-triggers a full-buffer `try_tool_call_parse_aggregate` on every appended delta (e.g. a malformed first candidate followed by long trailing prose). Gating revalidation on `last_complete_end` advancing to a *new* boundary would keep the `invalid_balanced_candidate` regression test passing without the per-delta reparse.

jthomson04 (commented):
Review of the incremental jail port. The speedups look real, but the incremental state machine loses two behaviors the old reparse-every-chunk approach guaranteed — both cause user-visible content to be silently dropped. Findings 1–2 were confirmed by differential repro against main; 3–5 are plausible mechanisms that weren't independently reproduced; the rest are efficiency/cleanup.

Below the inline comments, also noted but lower priority:
- The pythonic/no-marker family still takes the old full-reparse-per-chunk path (pre-existing; just not covered by the PR's incremental claim).
- The jail's new lexer duplicates the lexer inside `try_repair_truncated_json`, and the two already disagree on mismatched closers (single depth counter with `saturating_sub` vs a typed closer stack) — a shared boundary scanner in `tool_calling/json` would prevent drift.
- `find_tool_call_end_position` remains a parser-name allowlist one layer down, and the new design leans on it more heavily: a new JSON-family parser falls into the `_ => chunk.len()` arm and swallows trailing content into the tool-call emission.

parsers/src/tool_calling/jail/mod.rs
```diff
@@ -1026,87 +1111,203 @@ impl JailedStream {
...
+                    {
+                        return JailCompletion::Complete(completed);
+                    }
+                }
+
+                // Parsers without markers or JSON framing have no generic
+                // lexical completion signal. Preserve their existing streaming
+                // behavior by validating appended input as before.
+                if self.jail_end_sequences.is_empty()
+                    && self.json_completion_base(accumulated_content).is_none()
```
jthomson04: **Efficiency:** `json_completion_base` is computed twice per chunk (here and in the branch above), each time doing a parser-map lookup plus full-buffer `find` from offset 0 — for a base offset that can't change in an append-only buffer once found. Computing it once per call (or caching it in `JailCompletionProgress`, which exists for exactly this kind of state) removes an O(buffer) scan per delta from the hot path.

parsers/src/tool_calling/json/base_json_parser.rs
```diff
@@ -41,49 +40,83 @@ pub struct CalledFunctionNameOnly {
...
-                    // Return the last match directly
-                    return Some(matches.last().unwrap().clone());
-                } else {
-                    // Join the matches into a JSON array string
-                    return Some(format!("[{}]", matches.join(",")));
-                }
-            }
-            None
+#[derive(Debug)]
+enum JsonPayload<'a> {
```
jthomson04: **Cleanup:** `JsonPayload<'a>` reimplements `std::borrow::Cow<'a, str>` — `Cow::Borrowed`/`Cow::Owned` plus deref give you everything here, and it's the shape readers expect. Relatedly, with `parse_calls` now deserializing only `CalledFunctionRaw`, the `CalledFunctionParameters`/`CalledFunctionArguments`/`CalledFunctionNameOnly` structs above are dead in-crate (being `pub`, no lint will flag them), and the doc comment further down still describes parsing in terms of them.

parsers/v1/src/tool_calling/json/base_json_parser.rs
```diff
@@ -25,65 +28,84 @@ pub struct CalledFunctionParameters {
...
+    #[serde(default, deserialize_with = "deserialize_present_raw")]
+    parameters: Option<Box<RawValue>>,
+    #[serde(default, deserialize_with = "deserialize_present_raw")]
+    arguments: Option<Box<RawValue>>,
+}
+
+fn deserialize_present_raw<'de, D>(deserializer: D) -> Result<Option<Box<RawValue>>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
```
keivenchang, line 25: now that `parse_calls` deserializes only `CalledFunctionRaw`, the three structs above (`CalledFunctionParameters` / `CalledFunctionArguments` / `CalledFunctionNameOnly`) are dead — 0 references left in the crate, and since they're `pub` nothing lints them. the doc comment further down (~L317) still describes parsing in terms of them too. worth deleting so the file matches what parse_calls actually does.

keivenchang (commented):
hey John, I just ran the tests + conformance off this branch and it all comes back green (regressions have coverage now, golden set matches: 121/121 stream, 642/642 batch, 58/59 stream-on-batch), so this looks good to me. thanks for checking the code against the tests. one tiny cleanup below.

parsers/src/tool_calling/jail/mod.rs
```diff
@@ -1026,87 +1111,203 @@ impl JailedStream {
...
-                        if let Ok((_, _)) = try_tool_call_parse_aggregate(
-                            accumulated_content,
-                            Some(parser),
-                            tools_slice,
-                        )
-                        .await
+                if let Some(end_pos) =
+                    self.find_incremental_end_marker(accumulated_content, progress)
+                {
+                    let parse_result = self.parse_marker_tool_calls(accumulated_content).await;
```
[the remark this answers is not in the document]
keivenchang: looks handled on the current head (a9550a8) — check_jail_completion's MarkerBased branch now guards `self.tool_call_parser.is_none()` (mod.rs:1283) and completes at the detected marker end with `marker_parse_result: None`, so marker-only jails dont buffer to EOF anymore. thread's resolvable.

keivenchang (approved):
LGTM. thanks for keeping the golden set honest — i re-captured the dynamo side off this branch and re-ran the conformance checks myself, no deviation (121/121 stream, 642/642 batch, 58/59 stream-on-batch w/ the 1 known). regressions are fixed + tested.

PR #101, opened 2026-07-09
fix(jail): normalize terminal tool-call emissions

rmccorm4:
## Summary

- normalize each upstream terminal choice to exactly one emitted `finish_reason`
- suppress terminal whitespace-only content after a parsed tool call
- preserve meaningful trailing content and non-`stop` finish reasons
- restore causal prefix → parsed tool/content → trailing suffix output order
- cover Qwen3-Coder, Hermes, immediate JSON mode, split/coalesced terminals, and multi-choice streams

## Root cause

The v1 jail can split one terminal engine delta ending in `</tool_call>\n` into a parsed tool-call emission and a trailing newline emission. Both emissions inherited the upstream `stop`, and the finish-reason postprocessor rewrote both to `tool_calls`.

Strict clients then observed two terminal chunks for one choice.

### Before

```mermaid
flowchart TD
    A["vLLM terminal delta<br/>content = tool-call markup + newline<br/>finish_reason = stop"]
    A --> B["v1 jail parses and splits the delta"]
    B --> C["Structured tool-call emission<br/>inherits stop"]
    B --> D["Trailing newline emission<br/>inherits stop"]
    C --> E["Finish postprocessor sees a tool call"]
    D --> E
    E --> F["Tool-call chunk<br/>finish_reason = tool_calls"]
    E --> G["Newline chunk<br/>finish_reason = tool_calls"]
```

### After

```mermaid
flowchart TD
    A["vLLM terminal delta<br/>content = tool-call markup + newline<br/>finish_reason = stop"]
    A --> B["v1 jail parses and splits the delta"]
    B --> C["Structured tool-call emission<br/>temporarily inherits stop"]
    B --> D["Trailing newline emission<br/>temporarily inherits stop"]
    C --> E["Terminal-emission normalizer"]
    D --> E
    E --> F["Drop whitespace-only suffix<br/>Clear copied finish reasons<br/>Assign stop to the sole survivor"]
    F --> G["Finish postprocessor sees the tool call<br/>stop → tool_calls"]
    G --> H["One tool-call chunk<br/>finish_reason = tool_calls<br/>no newline content chunk"]
```

## Impact

A coalesced terminal tool-call delta now produces one structured tool-call chunk with one finish reason and no redundant newline content. If terminal whitespace arrives separately, it becomes a content-less terminal signal rather than a newline delta.

## Validation

Parser checks:

- `cargo test -p dynamo-parsers`: 745 passed, 5 ignored, 0 failed
- `cargo clippy -p dynamo-parsers --tests -- -D warnings`
- `git diff --check`

Live E2E:

- rebuilt Dynamo's Python core with `dynamo-parsers v4.1.0` and `dynamo-protocols v2.0.1` pinned to this PR commit (`5ad1e6e`)
- loaded the rebuilt core in `nvcr.io/nvstaging/ai-dynamo/vllm-runtime:1.3.0-rc.2`; host/container core SHA-256 matched
- unchanged RC2 vLLM 0.23 worker with NVIDIA Nemotron 3 Nano 30B A3B NVFP4, `qwen3_coder`, and `nemotron_v3`
- compared the RC2 default-v1 and experimental-v2 frontends with the rebuilt default-v1 frontend

Exact attached-request replay:

| | RC2 default v1 | RC2 experimental v2 (tool-call response) | Patched default v1 |
|---|---:|---:|---:|
| HTTP | 200 | 200 | 200 |
| tool events | 1 | 1 | 1 |
| non-null finishes | 2 | 1 | 1 |
| terminal finish | `tool_calls` ×2 | `tool_calls` | `tool_calls` |
| post-tool whitespace deltas | 1 | 0 | 0 |
| usage before `[DONE]` | yes | yes | yes |

The frontend and worker were removed after validation; both GPUs returned to idle.

Linear: [DIS-2374](https://linear.app/nvidia/issue/DIS-2374/fix-duplicate-terminal-finish-reason-in-v1-tool-call-jail)


rmccorm4:
~~@keivenchang did the red/green version of tool call parity table break: https://github.com/ai-dynamo/frontend-crates/pull/101#issuecomment-4921913591? It seems to be working for reasoning tables but not for the tool calling table~~

EDIT: This seems to have been fixed by some other recent changes.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -446,6 +469,75 @@ impl ChoiceJailState {
...
+    /// normalization both inherit its finish reason. Terminal formatting
+    /// whitespace after a tool call is also parser framing rather than assistant
+    /// content and should not become a second content delta.
+    fn normalize_terminal_emissions(
+        &self,
+        choice: &ChatChoiceStream,
+        had_tool_calls_before: bool,
+        emissions: &mut Vec<ChoiceEmission>,
+    ) {
+        let Some(finish_reason) = choice.finish_reason else {
```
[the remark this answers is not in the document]
rmccorm4: The purpose of this PR was to avoid having multiple finish_reason response chunks, not necessarily eliminate any/all trailing whitespace content, so will defer this for now.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -1018,15 +1125,6 @@ impl JailedStream {
...
-
-                        // Emit pass-through content with current metadata
-                        if !passthrough_emissions.is_empty() {
-                            let current_metadata = (response.id.clone(), response.event.clone(), response.comment.clone());
-                            let responses = self.emit_choice_emissions(passthrough_emissions, chat_response, current_metadata);
-                            for emitted_response in responses {
-                                yield emitted_response;
-                            }
-                        }
                     }
```
keivenchang, line 1132: The terminal finish is assigned to `emissions.last_mut()`, but the emit step below re-buckets into passthrough -> tool_content -> trailing and emits in that order. Correct only while `process_content` never yields a `PassThrough` after a `ToolCall`/`Trailing` in one choice-delta. Consider a comment pinning that invariant or a `debug_assert!` — a future change could otherwise emit the finish-reason chunk out of order and silently reintroduce the split-terminal bug.

keivenchang (approved):
Nice fix. I pulled the branch and ran it locally against the conformance suite — v1 batch, batch-via-stream, and stream parity all pass, plus the jail unit tests (75/0, including your 7 new terminal cases). Also double-checked the finish-reason behavior isn't in the fixture corpus, so nothing regressed there.

One tiny non-blocking note inline about the emit-ordering invariant. LGTM — thanks!

keivenchang:
@rmccorm4 thanks for this — approving. I'll take a look at the coloring (GUI) issue shortly and follow up separately.

parsers/v1/src/tool_calling/jail/mod.rs
rmccorm4: Added inline comment: https://github.com/ai-dynamo/frontend-crates/pull/101/commits/706df6367ac88ca3a71b5ca4ce6ba4d1e170c8b4

PR #106, opened 2026-07-10
feat(parsers): add DYNAMO_PARSERS_DEBUG env-gated stderr instrumentation

keivenchang:
## What

Opt-in stderr debug for `dynamo-parsers-v2`, so we can debug the Dynamo parser when it's embedded in a host framework.

Today there's no way to see, from the host process, which Dynamo parser got picked or whether it parsed anything — the only logs are `tracing` lines the host never surfaces. This adds a subscriber-free toggle. One wrapper at the `create_tool_parser_for_family` choke point covers every family; off by default, zero cost otherwise.

Linear: [DIS-2379](https://linear.app/nvidia/issue/DIS-2379)

## Use

`DYNAMO_PARSERS_DEBUG=1` →

```
[dynamo-parsers-v2] family=qwen3_coder created
[dynamo-parsers-v2] family=qwen3_coder finish emitted 1 call update(s) names=["get_weather"]
```

## Test

`cargo test -p dynamo-parsers-v2` — 31 passed. Bumps to `0.1.18`.

### DIFF
--- a/parsers/v2/src/tool_calling/debug.rs
+++ b/parsers/v2/src/tool_calling/debug.rs
@@ -1,0 +1,159 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! ENV-gated stderr debug tracing for the v2 stream parsers.
+//!
+//! Set `DYNAMO_PARSERS_DEBUG=1` to make the crate print a marker to stderr when
+//! a parser is created via `create_tool_parser_for_family` and whenever a
+//! parser emits tool-call updates. This lets a host process (for example
+//! vLLM's experimental Rust frontend) confirm that a Dynamo parser was
+//! selected and is actually parsing, without installing a `tracing` subscriber.
+//!
+//! The toggle is read once and defaults to off, so the normal path pays no cost.
+
+use std::io::Write;
+use std::sync::OnceLock;
+
+use super::traits::{Result, Tool, ToolParseResult, ToolParser};
+
+/// Env var that turns on `dynamo-parsers-v2` stderr debug output.
+pub const DEBUG_ENV: &str = "DYNAMO_PARSERS_DEBUG";
+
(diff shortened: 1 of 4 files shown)


parsers/v2/src/tool_calling/debug.rs
```diff
@@ -0,0 +1,138 @@
...
+/// overhead on the default path. One wrapper at the dispatch choke point
+/// instruments every family, so there is no per-parser duplication.
+pub(super) struct DebugToolParser {
+    family: String,
+    inner: Box<dyn ToolParser>,
+}
+
+impl DebugToolParser {
+    pub(super) fn wrap(family: &str, inner: Box<dyn ToolParser>) -> Box<dyn ToolParser> {
+        eprintln!("[dynamo-parsers-v2] family={family} created");
```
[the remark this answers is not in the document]
keivenchang: fixed in ef6f3c0 — both debug lines now go through a writeln! helper that discards the io::Result, so a broken/closed stderr cant panic the host

parsers/v2/src/tool_calling/debug.rs
```diff
@@ -0,0 +1,147 @@
...
+fn emit(args: std::fmt::Arguments<'_>) {
+    let mut stderr = std::io::stderr().lock();
+    let _ = writeln!(stderr, "[dynamo-parsers-v2] {args}");
+}
+
+/// Whether debug output is enabled. Read once from the environment.
+pub fn debug_enabled() -> bool {
+    static ENABLED: OnceLock<bool> = OnceLock::new();
+    *ENABLED.get_or_init(|| {
+        std::env::var(DEBUG_ENV).is_ok_and(|v| matches!(v.as_str(), "1" | "true" | "True" | "TRUE"))
```
rmccorm4: note for future, we can add a similar env_is_truthy() helper like we have in dynamo for this type of check

rmccorm4 (approved):

parsers/v2/src/tool_calling/debug.rs
keivenchang: added in a987db9 — is_truthy + env_is_truthy matching dynamo's exact pattern (1/true/on/yes, lowercased)

PR #110, opened 2026-07-11
fix(tokenizer): support tokenizer options

rainj-me:
# Overview

Support tokenizer options. This change is mandatory for Sglang rust server since the add_special_tokens is enabled by default during encode. Without this, some model gsm8k return low value.

Refer: https://github.com/huggingface/transformers/blob/1f8daee0c80cd53a44b1094b79309dd8675d6392/src/transformers/tokenization_utils_base.py#L2258

### DIFF
--- a/tokenizers/src/hf.rs
+++ b/tokenizers/src/hf.rs
@@ -6,12 +6,16 @@ use std::path::Path;
 use tokenizers::tokenizer::{AddedToken, Tokenizer as HfTokenizer};
 
 use super::{
-    Encoding, Error, Result, TokenIdType,
+    Encoding, Error, Result, TokenIdType, TokenizerOptions,
     traits::{DecodeResult, Decoder, Encoder, Tokenizer},
 };
 
 pub struct HuggingFaceTokenizer {
     tokenizer: HfTokenizer,
+    /// Construction-time options; see [`TokenizerOptions`]. Defaults preserve
+    /// the historical behavior (e.g. `add_special_tokens: false`). Set via
+    /// [`Tokenizer::with_options`].
+    options: TokenizerOptions,
 }
 
 impl HuggingFaceTokenizer {
@@ -27,11 +31,14 @@ impl HuggingFaceTokenizer {
             merge_special_tokens_from_config(&mut tokenizer, parent);
         }
 
-        Ok(HuggingFaceTokenizer { tokenizer })
+        Ok(Self::from_tokenizer(tokenizer))
     }
 
     pub fn from_tokenizer(tokenizer: HfTokenizer) -> Self {
-        HuggingFaceTokenizer { tokenizer }
+        HuggingFaceTokenizer {
+            tokenizer,
+            options: TokenizerOptions::default(),
+        }
     }
 
     /// Wrap an already-loaded `HfTokenizer` and merge in the sibling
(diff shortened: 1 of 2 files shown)


rmccorm4:
Hi @rainj-me, thanks for raising this!

1. Can you sign your commits to resolve the failing DCO check?
2. Can you address this [comment](https://github.com/ai-dynamo/frontend-crates/pull/110#discussion_r3562684274) to add a new helper to avoid the breaking change?
3. Please address the failing linter check: https://github.com/ai-dynamo/frontend-crates/actions/runs/29132693583/job/86490986448?pr=110

Otherwise LGTM!

tokenizers/src/hf.rs
```diff
@@ -136,7 +143,7 @@ impl Encoder for HuggingFaceTokenizer {
         // This self.tokenizer is the library
         let encoding = self
             .tokenizer
-            .encode(input, false)
+            .encode(input, self.options.add_special_tokens)
```
rmccorm4, line 146: _Comment generated by AI: for future PR follow up_

`add_special_tokens=true` breaks the exported `CachedTokenizer` invariant. L1 splits prompts at registered special-token boundaries and calls the inner `encode` once for each segment and tail; this line now runs the HF post-processor on every segment, duplicating BOS/EOS. With TinyLlama at this head, the direct target encoding was 23 tokens versus 30 on a cold cache and 26/27 after warming with extension off/on, so the returned token IDs also depend on cache history. Dynamo enables L1 by default. Please either bypass L1 for this mode or make the cache apply the post-processor exactly once, and add cold/warm scalar and batch parity coverage.

tokenizers/src/hf.rs
```diff
@@ -27,19 +31,22 @@ impl HuggingFaceTokenizer {
             merge_special_tokens_from_config(&mut tokenizer, parent);
         }
 
-        Ok(HuggingFaceTokenizer { tokenizer })
+        Ok(Self::from_tokenizer(tokenizer))
```
rmccorm4, line 34: _Comment generated by AI: for future PR follow up_

This loader only materializes the raw `tokenizer.json`; the sibling-config merge registers added tokens but does not honor `add_bos_token`/`add_eos_token` or build the effective post-processor. At `deepseek-ai/DeepSeek-R1-Distill-Llama-8B@6a6f4aa4197940add57724a7707d069478df56b1`, Transformers 4.57.1 encodes `hello` with specials as `[128000, 15339]`, while raw tokenizers 0.21.4—the path used here—returns `[15339]` with either flag value. Please materialize the config-derived effective post-processor (or equivalent model semantics) and add a config-only BOS regression test.

tokenizers/src/lib.rs
```diff
@@ -217,6 +228,19 @@ impl Encoding {
     }
 }
 
+/// Construction options for [`Tokenizer::from_file_with_options`] /
+/// [`create_tokenizer_from_file`], applied to concrete tokenizers via
+/// [`traits::Tokenizer::with_options`].
+#[derive(Debug, Clone, Copy, Default)]
+pub struct TokenizerOptions {
```
rmccorm4, line 235: _Comment generated by AI: for future PR follow up_

This new public all-fields options struct is not safely extensible: adding any subsequent option breaks external struct-literal construction, and adding `#[non_exhaustive]` later is itself a major change under Cargo's SemVer guidance. Since the type is new, please mark it `#[non_exhaustive]` now and provide a constructor/builder, or make the fields private behind construction APIs.

rmccorm4 (commented):
_Comment generated by AI: for future PR follow up_

Three follow-up findings are attached inline.

rmccorm4 (approved):
Few comments for us to follow up on.

Thanks for the contribution @rainj-me !

tokenizers/src/hf.rs
rmccorm4: @jthomson04 @ishandhanani let's fast follow on this

jthomson04: I'll put up a PR to fix.

tokenizers/src/lib.rs
rainj-me: good call out.

PR #111, opened 2026-07-11
ci: run copy-pr-bot mirror + gate conformance-table on conformance/parser changes

keivenchang:
## Overview

External-fork PRs to this repo fail CI in a way unrelated to what they change, and `/ok to test <sha>` doesn't help. Root causes, both in `.github/workflows/ci.yml`:

1. The fixtures moved to a **private** HF dataset (`ai-dynamo/conformance-fixtures`), so `conformance-table` (and `rust`) need `secrets.HF_TOKEN`. GitHub never exposes secrets to `pull_request` runs from forks, so the download 401s (`RepositoryNotFoundError`). Example: #110 (tokenizer-only) fails `conformance-table` on the HF download.
2. `copy-pr-bot` is enabled and creates the internal `pull-request/<n>` branch on `/ok to test`, but no workflow triggered on `pull-request/**`, so the vetted run never started.
3. `conformance-table` ran on every PR even when no conformance/parser code changed.

Tracking: [DIS-2384](https://linear.app/nvidia/issue/DIS-2384).

## Details

- **Run the copy-pr-bot mirror.** Add `pull-request/[0-9]+` to the `push` branch filter so a maintainer-vetted fork PR runs on internal runners **with** secrets (mirrors how `ai-dynamo/dynamo`'s `pr.yaml` is triggered).
- **Path-gate `conformance-table`.** New `changes` job (`dorny/paths-filter`, SHA-pinned) sets a `conformance` output from `conformance/**`, `parsers/**`, and the workflow file itself; `conformance-table` now `needs: changes` and runs only `if` that output is `true`. Unrelated PRs (tokenizer, docs) skip the expensive private-fixture render + chart-invariant guards.

## Verification

- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` → OK.
- Expected behavior:
  - Internal PR touching `conformance/**`/`parsers/**` → `conformance-table` runs (unchanged).
  - Internal PR touching only e.g. `tokenizers/**` → `conformance-table` skipped.
  - Fork PR + `/ok to test <sha>` → `pull-request/<n>` run starts on internal runners, has `HF_TOKEN`, and `conformance-table` passes (when relevant).

## Notes / follow-ups

- **Branch protection:** a job skipped via job-level `if:` reports as *skipped*, which satisfies a required check — but if `conformance-table` is a required check, confirm the org treats skipped as passing (standard GitHub behavior does).
- **Out of scope here:** the `rust` job also reads `HF_TOKEN` (fixture download during `cargo test`), so a fork's own `pull_request` `rust` run can still 401 until `/ok to test` runs the mirror. Not changed in this PR; flagged in DIS-2384.

---

@coderabbitai review

---

Reviewer, start here: `.github/workflows/ci.yml` — the `on.push.branches` addition and the new `changes` job + `conformance-table` `needs`/`if`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,13 @@ name: ci
 
 on:
   push:
-    branches: [main]
+    # `pull-request/<n>` is the internal mirror branch copy-pr-bot pushes after a
+    # maintainer runs `/ok to test <sha>`. Running here (not the fork's own
+    # `pull_request` event) is the only way an external contribution gets
+    # `secrets.HF_TOKEN`, which the private fixture dataset download requires.
+    branches:
+      - main
+      - "pull-request/[0-9]+"
   pull_request:
   workflow_dispatch:
 
@@ -10,6 +16,31 @@ permissions:
   contents: read
 
 jobs:
+  # Decide which optional jobs need to run for this change. `conformance-table`
+  # is expensive (downloads the private fixture dataset, renders both charts) and
+  # is only affected by conformance/parser changes, so skip it otherwise.
+  changes:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      conformance: ${{ steps.filter.outputs.conformance }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - name: Detect conformance/parser changes
+        id: filter
+        uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
+        with:
+          filters: |
+            conformance:
+              - 'conformance/**'
+              - 'parsers/**'
+              - '.github/workflows/ci.yml'
+
   rust:
     runs-on: ubuntu-latest
     timeout-minutes: 30
@@ -72,6 +103,8 @@ jobs:
   # breakage. Python + pyyaml/jinja2 only — no engines (vLLM/SGLang conformance runs in
   # dynamo's CI lanes; here the engine lanes are on-demand/local).
   conformance-table:
+    needs: changes
+    if: ${{ needs.changes.outputs.conformance == 'true' }}
     runs-on: ubuntu-latest
     timeout-minutes: 10
     permissions:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
   # dynamo's CI lanes; here the engine lanes are on-demand/local).
   conformance-table:
     needs: changes
-    if: ${{ needs.changes.outputs.conformance == 'true' }}
+    if: ${{ needs.changes.outputs.conformance == 'true' || github.event_name == 'workflow_dispatch' }}
     runs-on: ubuntu-latest
     timeout-minutes: 10
     permissions:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,6 @@ on:
     branches:
       - main
       - "pull-request/[0-9]+"
-  pull_request:
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


dmitry-tokarev-nv (approved):
Claude's Findings

1. Every internal PR now runs full CI twice (new, recurring cost). copy-pr-bot auto-mirrors trusted-user PRs without /ok to test — this PR's mirror push run started 2 seconds after its pull_request run. Before this PR those mirror pushes triggered nothing; now each internal PR push costs a duplicate rust (~5.5 min) + conformance-table (~5 min) run, ~10 duplicated runner-minutes per push. dynamo avoids this by having only the mirror push trigger and no pull_request trigger. Options:
- Accept it (my recommendation for this PR — low-volume repo, and dropping pull_request would remove instant feedback for forks and break the PR-comment step).
- Follow-up: adopt dynamo's push-only model, porting the comment step to derive the PR number from the branch name.
- Do not try to dedupe by skipping rust on internal-mirror push runs: a later skipped rust check run on the same SHA would satisfy the required check even if the earlier pull_request rust run failed — that would neuter branch protection for internal PRs.

2. workflow_dispatch on main now silently skips conformance-table. Verified in the paths-filter source: with no before payload and base == head == default branch, it warns and diffs only the last commit — so a manual dispatch (the natural way to re-render the matrix after the private HF dataset or an unpinned pip dep drifts, with no repo change) will almost always produce conformance: false. One-line fix worth taking in this PR:
if: ${{ needs.changes.outputs.conformance == 'true' || github.event_name == 'workflow_dispatch' }}

3. (Minor) The post-merge canary is now conditional. Previously every main push exercised the render against the live private dataset and unpinned pip install pyyaml jinja2 huggingface_hub pytest; now only conformance-touching pushes do. External drift will surface late (on the next conformance PR, as a confusing unrelated failure). Consider a schedule: trigger (combined with the fix in #2's if) as a cheap follow-up.

4. (Note) The skip path itself hasn't been exercised — this PR touches ci.yml, which is in the filter, so conformance-table ran in both runs. Worth watching the first tokenizer-only PR (e.g., a #110 rebase) to confirm it skips as intended. Also note changes becomes a prerequisite: if it ever fails (API flake), conformance-table is skipped — but the run goes red, so it's visible, not silent.

Quality / style / security

- Comments explain why at the constraint level (why push-not-pull_request, why the job is expensive) — genuinely useful, matches the file's existing style.
- Filter self-includes .github/workflows/ci.yml — correct, prevents the gate from hiding its own regressions.
- Minimal permissions throughout: top-level contents: read stands; the new job adds only pull-requests: read; the third-party action runs with no secrets. Secrets-with-vetted-code via copy-pr-bot is the established org model, and this repo is internal visibility per the enterprise ruleset.
- The rust-still-401s-on-fork-pull_request gap is honestly scoped out and tracked in DIS-2384; with rust as the required check, the mirror-run-supersedes mechanism covers merging.

.github/workflows/ci.yml
```diff
@@ -2,14 +2,45 @@ name: ci
...
   push:
-    branches: [main]
+    # `pull-request/<n>` is the internal mirror branch copy-pr-bot pushes after a
+    # maintainer runs `/ok to test <sha>`. Running here (not the fork's own
+    # `pull_request` event) is the only way an external contribution gets
+    # `secrets.HF_TOKEN`, which the private fixture dataset download requires.
+    branches:
+      - main
+      - "pull-request/[0-9]+"
   pull_request:
```
nv-anants: need to remove this

keivenchang:
thanks dmitry — addressed all three:

1. fixed in bb770d2 — removed `pull_request:` so CI only runs via the copy-pr-bot push path, no more double runs.
2. fixed in bb770d2 — added `|| github.event_name == 'workflow_dispatch'` to the conformance-table `if:`.
3. agree on the schedule follow-up — punting to a separate PR (low urgency, and this one's already grown scope).

.github/workflows/ci.yml
keivenchang: hey anant, removed `pull_request:` in `4fa5fe2` but kept the `pull-request/[0-9]+` branch pattern — that's what copy-pr-bot pushes to after `/ok to test <sha>`, and it's the only path that gets `HF_TOKEN` to fork PRs. removing it would mean fork PRs get no CI at all. lmk if you had something different in mind.

nv-anants (approved):

.github/workflows/ci.yml
```diff
@@ -2,14 +2,44 @@ name: ci
...
   push:
-    branches: [main]
-  pull_request:
+    # `pull-request/<n>` is the internal mirror branch copy-pr-bot pushes after a
+    # maintainer runs `/ok to test <sha>`. Running here (not the fork's own
+    # `pull_request` event) is the only way an external contribution gets
+    # `secrets.HF_TOKEN`, which the private fixture dataset download requires.
+    branches:
+      - main
+      - "pull-request/[0-9]+"
```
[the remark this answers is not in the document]
keivenchang: not fixing — `[0-9]+` works as intended here; GitHub's pattern engine (micromatch) treats `+` after a character class as "one or more digits", so `pull-request/[0-9]+` matches `pull-request/111`. confirmed: CI run [#29297062133](https://github.com/ai-dynamo/frontend-crates/actions/runs/29297062133) was triggered by copy-pr-bot on the `pull-request/111` branch. dynamo uses the same pattern.

PR #115, opened 2026-07-13
fix(tokenizers): reject incompatible prefix caching

jthomson04:
## Summary

- make prefix-cache compatibility an explicit, fail-closed tokenizer validation contract
- have `CachedTokenizer::new` call the inner tokenizer's validation and propagate its error
- have Hugging Face return the incompatibility reason for `add_special_tokens=true`, while FastTokenizer and TikToken explicitly opt in
- return a clear construction error without warning, fallback, or changes to cache behavior
- add one focused rejection test using the existing HF fixture

Follow-up to the tokenizer-cache correctness issue identified on #110.

## Breaking change

`CachedTokenizer::new` now returns `Result<Self>`. Custom tokenizer implementations must explicitly opt into prefix-cache compatibility through `validate_prefix_cache` and own any incompatibility error they return.

## Validation

- `cargo test -p dynamo-tokenizers rejects_hf_tokenizer_that_adds_special_tokens`
- `cargo test -p dynamo-tokenizers`
- `cargo test --workspace --exclude dynamo-conformance-fixtures-v2`
- `cargo test --workspace --doc --exclude dynamo-conformance-fixtures-v2`
- `cargo clippy --workspace --all-targets --all-features --exclude dynamo-conformance-fixtures-v2 -- -D warnings`


tokenizers/src/lib.rs
```diff
@@ -241,6 +241,35 @@ pub struct TokenizerOptions {
     pub add_special_tokens: bool,
 }
 
+impl TokenizerOptions {
+    /// Resolve whether tokenizer caching should be enabled for these options.
+    pub fn resolve_tokenizer_cache(&self, cache_enabled: bool) -> bool {
```
rmccorm4: (from Devin)

**New public method has no callers in this repository**

The new `resolve_tokenizer_cache` method at `tokenizers/src/lib.rs:246` is public but has no callers within this repository. Per the CONTRIBUTING.md sync model, this crate is consumed by `ai-dynamo/dynamo`, so the caller likely lives there. This is expected for a library crate, but worth confirming that the downstream consumer exists or is being added in a coordinated PR.

jthomson04: yeah, just noticed that myself. Pushing fix shortly

tokenizers/src/cache/mod.rs
```diff
@@ -104,19 +104,26 @@ impl CachedTokenizer {
...
+    ///
+    /// Returns the inner tokenizer's compatibility error when it cannot be
+    /// safely wrapped in the prefix cache.
     pub fn new(
         inner: Arc<dyn Tokenizer>,
         special_tokens: Vec<String>,
         max_memory_bytes: usize,
-    ) -> Self {
+    ) -> Result<Self> {
+        inner.validate_prefix_cache()?;
```
rmccorm4, line 117: (from Devin)

**Validation runs even in passthrough mode (empty specials), narrowing the API contract**

The `validate_prefix_cache()` call at `tokenizers/src/cache/mod.rs:117` runs unconditionally, before checking whether `special_tokens` is empty. When specials are empty, L1 is disabled and `encode`/`encode_batch` short-circuit to the inner tokenizer with zero cache interaction. Previously, a tokenizer with `add_special_tokens=true` could be wrapped in passthrough mode (empty specials) without issue. Now it will be rejected even though no caching occurs. This is a stricter contract than the README's "An empty list disables caching" implies. If downstream callers rely on empty-specials passthrough with `add_special_tokens=true` HF tokenizers, this will break. The fix would be to skip validation when `special_tokens.is_empty()`, but the current conservative approach is also defensible — just worth documenting.

jthomson04: Hmm. Given that we effectively fallback to the wrapped tokenizer when special tokens is none, I think this new api contract is fine. If somebody explicitly instantiates a cachedtokenizer, it seems reasonable to throw an error if we just end up falling back to the uncached version.

rmccorm4 (approved):

PR #117, opened 2026-07-14
ci: make crate version bumps manual, stop release-plz auto-bump

keivenchang:
#### Overview:

In the past, every merge to main auto-bumped crate versions (release-plz ran on each push). That caused constant churn: `dynamo-parsers-v2` went 0.1.17 → 0.1.21 in days, it burned version space, downstream pins like vLLM's `dynamo-parsers-v2 = =0.1.x` went stale, and it kept re-releasing v1 and v2 together even though they share no code and one hadn't changed.

So now version bumps are manual: a human edits a crate's `[package]` version in their PR, and CI only publishes what was bumped.

#### Details:

- `release.yml`: dropped `release-plz update`, the self-commit/push, the recursion guard, and the `cargo-semver-checks` install; the publish step runs unconditionally and only publishes crates whose version isn't on crates.io yet.
- `ci.yml`: added a `semver` job so we don't lose API-compat safety — `cargo-semver-checks` runs on every PR and fails if a breaking change isn't matched by a version bump.
- `release-plz.toml`: comment-only; points at the new CI gate and notes `semver_check` still applies to a local `release-plz update`.

#### Verification:

YAML/TOML validated; the `semver` job's real run happens on this PR's CI mirror.

`.github/workflows/release.yml`, then `.github/workflows/ci.yml`

/coderabbit profile chill

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,6 +98,41 @@ jobs:
       - name: Verify crates are publishable
         run: cargo publish --workspace --dry-run --locked
 
+  # Public-API SemVer gate. With version bumps now manual (DIS-2414), nothing in
+  # CI auto-computes the required bump anymore, so this job is the safety net:
+  # cargo-semver-checks compares each publishable crate's current public API
+  # against its latest crates.io release and fails if the change needs a bigger
+  # bump than the crate's [package] version reflects. A breaking change must
+  # therefore also bump the version IN THE SAME PR, before merge — a stricter,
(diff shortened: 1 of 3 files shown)


keivenchang:
Superseded by #118. The branch was renamed for the expanded scope (DIS-2414 release-flow work folded into [DIS-2432](https://linear.app/nvidia/issue/DIS-2432) with the git-lfs fixture migration), and GitHub closed this PR on the rename and refuses to reopen a PR whose head ref was deleted. All work continues in #118.

PR #120, opened 2026-07-16
feat(parsers): add Inkling tool-call and reasoning parsers

KrishnanPrash:
#### Overview
Adds Inkling (thinkingmachines/Inkling-NVFP4) tool-call and reasoning parsers to `dynamo-parsers`, hardened against the model's real output shape and backed by real end-to-end fixtures.

A live self-test surfaced that Inkling's assistant turn begins header-less: the chat template's generation primer `<|message_model|>` is consumed by the prompt (`add_generation_prompt`), so the model's first block arrives as `<|content_thinking|>...` or `NAME<|content_invoke_tool_json|>...` with no leading `<|message_model|>`. The reasoning parser only entered its block router on that leading token, so the first block was mis-routed: tool calls were dropped and reasoning leaked into content.

Repro (live dynamo+sglang Inkling deployment with the inkling parsers configured):
```
curl -s localhost:8000/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model":"inkling-nvfp4",
  "messages":[{"role":"user","content":"Get the weather in San Francisco."}],
  "tools":[{"type":"function","function":{"name":"get_weather","parameters":{"type":"object","properties":{"location":{"type":"string"}},"required":["location"]}}}],
  "chat_template_kwargs":{"reasoning_effort":"none"}}'
```
Before: `content: "get_weather{\"name\":...}"`, `tool_calls: null` (markup leaked, no call).
After: `tool_calls: [{name: get_weather, arguments: {location: "San Francisco"}}]`, `content: null`.

#### Details
Inkling serializes an assistant turn as a sequence of typed blocks rather than a `<think>...</think>` span or a standalone JSON wrapper. The [official chat template](https://huggingface.co/thinkingmachines/Inkling-NVFP4/blob/main/chat_template.jinja#L97-L128) defines the block grammar. For readability, the examples below are wrapped at token boundaries and the JSON is indented; the actual decoded completion is one contiguous string with minified JSON.

With reasoning followed by visible text, the raw completion looks like:

```
<|content_thinking|>
I should answer directly.
<|end_message|>
<|message_model|>
<|content_text|>
The answer is 4.
<|end_message|>
<|content_model_end_sampling|>
```

With reasoning followed by a tool call, it looks like:

```
<|content_thinking|>
I should call get_weather.
<|end_message|>
<|message_model|>
get_weather
<|content_invoke_tool_json|>
{
  "name": "get_weather",
  "args": {
    "location": "Paris"
  }
}
<|end_message|>
<|content_model_end_sampling|>
```

When reasoning is disabled and the tool call is the first generated block, the completion starts directly with the redundant tool name:

```
get_weather
<|content_invoke_tool_json|>
{
  "name": "get_weather",
  "args": {}
}
<|end_message|>
```

That missing initial `<|message_model|>` is expected. With `add_generation_prompt=true`, the chat template appends `<|message_model|>` to the prompt, so the model begins generating immediately after it. Subsequent blocks emit their own `<|message_model|>` separator.

The relevant emitted tokens are:
- `<|message_model|>` starts the next model block; for the first block it is normally already present in the prompt.
- `<|content_thinking|>` marks reasoning that becomes `reasoning_content`.
- `<|content_text|>` marks user-visible assistant text that becomes `content`.
- `<|content_invoke_tool_json|>` starts a tool-call object. Inkling uses `{"name": ..., "args": {...}}`; the bare name before this marker is redundant, and the JSON `name` is authoritative.
- `<|end_message|>` closes one reasoning, text, or tool-call block.
- `<|content_model_end_sampling|>` terminates the sampled assistant turn when surfaced by the backend.

This PR mirrors that wire format with paired parsers. The reasoning parser strips framing, routes thinking and text to the correct response fields, and forwards complete tool blocks intact. The tool-call parser then extracts the JSON call, removes the redundant outer name/header, and supports multiple back-to-back blocks. Both parsers handle the canonical header-framed form and the header-less first-block form seen in live generation, including markers and JSON split across streaming chunks.

`parsers/src/reasoning/inkling_parser.rs`: the `Primed` arm of `run()` and its handoff to `Idle`.

#### Test Plan
- `cargo test --workspace`: 0 failed (parser tests plus 9 new inkling header-less / streaming tests).
- `conformance/utils/check.sh dynamo batch`: 655/655.
- `cargo clippy -p dynamo-parsers --all-targets -- -D warnings`: clean.
- Live e2e on a custom dynamo built against this branch: tool-call parses clean, `reasoning_content` populated iff reasoning enabled, zero `<|...|>` marker leaks in content or reasoning.

#### Related Issues
None.

#### Landing note
Requires ai-dynamo/dynamo#11823 (frontend detok preserves Inkling's special-token markers) to land no later than this. Without it, dynamo's frontend strips the channel markers before the parser and these fixes are inert.


parsers/v1/src/reasoning/inkling_parser.rs
```diff
@@ -0,0 +1,570 @@
...
+                            self.state = State::InReasoning;
+                        } else if rem.starts_with(CONTENT_TEXT) {
+                            self.buffer = self.buffer[rem_start + CONTENT_TEXT.len()..].into();
+                            self.state = State::InContent;
+                        } else if content_type_undecided(rem) {
+                            // Routing undecided; hold from `<|message_model|>`.
+                            self.buffer = self.buffer[pos..].into();
+                            break;
+                        } else {
+                            // Tool-call block: preserve verbatim with its header.
```
keivenchang: I think this else branch can leak framing for the non-tool block types — `CONTENT_MARKERS` declares `<|content_image|>` and `<|content_audio_input|>`, but the router only handles thinking/text, so a completed image block falls into `InToolBlock` and passes through verbatim, and the downstream tool parser (no `<|content_invoke_tool_json|>` inside) never strips it. Verified: input `<|message_model|><|content_image|>IMGDATA<|end_message|><|message_model|><|content_text|>done<|end_message|>` gives `normal_text` = `<|message_model|><|content_image|>IMGDATA<|end_message|>done`. You could either give those two markers an explicit stripped/dropped route, or restrict the `InToolBlock` fallthrough to blocks whose header is a bare NAME (tool shape).

parsers/v1/src/reasoning/inkling_parser.rs
```diff
@@ -0,0 +1,570 @@
...
+        .any(|t| t.len() > rem.len() && t.starts_with(rem))
+}
+
+impl InklingReasoningParser {
+    /// Drive the state machine, holding a trailing partial marker (or an undecided
+    /// post-`<|message_model|>` region) in `self.buffer` for the next chunk.
+    fn run(&mut self, reasoning: &mut String, normal: &mut String) {
+        loop {
+            match self.state {
+                State::Primed => {
```
keivenchang, line 161: One more in this state machine: `Primed` can hold forever when the turn starts with prose before the first `<|message_model|>`, since it only ever checks `starts_with` — `detect_and_parse_reasoning("Hello. <|message_model|><|content_thinking|>think<|end_message|>...")` returns the entire input (framing included) as `normal_text` with empty `reasoning_text`, and in streaming nothing is emitted until finish. Probably unreachable after a real prefill (the primer eats that position), but the batch API accepts arbitrary text. Handling `buffer.contains(MESSAGE_MODEL)` in `Primed` (emit the prefix, route the rest via `Idle`) would cover it — or pin the current behavior with a test if it's intentional.

parsers/v1/src/tool_calling/inkling/parser.rs
```diff
@@ -0,0 +1,286 @@
...
+    // it never leaks into normal_text.
+    let header_pos = message[..invoke_pos].rfind(MESSAGE_MODEL);
+    let block_start = header_pos.unwrap_or(invoke_pos);
+    let mut prefix = message[..block_start].trim_end().to_string();
+
+    let mut calls = Vec::new();
+    let mut cursor = block_start;
+    while let Some(rel) = message[cursor..].find(INVOKE) {
+        let json_start = cursor + rel + INVOKE.len();
+        let (json_str, next_cursor) = match message[json_start..].find(END_MESSAGE) {
```
keivenchang: Not sure this shape can come out of real token ids, but at the text level this `find` truncates the JSON when `<|end_message|>` appears inside a string argument — `{"name":"echo","args":{"text":"a<|end_message|>b"}}` parses to zero calls AND empty `normal_text` (call silently dropped, surrounding text swallowed). Scanning for the fence only at a JSON-complete boundary would fix it (the streaming deserializer already reports the consumed length); otherwise a test pinning the drop would at least make the tradeoff explicit.

parsers/v1/src/reasoning/inkling_parser.rs
```diff
@@ -0,0 +1,570 @@
...
+            State::InReasoning => ParserResult {
+                reasoning_text: buffered,
+                normal_text: String::new(),
+            },
+            State::InContent | State::InToolBlock => ParserResult {
+                normal_text: buffered,
+                reasoning_text: String::new(),
+            },
+            // Idle leftover is a held partial marker: parser-owned markup, so drop it.
+            State::Idle => ParserResult::default(),
```
keivenchang: nit: a stream whose final text legitimately ends in `<` or `<|` loses that suffix — the Idle finish drops any held partial-marker bytes. Streaming `["<|message_model|><|content_text|>hi<|end_message|>", "tail <|"]` finishes as `hitail ` (the `<|` is gone). Worth a test pinning whichever behavior you want.

parsers/v1/src/reasoning/inkling_parser.rs
```diff
@@ -0,0 +1,570 @@
...
+//! Reasoning is block-structured, not a `<think>` prefix:
+//! `<|message_model|><|content_thinking|>REASONING<|end_message|><|message_model|><|content_text|>ANSWER<|end_message|>`.
+//! `<|content_thinking|>` blocks route to `reasoning_text`, `<|content_text|>` to
+//! `normal_text`, with framing stripped. A tool-call block is passed through verbatim
+//! into `normal_text` (framing intact): the reasoning parser runs before the tool-call
+//! parser, which extracts calls from that `normal_text`.
+
+use crate::{ParserResult, ReasoningParser};
+
+const MESSAGE_MODEL: &str = "<|message_model|>";
```
keivenchang: nit: these token constants duplicate `MESSAGE_MODEL` / `<|content_invoke_tool_json|>` / `END_MESSAGE` in `tool_calling/inkling/parser.rs` — a single shared inkling token module that both parsers import would keep the two copies from drifting.

conformance/fixtures-manifest.json
```diff
@@ -32,18 +32,18 @@
     },
     {
       "path": "toolcalling/fixtures-batch-v1/dynamo_v1-3.0.0.tar.gz",
-      "sha256": "65290c8070e6a5f7b50665375a99335257625419b672fc2379bb77506b2f14b2",
-      "size": 10667
+      "sha256": "673a5ef465a91c8c1f1af8a23f91b6bc87cdc47404017b1c65ac0acce7558c5a",
+      "size": 11228
     },
     {
       "path": "toolcalling/fixtures-batch-v1/inputs.tar.gz",
```
keivenchang, line 49: Concrete slots to fill in the matrix — inkling currently has batch.{1, 1.a, 2, 3, 4, 5, 5.c, 7, 9.a, 9.b, 10, 13, 13.c} and is missing groups 6, 8, 30 that other families cover. Input model_text -> expected, capture-pinned against vLLM/SGLang:

**TOOLCALLING.batch.4.b — Two DIFFERENT tools back-to-back** (your batch.4 and batch.10 both reuse get_weather):
input: `<|message_model|>get_weather<|content_invoke_tool_json|>{"name":"get_weather","args":{"location":"Paris"}}<|end_message|><|message_model|>book_flight<|content_invoke_tool_json|>{"name":"book_flight","args":{"destination":"SFO"}}<|end_message|>`
expected: calls=[get_weather, book_flight], normal_text=''

**TOOLCALLING.batch.6.a / 6.b — canonical empty args** (group 6 exists for every other family, missing for inkling):
6.a input: `<|content_invoke_tool_json|>{"name":"f","args":{}}<|end_message|>` -> calls=[f({})], normal_text=''
6.b input: `<|content_invoke_tool_json|>{"name":"f","args":null}<|end_message|>` -> calls=[f({})], normal_text=''

**TOOLCALLING.batch.7.a — JSON-edge argument payload** (escaped quotes + non-ASCII; argument bytes must survive verbatim; your bare batch.7 covers typed args only):
input: `<|content_invoke_tool_json|>{"name":"echo","args":{"text":"say \"hi\" — héllo"}}<|end_message|>`
expected: calls=[echo({"text":"say \"hi\" — héllo"})], normal_text=''

**TOOLCALLING.batch.8.b / 8.d — narration around calls** (group 8 = narration before/after/between; missing entirely for inkling — your batch.5 covers only the before case; dropped-after/between today matches the dsml/vLLM convention, the fixture makes it explicit):
8.b input: `<|message_model|>a<|content_invoke_tool_json|>{"name":"a","args":{}}<|end_message|>TRAILING` -> calls=[a], normal_text=''
8.d input: `<|message_model|>a<|content_invoke_tool_json|>{"name":"a","args":{}}<|end_message|>MIDDLE<|message_model|>b<|content_invoke_tool_json|>{"name":"b","args":{}}<|end_message|>` -> calls=[a, b], normal_text=''

**TOOLCALLING.batch.11.a / 11.b — name edge cases** (11 is the next free group; inkling-specific because of the redundant NAME header):
11.a header/JSON disagree — input: `<|message_model|>get_weather<|content_invoke_tool_json|>{"name":"other_tool","args":{}}<|end_message|>` -> today JSON silently wins: calls=[other_tool({})], normal_text='' — worth capturing what vLLM does here
11.b empty name — input: `<|content_invoke_tool_json|>{"name":"","args":{"x":1}}<|end_message|>` -> calls=[], normal_text=''

**TOOLCALLING.batch.30.a — own marker inside an argument string value** (group 30 is exactly this for other families, e.g. mistral pins its wrapper marker inside a string arg):
input: `<|message_model|>echo<|content_invoke_tool_json|>{"name":"echo","args":{"text":"a<|end_message|>b"}}<|end_message|>`
today: calls=[], normal_text='' (call silently dropped AND text swallowed) — strong divergence-documentation candidate, capture what vLLM/SGLang emit.

conformance/fixtures-manifest.json
```diff
@@ -12,8 +12,8 @@
   "shards": [
     {
       "path": "reasoning/fixtures-v1/inputs.tar.gz",
```
keivenchang, line 14: Concrete slots — inkling REASONING currently has batch.{1.a, 1.b, 2.a, 2.b, 2.c, 3.a, 5, 7} and stream.{1.a, 1.b, 2.a, 3.a, 3.b, 3.c}. Suggested additions, input model_text -> expected:

**REASONING.batch.1.c — empty block**:
input: `<|message_model|><|content_thinking|><|end_message|>` -> reasoning_text='', normal_text=''

**REASONING.batch.1.d — prose before the first `<|message_model|>`** (today dynamo returns the ENTIRE input raw as normal_text, framing included; vLLM drops pre-content-marker text — a divergence-documentation case in the same spirit as your 1.b explanation):
input: `Hello. <|message_model|><|content_thinking|>think<|end_message|><|message_model|><|content_text|>answer<|end_message|>`

**REASONING.batch.3.b — `<|content_image|>` block routing** (today dynamo leaks the framing verbatim into normal_text — see my inline comment on the router; the fixture pins whatever the fix decides, against vLLM):
input: `<|message_model|><|content_image|>IMGDATA<|end_message|><|message_model|><|content_text|>done<|end_message|>`

**REASONING.batch.4.a — content terminated by `<|content_model_end_sampling|>` alone, no `<|end_message|>`** (group 4 = dangling end markers in other families; the code handles this via find_earliest but nothing pins it):
input: `<|message_model|><|content_text|>answer<|content_model_end_sampling|>` -> reasoning_text='', normal_text='answer'

**REASONING.batch.6.a — two thinking blocks in one turn** (group 6 = multiple thought spans, granite/harmony have it; concatenated with no separator today):
input: `<|message_model|><|content_thinking|>a<|end_message|><|message_model|><|content_thinking|>b<|end_message|>` -> reasoning_text='ab', normal_text=''

**REASONING.stream.3.d — final chunk ending in a bare `<` / `<|`** (today that suffix is silently dropped on finish):
chunks: `["<|message_model|><|content_text|>hi<|end_message|>", "tail <|"]`
expected today: normal_text='hitail ' (the `<|` is gone) — pin whichever behavior is intended.

conformance/fixtures-manifest.json
rmccorm4, line 1: Two notes:
1. Seems like a lot of tool call batch test cases are missing in parity table
2. Seems like all tool call stream test cases are missing in parity table

rmccorm4: https://github.com/ai-dynamo/frontend-crates/pull/120#issuecomment-5001211747

parsers/v1/src/tool_calling/inkling/parser.rs
```diff
@@ -0,0 +1,247 @@
...
+        }
+    }
+    false
+}
+
+pub fn find_tool_call_end_position_inkling(chunk: &str, _config: &InklingParserConfig) -> usize {
+    chunk
+        .rfind(END_MESSAGE)
+        .map(|pos| pos + END_MESSAGE.len())
+        .unwrap_or(chunk.len())
```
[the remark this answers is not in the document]
KrishnanPrash: b69c915 - Return `None` until serde finds a complete JSON value followed by the real outer end marker, with incomplete and marker-in-string regression tests.

parsers/v1/src/tool_calling/config.rs
```diff
@@ -356,6 +366,9 @@ pub enum ParserConfig {
     /// delimiters, and fixed `<|tool_call>...<tool_call|>` markers. No
     /// configuration is required at runtime — markers are not user-tunable.
     Gemma4,
+    /// Inkling uses `<|message_model|>NAME<|content_invoke_tool_json|>{...}<|end_message|>`
+    /// framing around a `{"name":..., "args":{...}}` JSON object.
+    Inkling(InklingParserConfig),
```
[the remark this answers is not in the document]
KrishnanPrash: b69c915 - Kept the crate at 5.1.2 in this feature PR because the release-plz plus cargo-semver-checks workflow detects this public API break and publishes the required 6.0.0 release.

parsers/v1/src/tool_calling/inkling/parser.rs
```diff
@@ -0,0 +1,286 @@
...
+        if chunk.contains(marker) || ends_with_partial_marker(chunk, marker) {
+            return true;
+        }
+    }
+    false
+}
+
+fn ends_with_partial_marker(chunk: &str, marker: &str) -> bool {
+    for (i, _) in marker.char_indices().skip(1) {
+        if i < MIN_PARTIAL_MARKER_LEN {
```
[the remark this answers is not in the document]
KrishnanPrash: b69c915 - Detect every non-empty partial opener including `<` and `<|`, with exhaustive jail tests over every single split boundary.

parsers/v1/src/reasoning/inkling_parser.rs
KrishnanPrash: b69c915 - Added explicit image and audio discard routing so their payloads and framing never reach reasoning or normal text.

parsers/v1/src/reasoning/inkling_parser.rs
KrishnanPrash: b69c915 - `Primed` now hands prose before a real header to `Idle`, with batch, streaming, prose-to-tool, and precedence regression coverage.

parsers/v1/src/tool_calling/inkling/parser.rs
KrishnanPrash: b69c915 - Replaced lexical fence scans with one serde-derived JSON boundary shared by reasoning, tool parsing, and the jail, preserving marker literals inside arguments.

parsers/v1/src/reasoning/inkling_parser.rs
KrishnanPrash: b69c915 - Preserve ambiguous trailing `<` and `<|` as text while dropping longer committed protocol fragments, with EOF regression tests.

parsers/v1/src/reasoning/inkling_parser.rs
KrishnanPrash: b69c915 - Moved shared Inkling tokens into `inkling/tokens.rs` and imported them from both parser stages.

conformance/fixtures-manifest.json
KrishnanPrash: b69c915 - Expanded Inkling tool batch coverage to 40 authored cases across the missing groups, with current Dynamo captures and conservative peer N/A overlays.

conformance/fixtures-manifest.json
KrishnanPrash: b69c915 - Expanded Inkling reasoning coverage to 20 batch and 13 stream cases, including empty, multi-span, non-text, dangling, and split-marker paths.

conformance/fixtures-manifest.json
KrishnanPrash: b69c915 - Added 30 authored Inkling tool stream cases and current Dynamo capture shards so the stream row renders in the conformance table.

keivenchang (approved):
Thanks Krishnan — approving, these are all minor/non-blocking. Just noting the Inkling fixtures use their own filler strings/args instead of the shared per-case payloads, so the Inkling rows don't line up with the other families in the matrix. Worth normalizing when convenient:

Tool-calling (batch):
- `7.b` (unicode) — `echo(text="say \"hi\" — héllo")` -> standard is `get_weather(location="Tōkyō \"central\"")` (the standout)
- `1` (happy path) — `location:"Paris"` plus an extra `unit:"celsius"` -> standard is just `location:"NYC"`
- multi-call (`2.a`/`2.b`/`2.c`/`5.d`/`5.e`) — second tool `book_flight`/`destination`/`SFO`; the precedent (nemotron_deci) is `get_time`

Reasoning (batch):
- `2.f` — `héllo — 数学` / `line two` / `done` -> shared payload is `Line 1` / `Unicode: café` / `JSON-ish: {"x": [1, true]}` / `Done.`
- `6.a` spans `a`/`b` -> `first`/`second`; `6.b` `a` -> `preamble`/`body`/`answer`
- `2.c` `reason`/`answer` -> `compute it` / `The answer is 42.`; `2.d` reasoning filler `reason` -> `thinking`
- `2.e` — add a trailing `Direct answer.` as normal_text

All fixture-payload only, nothing blocks.

PR #129, opened 2026-07-17
chore(renderer)!: release v2.0.0 with dynamo-protocols v3

Chokoyo:
## Why

[`dynamo-protocols` 3.0.0](https://crates.io/crates/dynamo-protocols/3.0.0) introduced cached multimodal UUID content parts in [#119](https://github.com/ai-dynamo/frontend-crates/pull/119), while the latest published [`dynamo-renderer` 1.3.12](https://crates.io/crates/dynamo-renderer/1.3.12) still depends on `dynamo-protocols` 2.0.2.

`dynamo-renderer` exposes `dynamo-protocols` message types through its public `OAIChatLikeRequest` API. Mixing renderer on protocols v2 with consumers on protocols v3 therefore creates distinct, incompatible Rust types and prevents the dependency graph from compiling. Since this changes a public dependency type, the renderer requires a major-version release.

Publishing this release helps unblock [ai-dynamo/dynamo#8957](https://github.com/ai-dynamo/dynamo/pull/8957), allowing Dynamo to replace its temporary frontend-crates git pin with published crates while supporting vLLM cached multimodal UUID passthrough.

## What Change

- Bump `dynamo-renderer` from 1.3.12 to 2.0.0.
- Release the renderer against the workspace's `dynamo-protocols` 3.0.0.
- Update the workspace dependency, lockfile, and renderer changelog.

There are no renderer behavior changes in this PR; the major bump reflects the public protocol-type dependency change.

## Test Plan

- `cargo clippy -p dynamo-renderer --all-targets -- -D warnings`
- `cargo test -p dynamo-renderer` — 104 passed, 1 ignored
- `cargo check --workspace --all-targets`
- `cargo package -p dynamo-renderer --allow-dirty --no-verify`
- Verified the packaged renderer manifest depends on `dynamo-protocols` 3.0.0

### DIFF
--- a/renderer/CHANGELOG.md
+++ b/renderer/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+## [2.0.0](https://github.com/ai-dynamo/frontend-crates/compare/dynamo-renderer-v1.3.12...dynamo-renderer-v2.0.0) - 2026-07-17
+
+### Miscellaneous
+
+- Updated the following local packages: dynamo-protocols
+
 ## [1.3.12](https://github.com/ai-dynamo/frontend-crates/compare/dynamo-renderer-v1.3.11...dynamo-renderer-v1.3.12) - 2026-07-14
 
 ### Miscellaneous
--- a/renderer/CHANGELOG.md
+++ b/renderer/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [2.0.0](https://github.com/ai-dynamo/frontend-crates/compare/dynamo-renderer-v1.3.12...dynamo-renderer-v2.0.0) - 2026-07-17
 
+### Changed
+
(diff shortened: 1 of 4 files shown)


renderer/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+## [2.0.0](https://github.com/ai-dynamo/frontend-crates/compare/dynamo-renderer-v1.3.12...dynamo-renderer-v2.0.0) - 2026-07-17
```
keivenchang, line 7: Hey Jesse, tiny nit — this entry reads like a routine dep-patch line, but 2.0.0 is the breaking release. I think a one-line note under a Changed/Breaking heading (public request types now come from dynamo-protocols 3.x, so consumers must upgrade protocols too) would help crates.io users see why the major bump. The PR description already explains it well — this just gets it into the changelog consumers actually read. Fine either way.

keivenchang (approved):
hey Jesse, thanks so much for this — just gave it a review. Verified the packaged renderer picks up protocols 3.0.0 via the workspace dep, parsers 5.1.0 already requires ^3.0.0, so this is the last missing link for the dynamo unblock. One optional changelog nit inline, non-blocking. LGTM

renderer/CHANGELOG.md
Chokoyo: Thanks Keiven! I updated this changelog with the breaking headings and explaination

PR #133, opened 2026-07-18
fix(parsers): Harden MiniMax M3 parser

indrajit96:
## Summary

Strengthens the MiniMax M3 reasoning parser for responses that begin a native tool call without first emitting `</mm:think>`.

The parser now treats `]<]minimax[>[` as an alternate reasoning boundary and forwards the marker losslessly to the existing tool-call parser. Normal responses containing `</mm:think>` remain unchanged.

Adds one-shot and streaming regression coverage, including namespace markers split across chunks.

This is in response to a user reporting a `]<]minimax[>[` leak in reasoning content.

## Streaming behavior note

A fresh MiniMax M3 streaming parser has no authoritative state until it sees `<mm:think>`, `</mm:think>`, the native tool namespace, or EOF. It now buffers that undecided prefix so a later boundary can classify the bytes correctly.

This intentionally changes existing incremental behavior: unprimed plain text and a partial reasoning-start prefix that previously streamed as normal content may now be emitted only when a decisive boundary or EOF arrives. Callers that know the initial state should call `set_in_reasoning(true)` for prompt-prefilled reasoning or `set_in_reasoning(false)` for a known-normal stream to preserve per-chunk deltas. Dynamo text-chat requests normally provide the reasoning state; the remaining unknown-state case is token-input chat and out-of-tree parser use.

## Validations

- Added unit coverage for:
  - Missing `</mm:think>` before the MiniMax native tool namespace.
  - Both `minimax_m3` and `minimax-m3` parser aliases.
  - Streaming responses where `]<]minimax[>[` arrives in one chunk or is split across chunks.
  - Lossless forwarding of the complete namespace to the tool parser.

- Replayed the boundary-stress request 512 times against the fixed image:
  - 506 structured tool-call responses.
  - Zero namespace, tool-wrapper, or invocation-marker leaks.
  - Zero SSE parsing errors.

- Ran 67 broader E2E requests covering thinking modes, forced/named tool calls, structured arguments, parallel calls, multi-turn calls, and streaming/non-streaming responses. No marker leaks or tool-call schema errors were observed.
- Full validation:
  - 751 parser tests passed, 5 ignored.
  - 48 Dynamo postprocessor tests passed.
  - Formatting and Clippy checks passed.

## Tracking

- Linear: [DIS-2509](https://linear.app/nvidia/issue/DIS-2509/harden-minimax-m3-parser)


parsers/v1/src/reasoning/mod.rs
```diff
@@ -301,7 +306,8 @@ impl ReasoningParserType {
                         false,
                         true,
                     )
-                    .with_dangling_end_recovery(),
+                    .with_dangling_end_recovery()
+                    .with_tool_start_token(MINIMAX_M3_TOOL_NAMESPACE),
```
keivenchang, line 308: confirmed dynamo-review-agent's finding here - repro'ed on this head.

Chunks `["reasoning]", "<]minimax[>[<tool_call>"]` give:

```
reasoning_text = "reasoning]<]minimax[>[<tool_call>"
normal_text    = ""
```

which is the exact leak this PR fixes, just at a different chunk boundary. Cause is the `ol >= 2` gate in `base_parser.rs:367` - a 1-char overlap gets flushed into reasoning and destroys the marker before the next chunk arrives.

I think this is worth fixing in the shared helper rather than for M3 only, because the same boundary already leaks for the sibling parsers:

```
minimax_m2  ["reasoning<", "minimax:tool_call>x"]        -> reasoning="reasoning<minimax:tool_call>x", normal=""
kimi_k2.5   ["reasoning<", "|tool_calls_section_begin|>x"] -> reasoning="reasoning<|tool_calls_section_begin|>x", normal=""
```

I tried gating the tool-token overlap at `>= 1` while leaving `ol_end` at `>= 2`, and it fixes all three with the full suite still green (671 + 75 + 13 passing).

parsers/v1/src/reasoning/mod.rs
```diff
@@ -466,6 +472,72 @@ mod tests {
...
+                    "{MINIMAX_M3_TOOL_NAMESPACE}<tool_call>\n\
+                     {MINIMAX_M3_TOOL_NAMESPACE}<invoke name=\"get_weather\">"
+                )
+            );
+        }
+    }
+
+    #[test] // MiniMax M3
+    fn test_minimax_m3_native_tool_namespace_is_lossless_across_stream_chunks() {
+        let chunkings = [
```
keivenchang: Since this test is the coverage claim for split markers, could you add the 1-byte boundary too - `["reasoning]", "<]minimax[>[", "<tool_call>"]`? That's the split that still fails today; `]<]mini` / `max[>[` passes because the overlap is >= 2.

parsers/v1/src/reasoning/mod.rs
```diff
@@ -21,6 +21,11 @@ pub use minimax_append_think_parser::MiniMaxAppendThinkParser;
 /// `KimiK2ParserConfig::default().section_start` in `crate::tool_calling::config`.
 pub(crate) const KIMI_K2_TOOL_SECTION_BEGIN: &str = "<|tool_calls_section_begin|>";
 
+/// MiniMax M3 namespace token that prefixes its native XML tool-call tags.
+/// The reasoning parser treats this as an alternate boundary because the model
+/// can begin a tool call without first emitting `</mm:think>`.
+pub(crate) const MINIMAX_M3_TOOL_NAMESPACE: &str = "]<]minimax[>[";
```
keivenchang: nit: this duplicates `MiniMaxM3ParserConfig::default().namespace_token` (`tool_calling/config.rs:335`). `KIMI_K2_TOOL_SECTION_BEGIN` right above carries a "Keep in sync with `KimiK2ParserConfig::default().section_start`" note - worth the same line here so the two copies don't drift.

keivenchang (commented):
hey Indrajit, thanks for this - just gave it a review.

I repro'ed, some issues - details inline. The dynamo-review-agent finding is real, and it hits minimax_m2 and kimi_k2.5 today too, so I'd fix it in the shared helper.

Two smaller things:

1. Both new tests call `set_in_reasoning(true)`, so the case `with_dangling_end_recovery` exists for - namespace arrives and the caller never set in_reasoning - is untested and still leaks. Today `detect_and_parse_reasoning("reasoning]<]minimax[>[<tool_call>")` on a fresh parser returns `reasoning_text=""`, `normal_text="reasoning]<]minimax[>[<tool_call>"`, so the reasoning prefix ends up in content. Same in streaming. If `</mm:think>` without an opener implies an implicit opener, I think the namespace should too - or is the pre-filled template guaranteed to always set it?

2. Title is `chore(parsers):` but this is a user-reported leak fix. release-plz's `commit_parsers` puts `chore` under Miscellaneous and `fix` under Bug fixes, so `fix(parsers):` would land it in the right changelog section.

rmccorm4:
@KrishnanPrash to help address @keivenchang 's comments

parsers/v1/src/reasoning/mod.rs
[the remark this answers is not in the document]
KrishnanPrash: c08a369 - The shared streaming parser now preserves one-byte tool-marker overlaps, with regression coverage for MiniMax M2, Kimi K2.5, and MiniMax M3.

KrishnanPrash: c08a369 - Updated the shared helper to buffer one-byte tool-marker overlaps while keeping the existing two-byte gate for think markers.

parsers/v1/src/reasoning/mod.rs
KrishnanPrash: c08a369 - Added the exact reasoning] split and exhaustive coverage for every byte boundary in the MiniMax M3 namespace.

parsers/v1/src/reasoning/mod.rs
KrishnanPrash: c08a369 - Centralized the MiniMax M3 namespace token in tool_calling/config.rs and reused it from the config, reasoning registration, and tests.

parsers/v1/src/reasoning/base_parser.rs
```diff
@@ -416,30 +467,51 @@ impl ReasoningParser for BasicReasoningParser {
...
-                        }
+                        accumulated_normal.push_str(&current_text[..safe_end]);
+                        self.recover_tool_start_without_opener = false;
                     }
                     self._buffer = current_text[safe_end..].to_string();
                 } else {
                     accumulated_normal.push_str(&current_text);
                     self._buffer.clear();
+                    if !current_text.is_empty() {
+                        self.recover_tool_start_without_opener = false;
```
keivenchang: this line is where the implicit recovery dies - any non-empty chunk with no marker overlap clears the flag for good, so recovery only survives when the namespace is in the first chunk or overlaps its tail.

repro'ed at c08a369. same M3 response (prompt-prefilled opener, so no `<mm:think>` in the output), four ways it can arrive over the wire:

```
A  ["reasoning]<]minimax[>[<tool_call>"]                       one chunk
B  ["reasoning]", "<]minimax[>[<tool_call>"]                   split inside the marker
C  ["reasoning", "]<]minimax[>[<tool_call>"]                   split right before the marker
D  ["let ", "me ", "think", "]<]minimax[>[", "<tool_call>"]    per-token
```

EXPECTED for all four - this is what `detect_and_parse_reasoning` already returns for every one of them:

```
reasoning_text = "reasoning"                 ("let me think" for D)
normal_text    = "]<]minimax[>[<tool_call>"
```

ACTUAL from `parse_reasoning_streaming_incremental`:

```
A  reasoning_text="reasoning"  normal_text="]<]minimax[>[<tool_call>"                ok
B  reasoning_text="reasoning"  normal_text="]<]minimax[>[<tool_call>"                ok
C  reasoning_text=""           normal_text="reasoning]<]minimax[>[<tool_call>"       reasoning leaks into content
D  reasoning_text=""           normal_text="let me think]<]minimax[>[<tool_call>"    same
```

A and B are the two shapes the new test covers, and they're the two where the first chunk either contains the marker or ends on a 1-byte overlap of it. C is what a real stream looks like - reasoning is many tokens and the namespace arrives as its own token after them - and there the first chunk is plain text with zero overlap, so it flushes as normal and takes the flag with it.

I think you can either keep the flag alive until an explicit boundary decides it (`<mm:think>`, `</mm:think>`, or a tool marker) and hold the pending normal text until then, or accept that implicit recovery can't work in streaming and lean on `set_in_reasoning(true)` for it. the first costs streaming latency, the second leaves batch and stream disagreeing. either way `["reasoning", "]<]minimax[>[<tool_call>"]` in the recovery test would pin whichever you pick.

parsers/v1/src/reasoning/base_parser.rs
```diff
@@ -416,30 +467,51 @@ impl ReasoningParser for BasicReasoningParser {
...
+                    // An implicit close/tool boundary determines whether every
+                    // preceding byte is reasoning. Keep the whole undecided
+                    // prefix until the next chunk confirms or rejects it;
+                    // emitting the prefix now would make a marker fakeout
+                    // impossible to restore as normal text.
+                    if self.recover_dangling_end
+                        && ol_boundary > ol_start
+                        && (ol_end >= 2 || ol_tool >= 1)
+                    {
+                        break;
```
keivenchang, line 511: minor: this `break` withholds the whole pending buffer rather than just the ambiguous tail, so M3 emits nothing across a run of chunks ending in `]`.

input `["see item [1]", " and [2]", " and [3]", " done."]` on a fresh minimax_m3 parser:

```
EXPECTED (what kimi_k2.5 does on the same input, one chunk in -> one chunk out):
  1  normal_text="see item [1]"
  2  normal_text=" and [2]"
  3  normal_text=" and [3]"
  4  normal_text=" done."

ACTUAL:
  1  normal_text=""
  2  normal_text=""
  3  normal_text=""
  4  normal_text="see item [1] and [2] and [3] done."
```

content is correct in the end and `finish_reasoning_stream` flushes the tail, so this is smoothness not correctness - just flagging because `]` is a common token ending in code and markdown answers.

parsers/v1/src/reasoning/base_parser.rs
```diff
@@ -121,17 +126,21 @@ impl BasicReasoningParser {
         self
     }
 
-    /// Enables streaming dangling-end recovery: a close marker without a visible
-    /// opener routes the preceding text to reasoning instead of normal text.
+    /// Enables streaming dangling-close recovery and implicit tool-marker
+    /// recovery in both parsing paths.
     pub fn with_dangling_end_recovery(mut self) -> Self {
```
keivenchang, line 133: nit: this setter now switches on two independent behaviors and the name only covers one. M3 is the only caller today, but the next parser that wants dangling-close recovery picks up implicit tool-start recovery silently. maybe a second builder method, or rename to something like `with_implicit_reasoning_recovery`.

keivenchang (commented):
almost there! re-reviewed at c08a369 - shared-helper fix, the m2/kimi/m3 rstest and the all-split-positions loop all look good.

one gap left: the implicit recovery only fires when the namespace lands in the first chunk or overlaps its tail, so a real token stream still leaks. details inline, plus two smaller notes.

parsers/v1/src/reasoning/base_parser.rs
KrishnanPrash: 727835b - Keeps an unprimed M3 stream buffered until a start, close, tool boundary, or EOF decides its state, with tests for standalone and per-token namespace chunks.

parsers/v1/src/reasoning/base_parser.rs
KrishnanPrash: 727835b - Keeps fresh unprimed text buffered for correct implicit recovery, while set_in_reasoning(false) preserves immediate one-chunk-in, one-chunk-out normal streaming.

parsers/v1/src/reasoning/base_parser.rs
KrishnanPrash: 727835b - Splits dangling-close recovery and implicit tool-start recovery into separate builder methods and adds an independence test.

parsers/v1/src/reasoning/base_parser.rs
```diff
@@ -416,25 +479,41 @@ impl ReasoningParser for BasicReasoningParser {
...
                     continue;
                 }
 
+                // With implicit tool-start recovery, all bytes before the first
+                // decisive boundary are ambiguous. A future chunk may contain
+                // the tool marker and retroactively establish that the entire
+                // prefix was prompt-prefilled reasoning. Because streaming
+                // deltas cannot retract previously emitted normal text, retain
+                // the complete prefix until a boundary or EOF decides it.
+                if self.recover_tool_start_without_opener {
```
keivenchang, line 492: one residual, non-blocking. with the flag on nothing is emitted until a boundary or EOF, but I chased the consumer and dynamo mostly can't hit it - `thinking_mode=disabled` sets `should_parse_reasoning=false` so the parser never runs (`preprocessor.rs:2233`), and every other M3 chat request prefills `<mm:think>` so `set_in_reasoning(true)` fires (`:3150`, `:3027`).

what's left is chat with token input, where `apply_template` returns `None` and `prompt_injected_reasoning` is false, plus out-of-tree users of the published crate. those lose the reasoning delta stream:

```
["let ", "me ", "think", "</mm:think>", "the ", "answer"], no set_in_reasoning

ACTUAL:    1-3 emit nothing, 4 emits reasoning_text="let me think" in one blob, content streams after
EXPECTED:  reasoning_text streams per chunk (what set_in_reasoning(true) gives)
```

you documented that on the builder so I'm fine shipping it. cheap follow-up whenever: have dynamo call `set_in_reasoning(false)` in the else branch so the escape hatch your `..._streams_each_chunk_immediately` test documents is actually reachable.

parsers/v1/src/reasoning/mod.rs
```diff
@@ -536,17 +801,19 @@ mod tests {
     }
 
     #[test] // MiniMax M3
-    fn test_minimax_m3_streaming_partial_start_prefix_becomes_normal_text() {
+    fn test_minimax_m3_streaming_partial_start_prefix_becomes_normal_text_at_eof() {
```
keivenchang, line 804: worth calling out in the PR description: this rename is an existing behavior change, not new coverage - partial-start-prefix text used to stream out incrementally and now lands at EOF. same root cause as the other comment.

keivenchang (approved):
minor, just 1 requested change, pre-approved

parsers/v1/src/reasoning/base_parser.rs
KrishnanPrash: 727835b - Confirmed locally that an unprimed stream batches reasoning until the boundary, true streams reasoning per chunk, and false streams the prefix as normal; documented this unknown-state limitation in the PR description.

parsers/v1/src/reasoning/mod.rs
KrishnanPrash: 727835b - Updated the PR description to call out that unprimed plain text and partial start prefixes now wait for a decisive boundary or EOF.

PR #134, opened 2026-07-20
fix(tokenizers): support Kimi Linear tokenizer

rmccorm4:
## Summary

- Recognize `model_type = "kimi_linear"` in TikToken auto-detection.
- Reuse the existing Kimi BPE regex used by the other Kimi model types.
- Add focused regression coverage for the previously rejected model type.

## Root cause

`TikTokenTokenizer::from_file_auto` selects its BPE regex from `config.json`. Kimi Linear ships a `tiktoken.model` with `model_type = "kimi_linear"`, but that value was absent from the supported Kimi model-type allowlist, so loading failed before tokenizer construction.

This PR is intentionally limited to model-type recognition. It does not change special-token handling or renderer/chat-template behavior.

## Validation

- `cargo test -p dynamo-tokenizers` (52 unit tests and 8 integration tests passed)
- `cargo clippy -p dynamo-tokenizers --all-targets -- -D warnings`

## Tracking

- [DIS-2465](https://linear.app/nvidia/issue/DIS-2465/support-kimi-linear-tokenizer-model-type)


rmccorm4:
From @mispa-ms who requested this support:

> I got a smoke test of Kimi-Linear 1P1D disagg working. With the tokenizer fix, the Dynamo frontend registers cleanly, and the KDA/NIXL state transfer completes successfully with 0/83 request errors.
>
> I also opened an srt-slurm[ PR ](https://github.com/NVIDIA/srt-slurm/pull/280) for the reusable change that made this testable—allowing Dynamo to be built against an unmerged crate branch on the fly.

indrajit96 (approved):
LGTM!

PR #139, opened 2026-07-23
perf(tokenizers): prototype Gigatoken backend

rmccorm4:
## Summary

This is an experimental, non-merge-ready prototype of [marcelroed/gigatoken](https://github.com/marcelroed/gigatoken) in `dynamo-tokenizers`.

- add an optional `gigatoken` encoder backend for Hugging Face byte-level BPE JSON and native TikToken rank files
- preserve Dynamo's existing HF/TikToken decoding paths
- require exact token-ID parity before benchmarking
- add a reusable Criterion benchmark for Qwen3-0.6B, Kimi-K2.6, GLM-5.2, DeepSeek-V4 Pro, and MiniMax-M3 at 128, 512, 2K, 8K, and 32K tokens

Linear: [DIS-2494](https://linear.app/nvidia/issue/DIS-2494/prototype-and-benchmark-gigatoken-tokenizer-backend)

## Performance

Single pinned logical CPU, `RAYON_NUM_THREADS=1`, fresh process per model/length, Criterion 3 s warmup + 7 s measurement target with 60 samples. The host was shared and running the powersave governor, so treat absolute throughput as directional; the within-process ratios are the primary result.

| Model | HF speed | TikToken speed | Fastokens speed | Gigatoken speed | Gigatoken vs reference | Gigatoken vs Fastokens |
|---|---:|---:|---:|---:|---:|---:|
| Qwen3-0.6B | 0.43–0.74 | — | 4.94–8.63 | 56.12–89.10 | 75.4–199.2× | 10.0–11.4× |
| Kimi-K2.6 | — | 1.08–1.21 | — | 58.28–80.58 | 49.0–72.4× | — |
| GLM-5.2 | 0.48–0.86 | — | 4.72–8.40† | 39.80–65.18 | 46.4–136.9× | 7.5–8.4×† |
| DeepSeek-V4 Pro | 0.50–0.80 | — | 3.53–5.76† | 30.78–44.74 | 38.6–89.2× | 7.3–8.7×† |
| MiniMax-M3 | 0.48–0.76 | — | 3.95–7.86 | 54.67–86.00 | 72.3–174.1× | 10.7–13.9× |

All speed columns are Mtok/s and show the range across 128–32K input tokens. † Fastokens completed four of five GLM lengths and three of five DeepSeek lengths because of PCRE2 panics.

Across 67 measured backend/length cells, Gigatoken sustained 30.8–89.1 Mtok/s: 38.6–199.2× the official reference backend and 7.3–13.9× Fastokens where Fastokens completed. Gigatoken produced exact token IDs for all 175 measured inputs.

Kimi publishes a native `tiktoken.model` rather than a compatible HF tokenizer JSON, so its official Dynamo TikToken path is the reference and Fastokens is not applicable.

## Findings and limitations

- The advantage persists at every tested length. Qwen, Kimi, and MiniMax reach roughly 76–89 Mtok/s at long lengths; GLM plateaus around 58–65 Mtok/s; DeepSeek is the hardest vocabulary/pattern at roughly 42–45 Mtok/s.
- Fastokens is much closer to Gigatoken than the official HF path when it works, but its PCRE2 splitter panicked for GLM at 8K and DeepSeek at 8K/32K. GLM 32K completed, so the failure is not a simple monotonic length limit.
- These are hot, single-core encoder measurements. Cold first encode, concurrent throughput, memory/RSS, multi-core scaling, SentencePiece-family tokenizers, special-token policy, and prefix-cache behavior remain unqualified.
- Backend order was fixed reference → Fastokens → Gigatoken. Fresh processes isolate model/length cells, but the benchmark does not yet perform A/B and B/A order reversal.
- Upstream currently requires nightly `portable_simd` and unstable Cargo profile rustflags. Stable Cargo cannot parse the workspace while the git dependency is present, even behind an optional feature.
- The git dependency is not `cargo package` compatible and adds about 135 normal packages. Its unconditional PyO3 0.29 link also forces this branch to move the workspace from PyO3 0.22 to 0.29.

Direct adoption is therefore not recommended yet. A practical upstream path would be a stable, tokenizer-only core with optional Python/Arrow/dataframe integrations and packageable releases.

## Validation

- `CARGO_UNSTABLE_PROFILE_RUSTFLAGS=true cargo +nightly test -p dynamo-tokenizers --features gigatoken`
- gated Qwen and Kimi exact-parity tests: 2/2 passed
- `CARGO_UNSTABLE_PROFILE_RUSTFLAGS=true cargo +nightly clippy -p dynamo-tokenizers --all-targets --features gigatoken -- -D warnings`
- `CARGO_UNSTABLE_PROFILE_RUSTFLAGS=true cargo +nightly test -p dynamo-parsers-v2-py`
- `cargo +nightly fmt --all -- --check`
- locked workspace metadata and diff whitespace checks

## HTML Report

[gigatoken-tokenizer-report.html](https://github.com/user-attachments/files/30326168/gigatoken-tokenizer-report.html)

## Disclaimer

This was run on desktop class hardware, not on high end server hardware -- so the perf numbers may not be representative of a real deployment on machines with faster CPUs, more CPU cores, etc.


rmccorm4:
Context for the `#[pyclass(..., skip_from_py_object)]` change on `PyToolCallChunk`:

- The pinned Gigatoken revision depends directly on PyO3 0.29. Because `pyo3-ffi` links the Python runtime, Cargo cannot keep the workspace's prior PyO3 0.22 alongside it, so this prototype moves the workspace to 0.29.
- PyO3 0.29 is phasing out its automatic `FromPyObject` implementation for types that are both `#[pyclass]` and `Clone`. It emits a deprecation unless the class explicitly selects `from_py_object` or `skip_from_py_object`.
- `PyToolCallChunk` is output-only: Rust constructs it and `parse`, `parse_text`, and `finish` return `Vec<PyToolCallChunk>` to Python. No Python-exposed function accepts a `PyToolCallChunk` as an owned Rust input.
- Therefore `skip_from_py_object` is the narrower option. It opts out of an unused Python-object-to-owned-Rust extraction path and satisfies the `-D warnings` validation.

This does not change returning chunks to Python, `get_all`, frozen behavior, class identity, or `__repr__`. If a future API needs to accept `PyToolCallChunk` by value from Python, it should switch to `from_py_object` or add a custom conversion.

tokenizers/Cargo.toml
```diff
@@ -14,12 +14,16 @@ repository.workspace = true
...
+[features]
+gigatoken = ["dep:gigatoken"]
+
 [dependencies]
 aho-corasick.workspace = true
 anyhow.workspace = true
 blake3.workspace = true
 moka = { workspace = true, features = ["sync"] }
 fastokens.workspace = true
+gigatoken = { package = "gigatoken", git = "https://github.com/marcelroed/gigatoken.git", rev = "ecf968da2b7300e33f90e8bd9c96a11a335a01ae", optional = true }
```
rmccorm4, line 26: We'll need a published crate of gigatoken to be able to publish the dynamo-tokenizers crate with this as a dependency

PR #140, opened 2026-07-24
chore(conformance): vLLM 0.26.0 + vLLM Rust 0.24.0/0.26.0 peer captures — NO PARSER CHANGE (DIS-2408)

keivenchang:
## Overview

**No parser was added or changed.** Nothing under `parsers/`, `protocols/`, `tokenizers/` or `renderer/` is touched — verified: `git diff --name-only origin/main...HEAD -- parsers/ protocols/ tokenizers/ renderer/` is empty. This PR only re-captures what existing parsers already produce, plus fixes to the harness that does the capturing. No crate version moves and no published behaviour changes.

Dynamo pins vLLM **0.26.0** (`pyproject.toml`), so the conformance peer stub moves to 0.26.0 and the tool-calling corpora gain captures at that version. vLLM Rust **0.24.0** is captured as well: it is the last release where gemma4 still implements `tool::ToolParser`, so it covers 40 gemma4 cases that 0.25+ cannot reach through the tool probe.

Rebased onto current `main` and squashed to one commit (the repo is squash-merge only).

## Why the diff is large for a capture-only change

| area | files | +/- |
|---|---|---|
| conformance harness code | 18 | +1067 / −385 |
| fixtures (binary shards + manifest) | 24 | +125 / −69 |
| conformance tests | 3 | +109 / −59 |
| conformance docs | 3 | +61 / −35 |
| CI | 1 | +6 / −1 |

**Two thirds of the harness lines are one file.** `capture_peer_versions.py` is +734/−0 and replaces `capture_streamv2_versions.py` (−195): one `--corpus {batch,stream,reasoning}` × `--impl {vllm_python,sglang_python,vllm_rust}` entry point instead of a per-corpus capturer. That is a rewrite of the capture driver, not new parser logic. Honest caveat: `recapture_peer.py` still exists on `main` and is still referenced by `capture_layers.py`, so this consolidates the stream capturer, not every per-version capturer.

The remaining ~330 lines are small, and mostly forced by the 0.25 crate rename or found by review:

- `capture_vllm_rust.py` (+99/−50) — vLLM renamed the Rust crate in 0.25.0 (`vllm-tool-parser` @ `rust/src/tool-parser` -> `vllm-parser` @ `rust/src/parser`) and reworked `ToolParserOutput` from public fields into accessors. The probe hardcoded the new layout, so **0.24.0 could not be captured at all**. The probe body is now one template with per-dialect substitutions rather than two near-duplicate Rust blobs, and layout resolution moved into `capture_driver.vllm_rust_crate_dir()`, replacing three copies of the same path check.
- `markers.py` / `generate_conformance_table.py` / `model.py` / `conformance.js` — one `exception` contract. `cmp_model` now emits `err` (parser RAN and RAISED, distinct from `unavailable` = no such parser), produced once via `_is_exception` + a new `cmp_entry`. All **three** cmp producers (`cmp_model`, the Unified tab, the reasoning tab) build entries through `cmp_entry`, so the payload key set cannot drift from the JS view again.
- `resolve_reasoning_fixtures.py` — stamped the *selected* peer version onto fixtures whose data came from a lower one, so a pinned 0.26.0 labelled 0.24.0's captured output. It now stamps the version actually resolved.
- `build_stream_fixtures.py` (+16/−50) — deleted a duplicate `VLLM_RUST_UNAVAILABLE` and both vLLM Rust helpers that already had owners in `capture_driver`.
- `parser_families.yaml` (+5/−2) — maps `minimax_m3` and `phi4` for `vllm_rust`. Those parsers landed in vLLM Rust 0.24.0, but the `null` mapping meant the families were never probed at *any* version, so they rendered "No vLLM Rust parser" regardless of the version selected. Recovers 13 + 37 cases. `jamba`/`pythonic` stay `null` — 0.26.0 genuinely has no Rust parser for them.

The fixture diff is small in line terms because the corpus ships as binary LFS shards; the real content is the new `vllm_rust-0.24.0` / `vllm_rust-0.26.0` / `vllm_python-0.26.0` version dirs plus the regenerated manifest.

## The fixture change is purely additive

Every capture in this PR was re-taken against `main`'s inputs and `main`'s anchors, so no released label is rewritten:

```
pre-existing VERSION-DIR files modified:   0
files added:                             435
    vllm_python-0.25.1  115    vllm_rust-0.24.0  61
    vllm_python-0.26.0  115    vllm_rust-0.25.1  72
                               vllm_rust-0.26.0  72
only in-place change:  fixtures-batch-on-stream-v2  139
```

`dynamo_v1-3.0.0`, `vllm_python-0.23.0`, `sglang_python-0.5.12.post1` and every other existing version dir are byte-identical to `main`. The single in-place corpus is `fixtures-batch-on-stream-v2`, which is single-snapshot by design (no version dirs) and is refreshed to vLLM 0.26.0 / SGLang 0.5.14.

Why it needed re-capturing: this branch and `main` had independently re-authored some fixture inputs (e.g. `glm47/TOOLCALLING.batch.2.c` was `'Checking: …'` here and `'Both: …'` on `main`). Fixtures ship as binary LFS shards, so a rebase can only take one side — and taking `main`'s `inputs/` while keeping this branch's `expected` blocks paired one lineage's input with the other's output across 145 files. Re-capturing from a single lineage fixes that by construction rather than by hand-merging. The Python stream dirs needed a second pass: restoring `main`'s `vllm_python-0.23.0` anchor (38 files different) invalidates any changed-only overlay computed against the old anchor.

Also newly enforced: version-dir publication is **atomic**. `_version_outdir()` is the single owner of every write (batch, container stream, reasoning, Rust stream previously each did `makedirs(exist_ok=True)` + truncating `open(w)` into the final path). A run now stages its whole tree beside the corpus and `_publish_staged()` renames it into place only after the run succeeds — `os.rename` onto a non-empty directory raises ENOTEMPTY, so the rename *is* the claim: two concurrent runs cannot both publish, and a run that dies midway leaves no `<impl>-<version>/` for the next run to inherit. An `exists()` check alone was not sufficient; a synchronized two-process repro had both processes pass it and write. Covered by `test_two_concurrent_runs_cannot_both_publish` and `test_version_dirs_are_append_only`.

## Where

`conformance/utils/src/` (capture plumbing + markers/model contract) + `conformance/fixtures/` shards + `conformance/fixtures-manifest.json`.

## Deferred, deliberately

The vLLM Rust **reasoning** probe and the reasoning peer captures are **not** in this change. Rebasing onto main dropped this branch's earlier multi-version reasoning tab (`e6cfbe74`) — main's rewritten `tables/reasoning/table.py` renders one version per impl — so any additional reasoning version dir either mislabels its peer version or fails `test_v2_reasoning_uses_current_peers`. Reasoning stays at main's single `vllm_python-0.24.0` until that feature is re-applied on the rewritten module. The Rust reasoning probe itself works (hermetic — synthetic tokenizer, no model, no GPU) and lands with it.

Also unchanged: `vllm_rust` batch-on-stream stays at its existing version — the only tool that refreshes it also rewrites Dynamo v2 expectations, which needs a crate version bump.


.github/workflows/ci.yml
```diff
@@ -102,7 +102,12 @@ jobs:
     needs: changes
     if: ${{ needs.changes.outputs.conformance == 'true' || github.event_name == 'workflow_dispatch' }}
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    # Render cost scales with the number of peer version dirs: every <impl>-<version>
+    # is another compare candidate on every cell. main sat at ~8m24s of a 10m budget,
+    # so adding the vLLM 0.26.0 / vLLM Rust 0.24.0+0.26.0 captures timed the job out.
+    # Raised to match the `rust` lane's headroom. If this creeps toward 20m, the fix is
+    # to make the render cheaper (or shard it), not to raise the ceiling again.
+    timeout-minutes: 20
```
rmccorm4, line 110: Not blocking for this PR, but maybe we can just hold onto version N and N-1, for example get rid of vllm 0.24.0, and keep just vllm 0.26.0/0.25.1, then when we eventually bump again get rid of 0.25.1 and add the new version - rather than letting this get slower and slower

rmccorm4 (approved):
One comment, otherwise LGTM: https://github.com/ai-dynamo/frontend-crates/pull/140/changes#r3740064246

indrajit96 (approved):
LGTM!

PR #141, opened 2026-07-24
perf(parsers): switch Pythonic parser to Ruff

saturley-hall:
## Summary

- replace the RustPython parser used by Pythonic tool-call parsing with `rustpython-ruff_python_parser` and its AST crate
- preserve the existing JSON conversion behavior for booleans, nulls, integers, floats, strings, lists, dictionaries, and unsupported arguments
- add regression coverage for scalar literals, very large integers, and skipped unsupported arguments
- remove an obsolete cargo-deny exception made unnecessary by the new dependency graph

## Performance

Criterion release benchmarks were run on Linux ARM64 with Rust 1.96.1, a 1-second warm-up, 3-second measurement window, and 60 samples per case.

| Case | Before | After | Median change | Speedup |
|---|---:|---:|---:|---:|
| AST parse, simple call | 5.0495 µs | 1.3501 µs | -73.25% | 3.74× |
| AST parse, nested / 3 calls | 42.705 µs | 7.0762 µs | -84.05% | 6.04× |
| End-to-end, simple call | 5.7070 µs | 1.8763 µs | -67.73% | 3.04× |
| End-to-end, nested / 3 calls | 33.891 µs | 8.6899 µs | -72.88% | 3.90× |

All four changes were statistically significant (`p < 0.05`). The benchmark release executable decreased from 8,780,464 bytes to 7,053,552 bytes, a 19.67% reduction.

## Validation

- `cargo clippy -p dynamo-parsers --all-targets --locked -- -D warnings`
- full `dynamo-parsers` unit and integration suite: 751 passed, 0 failed, 5 ignored
- tool-calling batch parity: 1 passed
- dependency policy checks passed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -194,6 +194,36 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
+name = "attribute-derive"
+version = "0.10.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "05832cdddc8f2650cc2cc187cc2e952b8c133a48eb055f35211f61ee81502d77"
+dependencies = [
+ "attribute-derive-macro",
+ "derive-where",
+ "manyhow",
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "attribute-derive-macro"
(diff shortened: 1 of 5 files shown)


nv-anants (approved):

keivenchang (approved):
thanks Harrison — clean swap. verified it's output-identical: the pythonic parser matches byte-for-byte across all conformance inputs, the rendered conformance table is unchanged, and the full dynamo-parsers suite is green. LGTM.

PR #143, opened 2026-07-26
feat(protocols)!: support media in tool results

michaelfeil:
## Summary

Allow OpenAI-compatible tool messages to carry image, video, and audio URL parts, and Anthropic tool results to carry typed image blocks.

Text-only results are unchanged. This owns the tool-message request types, so the crate release is semver-major. Consumed by [ai-dynamo/dynamo#12180](<https://github.com/ai-dynamo/dynamo/issues/12180>).

## Validation

* `cargo test -p dynamo-protocols`
* `cargo clippy -p dynamo-protocols --tests -- -D warnings`
* `cargo fmt --all --check`

Related to ai-dynamo/dynamo#12179.

### DIFF
--- a/protocols/src/types/anthropic.rs
+++ b/protocols/src/types/anthropic.rs
@@ -333,7 +333,7 @@ impl ToolResultContent {
                 .into_iter()
                 .filter_map(|b| match b {
                     ToolResultContentBlock::Text { text } => Some(text),
-                    ToolResultContentBlock::Other(_) => None,
+                    ToolResultContentBlock::Image { .. } | ToolResultContentBlock::Other(_) => None,
                 })
                 .collect::<Vec<_>>()
                 .join(""),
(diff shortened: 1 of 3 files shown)


michaelfeil:
@ishandhanani @KrishnanPrash, could you take a look at the protocol shape? The Dynamo consumer is ai-dynamo/dynamo#12180.

protocols/src/types/chat.rs
```diff
@@ -309,6 +305,51 @@ pub struct ImageUrl {
     pub uuid: Option<Uuid>,
 }
 
+/// Tool messages accept the same multimodal content parts as user messages.
+///
+/// OpenAI's schema currently limits tool content parts to text, but
+/// OpenAI-compatible multimodal backends also accept media observations
+/// returned by tools.
+pub type ChatCompletionRequestToolMessageContentPart = ChatCompletionRequestUserMessageContentPart;
```
indrajit96: Tool messages unintentionally accept audio and video.

Aliasing the complete user-message enum also allows `video_url`, `audio_url`, and `input_audio` in tool messages. 
This PR only implements image results, and Dynamo currently ignores at least input_audio during media extraction.
Could we define a dedicated tool-content enum containing only Text and ImageUrl? 
That keeps this change narrowly scoped and prevents the API from accepting modalities that are not fully supported.

protocols/src/types/chat.rs
```diff
@@ -309,6 +305,51 @@ pub struct ImageUrl {
...
+}
+
+/// Tool message using Dynamo's image-capable content type.
+#[derive(Debug, Serialize, Deserialize, Default, Clone, Builder, PartialEq)]
+#[builder(name = "ChatCompletionRequestToolMessageArgs")]
+#[builder(pattern = "mutable")]
+#[builder(setter(into, strip_option), default)]
+#[builder(derive(Debug))]
+#[builder(build_fn(error = "OpenAIError"))]
+pub struct ChatCompletionRequestToolMessage {
```
indrajit96, line 356: External Rust users of `dynamo-protocols`: potentially affected at compile time.
Before this PR

dynamo-protocols re-exported the type from async-openai. These two names referred to the exact same Rust type:

  `dynamo_protocols::types::ChatCompletionRequestToolMessage`
  `async_openai::types::chat::ChatCompletionRequestToolMessage`

A caller could use them interchangeably:

After this PR

Dynamo defines a new struct with the same name:
```
  pub struct ChatCompletionRequestToolMessage {
      pub content: ChatCompletionRequestToolMessageContent,
      pub tool_call_id: String,
  }
```

Could be compatibilty issue for exisitng cosumers of `dynamo-protocols` ? @rmccorm4 
AFAIK it's only dynamo which the corresponding PR takes care of.

protocols/src/types/anthropic.rs
```diff
@@ -348,7 +348,11 @@ pub enum ToolResultContentBlock {
     Text {
         text: String,
     },
-    /// Catch-all for non-text blocks (images, etc.) in tool results.
+    /// Image returned by a tool.
+    Image {
```
indrajit96, line 351: Because this enum is #[serde(untagged)], this variant does not check that the outer block has "type": "image". 
A valid Anthropic document block with a base64 source has the same fields and is therefore deserialized as an image. It can then be incorrectly sent through the image pipeline as application/pdf.
Serializing this variant also loses the outer "type": "image" field.

Could we dispatch using the block’s type field, as `AnthropicContentBlock` already does, and preserve unknown block types as Other? 
Please add tests proving that an image round-trips with its type and a document remains Other.

indrajit96 (commented):
Some comments on compatibilty and types

protocols/src/types/chat.rs
michaelfeil: true, we should also add support for audio and video, obviously. up to the model to not accept them. unlikley, since it produced them in the first place with a tool call.

michaelfeil: sorry, video is supported by kimik2.7 on baseten, today.

rmccorm4:
/ok to test [811870b](https://github.com/ai-dynamo/frontend-crates/pull/143/commits/811870b3fe9896eeb31e6561273b6896ba9f668e)

rmccorm4:
@ishandhanani to help test and drive to close

protocols/src/types/anthropic.rs
```diff
@@ -342,16 +342,71 @@ impl ToolResultContent {
...
+                let source = value
+                    .get("source")
+                    .cloned()
+                    .ok_or_else(|| serde::de::Error::missing_field("source"))
+                    .and_then(|value| {
+                        serde_json::from_value(value).map_err(serde::de::Error::custom)
+                    })?;
+                Ok(Self::Image { source })
+            }
+            _ => Ok(Self::Other(value)),
```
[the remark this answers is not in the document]
ishandhanani: Fixed in 811870b: a type-less block with a string `text` field is preserved as `Text`; the regression test covers that legacy shape.

protocols/src/types/anthropic.rs
ishandhanani: Fixed in 811870b: deserialization now dispatches on the outer `type`, so only `image` becomes `Image`; documents remain `Other`, and serialization retains `type: image`. The regression test covers image/document round-tripping.

protocols/src/types/chat.rs
ishandhanani: We're the only ones using this right now. I don't know if we need a full v5 major bump. I think once FWs pick this up as a full dep we can do that

ishandhanani (approved):

michaelfeil:
https://github.com/basetenlabs/dynamo/pull/502

PR #144, opened 2026-07-26
fix(parsers): tools are leaking in the reasoning content (deepseek)

slivanovich:
#### Overview:
Improves DeepSeek V4 reasoning parsing when a model starts emitting DSML tool-call markup before closing an open `<think>` block. The parser now treats the DeepSeek DSML tool section marker as a reasoning exit point, preventing tool-call markup from being counted or streamed as reasoning content.

#### Details:
**Rust - parsers/v1/src/reasoning/mod.rs**
- Added `DEEPSEEK_TOOL_SECTION_BEGIN` for the DeepSeek DSML tool-call marker
- Configured the DeepSeek V4 reasoning parser to force-exit reasoning when `<｜DSML｜` starts
- Added regression coverage for:
  - Batch parsing where DSML starts before `</think>`
  - Dangling partial DSML prefixes at the end of reasoning text
  - Streaming chunks where the DSML marker is split across deltas
  - Non-DSML lone `<` text that should remain reasoning content

**Rust - parsers/v1/src/reasoning/base_parser.rs**
- Updated truncated reasoning handling to withhold suffix bytes that may still become a configured tool-start marker
- Adjusted streaming overlap handling so even a one-character tool marker prefix is buffered until the next chunk disambiguates it
- Preserves existing behavior for parsers without tool-start markers

**Rust - parsers/v1/src/tool_calling/dsml/parser.rs**
- Changed DSML end-position detection to return `None` until the closing DSML block marker arrives
- Updated tests so incomplete DSML blocks stay buffered instead of reporting `chunk.len()` as a completed boundary

**Rust - parsers/v1/src/tool_calling/parsers.rs**
- Propagated the DSML parser's `Option<usize>` end-position result through the generic tool parser dispatch

- `parsers/v1/src/reasoning/mod.rs:23` - DeepSeek DSML marker constant.
- `parsers/v1/src/reasoning/mod.rs:238-242` - DeepSeek V4 reasoning parser registration with the DSML tool-start marker.
- `parsers/v1/src/reasoning/base_parser.rs:231-240` and `parsers/v1/src/reasoning/base_parser.rs:368-378` - Partial tool-start buffering in batch and streaming reasoning paths.
- `parsers/v1/src/tool_calling/dsml/parser.rs:51-63` - DSML end-position detection returning `None` for incomplete blocks.
- `parsers/v1/src/reasoning/mod.rs:378-438` - DeepSeek V4 regression tests covering DSML reasoning exits and split-marker streaming.

#### Related PRs:
- https://github.com/sgl-project/sglang/pull/22837 (the same problem with Qwen3)
- https://github.com/vllm-project/vllm/pull/33646 (the same problem with Kimi)


rmccorm4:
Hi @slivanovich, thanks for raising this!

Two quick questions:
1. Do you have some e2e curl requests that reproduce the leak? (@KrishnanPrash @indrajit96 we need to improve coverage here)
2. If you try setting `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2=1` in your deployment, do you still see the leak in your setup?

slivanovich:
> Hi @slivanovich, thanks for raising this!
> 
> Two quick questions:
> 
> 1. Do you have some e2e curl requests that reproduce the leak? (@KrishnanPrash @indrajit96 we need to improve coverage here)
> 2. If you try setting `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2=1` in your deployment, do you still see the leak in your setup?

Thanks for the quick response!
1. This problem seems to be probabilistic and appear while processing long context requests. So there is no exact curl that can show the bug. In out installation of DeepSeek-V4-Flash i've reproduced this on the very long context load test (N sessions up to 1M and more context).
2. No, i've not tried to use v2 parsers yet. However, as i can see (probably missing smth) there is only v2 tool parser for DeepSeek models (no reasoning parser). This fix is about "reasoning do not end while receiving tool start in it". Also the tool start token for DeepSeek models (dsml token) is not "atomic" in terms of generation process ("<", "|DSML|", ...), so part of this pr handles this particular case (cause empirically considered that `.with_tool_start_token(DEEPSEEK_TOOL_SECTION_BEGIN)` is not enough).

rmccorm4:
/ok to test [6c720ca](https://github.com/ai-dynamo/frontend-crates/pull/144/commits/6c720caa913b729ff9f075d487b04801c185b090)

parsers/v1/src/reasoning/base_parser.rs
```diff
@@ -364,7 +373,7 @@ impl ReasoningParser for BasicReasoningParser {
                         let ol_end = overlap(&current_text, &self.think_end_token);
                         let ol_tool = max_marker_overlap(&current_text, &self.tool_start_tokens);
                         let ol = ol_end.max(ol_tool);
-                        if ol >= 2 {
+                        if ol_end >= 2 || ol_tool > 0 {
```
rmccorm4: CC @indrajit96

slivanovich:
Have resolved @dynamo-review-agent[bot] comments

parsers/v1/src/reasoning/base_parser.rs
```diff
@@ -260,7 +260,7 @@ impl ReasoningParser for BasicReasoningParser {
                         exited_on_tool_start = true;
                     }
                     (None, None) => {
-                        // No end token — rest is reasoning (truncated)
+                        // No end token — rest is reasoning (truncated).
```
rmccorm4: unnecessary change
```suggestion
                        // No end token — rest is reasoning (truncated)
```

parsers/v1/src/reasoning/mod.rs
```diff
@@ -386,6 +392,121 @@ mod tests {
...
+            normal.push_str(&result.normal_text);
+        }
+
+        assert_eq!(reasoning, "literal<not dsml");
+        assert_eq!(normal, "");
+    }
+
     #[test] // Inkling
     fn test_inkling_detect_and_parse() {
         let mut parser = ReasoningParserType::get_reasoning_parser_from_name("inkling");
```
indrajit96, line 569: Can we add a new test that does something like

```
  let mut parser =
      ReasoningParserType::get_reasoning_parser_from_name("deepseek_v4");

  parser.set_in_reasoning(true);

  let result = parser.detect_and_parse_reasoning(
      "I need weather<｜DSML｜tool_calls>...",
      &[],
  );

  assert_eq!(result.reasoning_text, "I need weather");
  assert_eq!(
      result.normal_text,
      "<｜DSML｜tool_calls>..."
  );
```

  This checks for case where

  - There is no visible <think> opener.
  - There is no </think> closer.
  - The DSML tool marker still ends reasoning correctly.

IIUC this is a valid case where prompt has  Prompt:     ...<think> resulting in 
Generation: I need weather<｜DSML｜tool_calls>...
The <think> opener belongs to the prompt, so it is not present in the generated text.

parsers/v1/src/reasoning/mod.rs
```diff
@@ -243,7 +245,11 @@ impl ReasoningParserType {
             // distinct variant so V4-specific divergence has somewhere to land.
             // See `ReasoningParserType::DeepSeekV4` docstring for rationale.
             ReasoningParserType::DeepSeekV4 => ReasoningParserWrapper {
-                parser: Box::new(basic_parser),
+                parser: Box::new(
+                    BasicReasoningParser::new("<think>".into(), "</think>".into(), false, true)
+                        .with_tool_start_token(DEEPSEEK_TOOL_BLOCK_BEGIN)
+                        .with_tool_start_token(DEEPSEEK_TOOL_INVOKE_BEGIN),
+                ),
```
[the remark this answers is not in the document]
indrajit96: No. In the real Dynamo flow, Dynamo sees that the prompt already ended with <think> and tells the parser:

`parser.set_in_reasoning(true);`

The parser therefore knows that "I need weather" is reasoning. When it sees the DSML marker, it correctly switches to tool-call parsing.

indrajit96 (commented):
CHanges look good to me, just one small request for the additional prompt-injected batch scenario test described above.
Thanks for finding and fixing this!
We also recently merged foundational Kimi K3 changes that touch the shared parser logic. 
Could you merge the latest main into this branch and run the parser tests again to verify everything works together?

parsers/v1/src/reasoning/base_parser.rs
indrajit96: I checked the PR’s final diff no longer modifies `base_parser.rs.`

slivanovich:
> CHanges look good to me, just one small request for the additional prompt-injected batch scenario test described above. Thanks for finding and fixing this! We also recently merged foundational Kimi K3 changes that touch the shared parser logic. Could you merge the latest main into this branch and run the parser tests again to verify everything works together?

Hi, thanks for the response!
Done: test for the tool call without <think> opener, also add some into jail.rs part (V3.2 uses quite the same DSML notation, but anyway)

yeah, base parser fix has merged faster)

indrajit96:
/ok to test [d3f8d64](https://github.com/ai-dynamo/frontend-crates/pull/144/commits/d3f8d6456c58c4f9a251dd58643891f50daf7de8)

ishandhanani (approved):

PR #148, opened 2026-07-27
fix(renderer): honor tool_choice=none and full reasoning_effort range in DeepSeek native formatters

surajssd:
- inject_tools_and_response_format now strips tool definitions from the prompt when tool_choice=="none" (OpenAI semantics: the model must not call tools). Prevents raw <|DSML|> tool markup leaking into responses. response_format injection is unaffected. Covers both v4 and v3.2 native formatters; mirrors the jinja path's exclude_tools_when_tool_choice_none.
- resolve_reasoning_effort accepts the OpenAI-style range: none/minimal/ low/medium map to no preamble without warning, xhigh maps to Max. Previously anything but max/high logged a warning on every request.

### DIFF
--- a/renderer/src/deepseek/v4.rs
+++ b/renderer/src/deepseek/v4.rs
@@ -351,10 +351,13 @@ impl DeepSeekV4Formatter {
         let v = args.get("reasoning_effort")?;
         match v.as_str() {
             Some("max") | Some("xhigh") => Some(ReasoningEffort::Max),
-            Some("high") => Some(ReasoningEffort::High),
-            // Explicit "no extra effort" levels from the OpenAI-style range —
-            // valid input, no prompt preamble, and no warning.
-            Some("none") | Some("minimal") | Some("low") | Some("medium") => None,
+            // DeepSeek V4 only natively distinguishes none/high/max, so the
+            // OpenAI-style intermediate levels map to "high" — the model still
+            // reasons, it just doesn't get the max-effort preamble.
+            Some("high") | Some("minimal") | Some("low") | Some("medium") => {
+                Some(ReasoningEffort::High)
+            }
+            Some("none") => None,
             _ => {
                 tracing::warn!(
                     value = ?v,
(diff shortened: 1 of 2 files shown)


rmccorm4:
/ok to test [57f969a](https://github.com/ai-dynamo/frontend-crates/pull/148/commits/57f969aaf62aed293f7da9575d7f6bc1922f5b25)

renderer/src/deepseek/v4.rs
```diff
@@ -350,12 +350,15 @@ impl DeepSeekV4Formatter {
         let args = args?;
         let v = args.get("reasoning_effort")?;
         match v.as_str() {
-            Some("max") => Some(ReasoningEffort::Max),
+            Some("max") | Some("xhigh") => Some(ReasoningEffort::Max),
             Some("high") => Some(ReasoningEffort::High),
+            // Explicit "no extra effort" levels from the OpenAI-style range —
+            // valid input, no prompt preamble, and no warning.
+            Some("none") | Some("minimal") | Some("low") | Some("medium") => None,
```
rmccorm4: Since Deepseek only recognizes max/high/none, I think anything `none < level <= high` should map to high, because the expectation for levels like `medium`/`low` is that some thinking will happen, but currently this will disable thinking.

I see some similar mapping in vllm testing here: https://github.com/vllm-project/vllm/blob/272abd5f486967f1fb9db7ca7504f8c34235ef50/tests/tokenizers_/test_deepseek_v4.py#L214-L217

Proposing we map none->none, but **map minimal/low/medium/high -> high**,  and xhigh/max -> max.

What do you think?

surajssd: Yep makes sense. Added changes in the latest commit PTAL @rmccorm4

rmccorm4 (approved):

rmccorm4:
/ok to test [978d1f9](https://github.com/ai-dynamo/frontend-crates/pull/148/commits/978d1f91e7d0658181955493c3e147fdf7da3718)

PR #151, opened 2026-07-28
feat(parsers): UnifiedParser for Qwen3 — one ordered reasoning/text/tool stream

keivenchang:
#### Overview:

This PR replaces the two-parser chain Dynamo uses for **Qwen3** with one streaming state machine that owns reasoning, visible content and tool calls together, so the three come out in the order the model actually produced them. Today reasoning is parsed first over the whole output and merged into a single field, which silently moves any thought that happened between or after a tool call to the front. Qwen3 only — gemma4 and kimi_k2 keep the current split until their unified parsers land.

Input: `<think>Look it up.</think><tool_call>…get_weather(city=Paris)…</tool_call><think>Now answer.</think>It's 18C.`

```
before:  reasoning("Look it up.Now answer.")  |  tool_call(get_weather, {"city":"Paris"})  |  text("It's 18C.")
after:   reasoning("Look it up.")  |  tool_call(get_weather, {"city":"Paris"})  |  reasoning("Now answer.")  |  text("It's 18C.")
```

The second thought moved ahead of the call it followed and fused with the first. A client rendering thoughts inline shows them in the wrong place; one counting reasoning turns sees one where the model produced two. Ordering is not a field the split can add — it is gone at the seam between the two parsers.

Two independent bugs this surfaced, both on the **shipping** tool-only `qwen3_coder` parser, not just the unified path:

```
arg truncation   <parameter=cmd>git log </tool_call> --oneline</parameter>
   before:  {"cmd": "git log </tool_call>"}            <- value cut at a marker that was data
   after:   {"cmd": "git log </tool_call> --oneline"}

markup leak      <think>a</tool_call>b</think>
   before:  reasoning("a</tool_call>b")                <- stray markup became content
   after:   reasoning("ab")
```

image

#### Details:

- The existing `WrappedBlockScanner` (which already owns block open/close, bare-`<function=>` recovery, orphan-close stripping, chunk-boundary holdback and EOF drop) now emits ordered deltas, and reasoning is an opt-in channel on it. Tool-only parsers project the same deltas down to `ToolParseResult` and are unchanged. One scan, two views — no second scanner to drift.
- Nesting is deliberately **asymmetric**, because tool structure dominates: a call emitted inside a thought is extracted and the thought splits around it (burying it would drop the call *and* leak its markup, `I3`), while a reasoning marker inside a tool argument stays data (`I7`). Both directions are pinned by the golden corpus (`tool_in_reason`, `reason_markup_in_arg`) and by unit tests.
- The conformance tab is rendered by Python from a **committed capture shard** — it never runs the Rust parsers, so a parser change that is not re-captured leaves the page showing old behavior while every test passes. The capture now uses the unified parser per-family, and `committed_dynamo_capture_matches_the_live_parsers` fails when the shard drifts, naming the stale cases and printing the regenerate commands.
- Popup/highlighter fixes ride along: columns size to content and are equalized, and the colorizer no longer paints markers that live inside an argument value — declared per family in `parser_families.yaml` (`opaque:`), mirrored in both the Python source of truth and its JS port, with byte-parity tests.

- The shared scan core was cleaned up alongside: the `I8` run-merge existed in three copies and is now one helper used by both the scanner and `assemble`; the two "which marker comes first?" sites share one `earliest()`; per-family `UnifiedParser` impls collapse into a generic `ScannerUnified<E>` so a family is a factory rather than a struct plus impl; and `preserve_special_tokens` — declared, stored and plumbed, never called — is deleted. Behavior-preserving, and production code drops 348 -> 333.

#### Verification:

`cargo fmt`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, the full workspace suite, and `conformance/utils/check.sh ci` are all clean. All **33** qwen3 `UNIFIED.*` cases match the golden oracle, with stream/batch parity and per-stream isolation. Beyond the corpus: an exhaustive **18,623-splitting** chunk sweep over every case found 0 chunk-variance, and an adversarial sweep over **2,196** generated shapes (nested/duplicated openers, strays, empty segments) asserting no marker reaches a payload found 292 problems, all fixed, now 0. The 8 pre-existing tool-only `qwen3_coder` tests and the tool-calling conformance parity suites still pass, so the shared-scanner change is behavior-preserving for the tool-only contract.

Unified tab, Dynamo column vs golden: **qwen3 0 red (was 5)**; gemma4 5 and kimi_k2 13 unchanged (still the split).

#### Known gap (follow-up):

`tool_choice=required` and named functions use guided decoding that emits bare JSON rather than Qwen XML, and Dynamo can prefill the reasoning/content channel so the model never emits its own `<think>`. This parser handles neither, and the corpus has no cases for request modes at all. Tracked separately in [DIS-2544](https://linear.app/nvidia/issue/DIS-2544); PoC in #163.

`parsers/v2/src/unified/mod.rs`, then `parsers/v2/src/unified/qwen3.rs`, then `parsers/v2/src/tool_calling/scan.rs`, then `conformance/tests/unified_parity.rs`

/coderabbit profile chill

### DIFF
--- a/parsers/v2/src/tool_calling/scan.rs
+++ b/parsers/v2/src/tool_calling/scan.rs
@@ -28,10 +28,30 @@
 //! [`WrappedBlockSpec`] fields, not silent copy drift — e.g. MiniMax M2
 //! clears the normal-text suppression latch after a bare-invoke recovery
 //! while M3/qwen3/kimi keep it latched ([`BareRecoveryLatch`]).
+//!
+//! # Ordered output
+//!
+//! The drain loop emits [`UnifiedDelta`]s — text, reasoning and calls in the
+//! order the model produced them. Tool-only parsers project that down to
+//! [`ToolParseResult`] via `push`/`finish` and see no behavior change (that
+//! projection is lossy about order, which is all the tool-only contract ever
+//! promised). Unified parsers take the ordered list through `push_ordered` /
+//! `finish_ordered`. One scan, two views — so reasoning-aware and
+//! reasoning-unaware families cannot drift on marker handling, chunk-boundary
+//! holdback, or recovery.
+//!
+//! Reasoning is opt-in per scanner via [`WrappedBlockScanner::with_reasoning`].
+//! The two nestings are deliberately ASYMMETRIC, because tool structure
+//! dominates: a tool call emitted INSIDE a thought is a real call, so it is
+//! extracted and the thought splits around it; but a reasoning marker inside a
+//! tool argument is argument data (`I7`), because the in-block scan never looks
+//! for one. Burying a nested call would drop it AND leak its markup into the
+//! reasoning payload (`I3`).
 
 use std::collections::HashSet;
 
 use crate::tool_calling::traits::{ToolCallDelta, ToolParseResult};
+use crate::unified::UnifiedDelta;
 
 /// Longest non-empty proper prefix of any `marker` that `text` ends with, so a
 /// marker split across chunk boundaries is held back instead of leaked as
@@ -143,6 +163,20 @@ pub(crate) struct WrappedBlockSpec {
     pub drop_invoke_crossing_block_end: bool,
 }
 
+/// The reasoning channel a unified scanner also owns.
+///
+/// Only meaningful outside tool blocks — see the module docs.
+pub(crate) struct ReasoningSpec {
+    /// Opener, e.g. `<think>`.
+    pub start: String,
+    /// Closer, e.g. `</think>`.
+    pub end: String,
+    /// Stream begins INSIDE reasoning with no opener, because the chat template
+    /// pre-filled it (policy P5). Qwen3 is not one of these; DeepSeek-R1-style
+    /// forced-reasoning templates are.
+    pub forced_start: bool,
+}
+
 /// Per-family hook: parse one complete invoke (opener..closer inclusive) into
 /// a call delta. `None` means the invoke was malformed and is dropped.
 pub(crate) trait InvokeEmitter {
@@ -167,6 +201,35 @@ enum Marker {
     Block(usize),
     /// A bare invoke opener with no block wrapper (recovery path).
     BareInvoke,
+    /// A reasoning opener; carries the matched token length.
+    ReasoningStart(usize),
+}
+
+/// Append `text` as visible content, merging with a trailing text delta so one
+/// drain does not emit a run of adjacent same-kind fragments.
+fn push_text(out: &mut Vec<UnifiedDelta>, text: &str) {
+    if text.is_empty() {
+        return;
+    }
+    match out.last_mut() {
+        Some(UnifiedDelta::Text { text: prev }) => prev.push_str(text),
+        _ => out.push(UnifiedDelta::Text {
+            text: text.to_string(),
+        }),
+    }
+}
+
+/// Append `text` as reasoning, merging with a trailing reasoning delta.
+fn push_reasoning(out: &mut Vec<UnifiedDelta>, text: &str) {
+    if text.is_empty() {
+        return;
+    }
+    match out.last_mut() {
+        Some(UnifiedDelta::Reasoning { text: prev }) => prev.push_str(text),
+        _ => out.push(UnifiedDelta::Reasoning {
+            text: text.to_string(),
+        }),
+    }
 }
 
 /// The shared buffer-and-scan drain loop for wrapped-invoke grammars.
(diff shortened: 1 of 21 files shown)


rmccorm4:
Can share more offline, but for `tool_choice=required` and named functions are missing some support -- rough idea of some fixes here: https://github.com/ai-dynamo/frontend-crates/pull/163, but please take it over for completeness, that PR is just a PoC example, not planning to drive it forward myself at the moment

keivenchang:
taking it over, thanks Ryan. two things i found looking at #163: it's based on the pre-rebase #151, so its 4 commits still carry my old 3 and files that no longer exist (`conformance/unified/golden/*.yaml`, `unified_results.json`) — so i'll cherry-pick `4ba15bb` onto the current head rather than merge it, and expect real conflicts in `scan.rs`/`unified/mod.rs` since both moved since.

the bit i think matters most is that the corpus has no request-mode coverage at all — every case is native XML with a self-opened `<think>` — which is exactly why this stayed green while named/required were 0/14 for you. so i want to add UNIFIED cases for guided-JSON output and a prefilled reasoning channel FIRST, get them failing, then land the parser side. otherwise we fix it once and nothing stops it regressing.

keeping it on DIS-2544 as a follow-up and calling it out as a known gap in the #151 description, so this one isn't blocked on it.

rmccorm4:
Suggested implementation for the two `scan.rs` findings:

1. Store the reasoning markers as static references. All current markers are compile-time constants, so `ReasoningSpec` can be a small `Copy` value:

```rust
#[derive(Clone, Copy)]
pub(crate) struct ReasoningSpec {
    pub start: &'static str,
    pub end: &'static str,
    pub forced_start: bool,
}
```

Keep the required ownership at scanner construction time:

```rust
self.spec.holdback_markers.push(reasoning.start.to_string());
self.spec.holdback_markers.push(reasoning.end.to_string());
self.spec.orphan_markers.push(reasoning.end.to_string());
```

Then `drain_reasoning()` copies two string pointers instead of cloning two `String`s on every push.

2. Make `drain_reasoning()` fallible and propagate an invalid state through the existing parser error path:

```rust
fn drain_reasoning(
    &mut self,
    out: &mut Vec<UnifiedDelta>,
    flush: bool,
) -> anyhow::Result<bool> {
    let Some(reasoning) = self.reasoning else {
        anyhow::bail!("reasoning state active without a reasoning spec");
    };

    let (start, end) = (reasoning.start, reasoning.end);
    // ...
}
```

The caller becomes:

```rust
if self.drain_reasoning(&mut out, flush)? {
    continue;
}
```

This removes both per-push marker allocations and replaces the production `expect()` with a normal parser error. I prototyped this exact shape locally against `b6e28b4`: `cargo fmt --all -- --check` passed, and `cargo test -p dynamo-parsers-v2 unified --quiet` passed all 21 focused tests.

A broader follow-up could combine `reasoning`, `in_reasoning`, and `resume_reasoning` into one enum so the invalid state is unrepresentable, but I would keep this PR to the narrow change above.

parsers/v2/src/tool_calling/scan.rs
```diff
@@ -192,26 +277,173 @@ impl<E: InvokeEmitter> WrappedBlockScanner<E> {
...
+    /// Returns `true` once the reasoning span yielded (closer seen, tool opener
+    /// reached, or promoted at EOF) and the caller should keep draining; `false`
+    /// means more input is needed.
+    fn drain_reasoning(&mut self, out: &mut Vec<UnifiedDelta>, flush: bool) -> bool {
+        let (start, end) = {
+            let r = self
+                .reasoning
+                .as_ref()
+                .expect("in_reasoning implies a reasoning spec");
+            (r.start.clone(), r.end.clone())
```
rmccorm4: comment on these two clone() calls and the .expect() call: https://github.com/ai-dynamo/frontend-crates/pull/151#issuecomment-5126641442

rmccorm4 (approved):
LGTM other than https://github.com/ai-dynamo/frontend-crates/pull/151/changes#r3680087496

parsers/v2/src/tool_calling/scan.rs
keivenchang: both applied in 7246d28, thanks — good catch on both.

**1) `ReasoningSpec` as static refs — yes, done exactly as you wrote it.** It's `Copy` over `&'static str` now (`scan.rs:172-178`), with the allocation moved to construction (`scan.rs:302-304`). No `.clone()` left in the file.

**2) Fallible `drain_reasoning` — done.** It returns a `Result` and errors instead of panicking (`scan.rs:340-346`), caller propagates. No `expect()` or `unwrap()` left in the file. You were right that it wasn't a local invariant.

Re-verified against the parser tests, the golden gate, and an exhaustive chunk-splitting sweep — no behaviour change. And agreed, keeping the state-machine enum out of this PR.

keivenchang:
both applied in 7246d28, thanks — good catch on both.

**1) `ReasoningSpec` as static refs — yes, done exactly as you wrote it.** It's `Copy` over `&'static str` now (`scan.rs:172-178`), with the allocation moved to construction (`scan.rs:302-304`). No `.clone()` left in the file.

**2) Fallible `drain_reasoning` — done.** It returns a `Result` and errors instead of panicking (`scan.rs:340-346`), caller propagates. No `expect()` or `unwrap()` left in the file. You were right that it wasn't a local invariant.

Re-verified against the parser tests, the golden gate, and an exhaustive chunk-splitting sweep — no behaviour change. And agreed, keeping the state-machine enum out of this PR.

PR #154, opened 2026-07-28
feat(tokenizers): add basetenkenizer backend

michaelfeil:
## Summary

- add `basetenkenizer` 0.2.8 as a native `dynamo-tokenizers` backend for encode, batch encode, decode, and incremental decode
- add object-safe segmented encoding to preserve trusted renderer control tokens separately from untrusted user/tool content
- keep segmented encoding tiktoken-safe for Kimi K3 long-input token-ID parity
- honor existing tokenizer construction options and prefix-cache safety requirements
- disable the unused `hf-hub` feature for the local-file adapter

## Motivation

The existing `FastTokenizer` integration provides a faster encoder but falls back to Hugging Face for decoding and only accepts flattened strings. Newer model renderers, notably Kimi K2.7 Code and Kimi K3 XTML, produce segments carrying an `allow_special` trust boundary. Concatenating those segments before encoding can turn control-token-looking user content into structural tokens.

This PR makes segmented encoding part of the object-safe `Encoder` interface. Unsupported backends return an explicit error instead of flattening segments unsafely. `BasetenTokenizer` implements the capability using `basetenkenizer::Tokenizer::encode_segments_tiktoken_safe`; the frontend does not expose a non-tiktoken-safe segmented mode.

This intentionally does not change the default `.json` tokenizer factory, preserving current backend selection and model coverage. A native Kimi K3 XTML renderer can be added independently in `dynamo-renderer` and pass its segments through this API.

## Validation

- `cargo test -p dynamo-tokenizers --all-targets --locked`
- `cargo clippy -p dynamo-tokenizers --all-targets --all-features --locked -- -D warnings`
- `cargo check --workspace --all-targets --locked`
- `cargo package -p dynamo-tokenizers --allow-dirty --no-verify`
- `git diff --check`
- local crates.io `basetenkenizer` smoke probe against the vendored Kimi K3 tokenizer: load, raw encode/decode, and segmented XTML encode

`cargo deny` was not available in the local environment.

### DIFF
--- a/tokenizers/src/basetenkenizer.rs
+++ b/tokenizers/src/basetenkenizer.rs
@@ -6,7 +6,7 @@
 use std::path::Path;
 
 use super::{
-    EncodeSegment, Encoding, Error, Result, SegmentEncodingOptions, TokenIdType, TokenizerOptions,
+    EncodeSegment, Encoding, Error, Result, TokenIdType, TokenizerOptions,
     traits::{DecodeResult, Decoder, Encoder, Tokenizer},
 };
 
@@ -44,22 +44,14 @@ impl Encoder for BasetenTokenizer {
             .map_err(|e| Error::msg(format!("Baseten tokenizer batch encode error: {e}")))
     }
 
-    fn encode_segments(
-        &self,
-        segments: &[EncodeSegment<'_>],
-        options: SegmentEncodingOptions,
-    ) -> Result<Encoding> {
+    fn encode_segments(&self, segments: &[EncodeSegment<'_>]) -> Result<Encoding> {
         let segments = segments
             .iter()
             .map(|segment| (segment.text, segment.allow_special));
-        let ids = if options.tiktoken_safe {
-            self.tokenizer
-                .encode_segments_tiktoken_safe(segments, self.options.add_special_tokens)
-        } else {
-            self.tokenizer
-                .encode_segments(segments, self.options.add_special_tokens)
-        }
-        .map_err(|e| Error::msg(format!("Baseten tokenizer segment encode error: {e}")))?;
+        let ids = self
+            .tokenizer
+            .encode_segments_tiktoken_safe(segments, self.options.add_special_tokens)
+            .map_err(|e| Error::msg(format!("Baseten tokenizer segment encode error: {e}")))?;
         Ok(Encoding::Sp(ids))
     }
 }
@@ -204,9 +196,7 @@ mod tests {
             EncodeSegment::new(" world!", true),
         ];
 
-        let segmented = tokenizer
-            .encode_segments(&segments, SegmentEncodingOptions::default())
-            .unwrap();
+        let segmented = tokenizer.encode_segments(&segments).unwrap();
         let flattened = tokenizer.encode("Hello<ctl> world!").unwrap();
 
         assert_ne!(
@@ -254,7 +244,7 @@ mod tests {
 
         let plain = BasetenTokenizer::from_file(path.to_str().unwrap()).unwrap();
         let plain_ids = plain
-            .encode_segments(&segments, SegmentEncodingOptions::default())
+            .encode_segments(&segments)
             .unwrap()
             .token_ids()
             .to_vec();
@@ -265,10 +255,7 @@ mod tests {
                 add_special_tokens: true,
             });
         assert_eq!(
-            with_bos
-                .encode_segments(&segments, SegmentEncodingOptions::default())
-                .unwrap()
-                .token_ids(),
+            with_bos.encode_segments(&segments).unwrap().token_ids(),
             [&[23], plain_ids.as_slice()].concat()
         );
     }
--- a/tokenizers/src/basetenkenizer.rs
+++ b/tokenizers/src/basetenkenizer.rs
@@ -1,0 +1,219 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Baseten Tokenizer backend for high-performance BPE encoding and decoding.
+
+use std::path::Path;
+
+use super::{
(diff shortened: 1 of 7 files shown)


michaelfeil:
@codex review

michaelfeil:
@ishandhanani source code currently only available on crates.io, not on git.

tokenizers/src/lib.rs
```diff
@@ -65,6 +87,29 @@ pub mod traits {
...
+        /// The Baseten backend preserves legacy tiktoken behavior by splitting
+        /// each segment into chunks of at most 400,000 characters and splitting
+        /// whitespace/non-whitespace runs at 25,000 characters. Independent
+        /// chunks are encoded through the Rayon thread pool, then concatenated
+        /// in input order. Tokenizer post-processing is applied once after all
+        /// segment IDs have been joined.
+        ///
+        /// Backends must not implement this by flattening the segments first,
+        /// because that discards the special-token trust boundary.
+        fn encode_segments(&self, _segments: &[EncodeSegment<'_>]) -> Result<Encoding> {
```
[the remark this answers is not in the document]
michaelfeil: Fixed in `9eb1307`. `CachedTokenizer::encode_segments` now delegates directly to the inner tokenizer without consulting or populating L1, because flattening segments would discard their `allow_special` boundaries. Added regression coverage for both empty-specials (L1 disabled) and configured-specials (L1 enabled), including assertions that cache entries/hit/miss counters remain untouched. Full tokenizer tests, clippy, and workspace check pass.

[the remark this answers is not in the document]
ishandhanani: Lol - my codex found this too

tokenizers/src/basetenkenizer.rs
```diff
@@ -0,0 +1,266 @@
...
+/// Tokenizer backed by the `basetenkenizer` crate.
+pub struct BasetenTokenizer {
+    tokenizer: basetenkenizer::Tokenizer,
+    options: TokenizerOptions,
+}
+
+impl BasetenTokenizer {
+    /// Load a tokenizer from a Hugging Face `tokenizer.json` file.
+    pub fn from_file(path: &str) -> Result<Self> {
+        let tokenizer = basetenkenizer::Tokenizer::from_file(Path::new(path))
```
ishandhanani: [P1] Merge sibling `tokenizer_config.json` special tokens

`BasetenTokenizer::from_file` only passes `tokenizer.json` to basetenkenizer. The existing HF loader also promotes `special: true` `added_tokens_decoder` entries from sibling `tokenizer_config.json`, because some releases store control tokens only there (`hf.rs` documents Qwen2-VL's `<|image_pad|>` case). With those files, Baseten BPE-shatters the marker and `decode(..., true)` cannot remove it, so the advertised HF parity does not hold. Please apply the same merge before constructing Baseten and add a config-only token encode/decode regression test.

tokenizers/src/cache/mod.rs
```diff
@@ -248,6 +251,13 @@ impl Encoder for CachedTokenizer {
         // belongs here, not inside `encode`.
         inputs.iter().map(|&i| self.encode(i)).collect()
     }
+
+    fn encode_segments(&self, segments: &[EncodeSegment<'_>]) -> Result<Encoding> {
+        // L1 indexes flattened string offsets and cannot preserve each
+        // segment's allow_special boundary. Keep the operation correct by
+        // delegating without populating or consulting the cache.
+        self.inner.encode_segments(segments)
```
ishandhanani: [P2] Preserve token-observer accounting for segmented requests

Bypassing L1 here is correct, but this return also bypasses `token_observer`, whose contract says it runs after each successful encode while L1 is active. Dynamo uses it for cached/uncached tokenizer-token totals, so every segmented request would disappear from those metrics. After successful delegation, when `l1_enabled`, report `observe_token_usage(0, encoding.token_ids().len())`; keep cache hit/miss/entry stats untouched and extend this regression test to assert the observer receives the uncached total.

ishandhanani (commented):
Two remaining integration issues found against head `9eb1307`; both have narrow fixes and targeted regression tests.

tokenizers/src/lib.rs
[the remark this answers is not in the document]
michaelfeil: Confirmed fixed in `9eb1307`: segmented encoding passes through to the inner tokenizer without touching L1, with regression coverage for both enabled and disabled cache configurations. Resolving this thread. — mf via codex

tokenizers/src/basetenkenizer.rs
michaelfeil: Fixed in `603422e`. `BasetenTokenizer::from_file` now reads sibling `tokenizer_config.json` and idempotently promotes or adds `special: true` `added_tokens_decoder` entries before construction. Added the requested config-only token regression covering atomic encode and decode with `skip_special_tokens` false/true. — mf via codex

tokenizers/src/cache/mod.rs
michaelfeil: Fixed in `603422e`. Successful segmented pass-through now reports `observe_token_usage(0, encoding.token_ids().len())` whenever L1 is active, while cache entries/hit/miss counters remain untouched. The regression test covers observer behavior with L1 both disabled and enabled. — mf via codex

michaelfeil:
@codex review

jthomson04 (approved):
Approving the crate-level change at `603422e`.

The earlier integration issues are addressed: `CachedTokenizer` forwards segmented encoding without consulting or populating L1, segmented token usage is correctly reported as fully uncached, and sibling `tokenizer_config.json` special tokens are merged before constructing the Baseten tokenizer. The Rust CI is green.

I also verified ordinary Baseten encoding behind `CachedTokenizer` against the referenced Kimi K3 artifact: cached and uncached token IDs matched and L1 recorded real hits.

The segmented path remains intentionally uncached. I think that is acceptable for this upstream primitive, but the downstream Dynamo integration needs to preserve renderer trust metadata and explicitly account for the full-prompt retokenization cost.

PR #164, opened 2026-07-29
fix(renderer): adaptively normalize non-leading system messages

KrishnanPrash:
#### Overview

Agent clients (Claude Code on `/v1/messages`, Codex on `/v1/responses`) send `role:"system"` turns mid-conversation. Strict chat templates reject a system message that is not first and raise, so the frontend returns HTTP 500 on every request. See ai-dynamo/dynamo#11762.

Rather than unconditionally rewriting messages for every model, this detects at model load time whether the template actually rejects these streams, and normalizes only when it does. Permissive templates, which are the large majority, render byte for byte as they do today.

Before: `System message must be at the beginning.` -> HTTP 500.
After: renders normally. Permissive models are unchanged.

#### Details

`formatters.rs`: `detect_requires_system_normalization` runs inside the formatter constructor, alongside the existing `detect_content_array_usage` and `detect_passthrough_template` probes. It renders three message shapes through the compiled template (non-leading system, consecutive users, and a system after an assistant turn) with the model's own bos/eos/unk supplied, and flags the template if any raise while a plain `[system, user]` baseline still renders. The result is cached as `requires_system_normalization` on the formatter.

`oai.rs`: in `render()`, gated on that flag, `normalize_system_messages` merges a leading run of system messages into one, demotes any non-leading system to `user` in place, then coalesces the consecutive user turns that can produce.

Two details worth calling out for review:

- The demotion is in place rather than folded to the front. Folding changes the leading block whenever a mid-conversation reminder toggles, which invalidates the whole prefix; keeping it in place only shifts a block deeper in. Measured over 100 x 20-turn conversations with a toggling reminder, folding cost about 2.5x more prefill recompute (KV hit 0.78 vs 0.91 overall, 0.47 vs 0.91 on toggle turns).
- The coalescing step is not cosmetic. Demoting alone produces two consecutive `user` turns, which alternation-enforcing templates (Gemma 3, Mistral, Llama-2) reject with `Conversation roles must alternate`, trading one 500 for another.

`detect_requires_system_normalization` in `renderer/src/template/formatters.rs` (the probe and why those shapes), then `normalize_system_messages` in `renderer/src/template/oai.rs` (the three-step transform).

#### Validation

- `cargo test -p dynamo-renderer`: 109 passed. Five new hermetic tests cover permissive-untouched, strict-leading demotion, alternation coalescing, and the transform itself.
- Real-template audit (`adaptive_system_corpus_audit`, ignored by default): 126 chat templates pulled from popular served models, each rendered through the shipped probe across 7 agent-client message shapes. 11 templates flagged, 0 shape failures. Every unflagged template renders all shapes untouched (no probe false negatives) and every flagged template renders after normalization (normalization sufficient).
- Flagged families: Qwen3.5 and Qwen3.6-NVFP4 (reject a non-leading system), Gemma 3, Mistral 7B / Small / Nemo, and Ornith-1.0 (also reject consecutive users).
- End to end through Dynamo: loading Qwen3-0.6B with a strict template via `--custom-jinja-template` renders a mid-conversation-system request instead of returning 500, and the same request against the stock permissive template is unchanged.
- `cargo clippy --all-targets -D warnings` clean.

Two probe details were driven by the corpus rather than by design: no single "non-leading system" shape detects Gemma 3 (it accepts `[system, user, system]` but rejects `[system, user, assistant, system, user]`), and the probe must supply bos/eos or templates like Falcon3 and Zephyr false-flag on a missing-token raise.

The experiment scripts and the template corpus are not in this PR to keep it focused, and are available on request.

#### Related Issues

Addresses ai-dynamo/dynamo#11762. Generalizes the leading-system merge from ai-dynamo/dynamo#10283: that merge now happens here for strict templates, conditionally, and covers `/v1/messages`, `/v1/responses`, and `/v1/chat/completions` in one place.

### DIFF
--- a/renderer/src/template/formatters.rs
+++ b/renderer/src/template/formatters.rs
@@ -33,7 +33,12 @@ struct ProbeTokens {
 
 /// Unlike [`render_default_probe`], surfaces the error: a template's
 /// `raise_exception` is the signal that it rejects this message shape.
-fn probe_default_raises(env: &Environment, messages: serde_json::Value, tok: &ProbeTokens) -> bool {
+fn probe_template_raises(
+    env: &Environment,
+    template_name: &str,
+    messages: serde_json::Value,
+    tok: &ProbeTokens,
+) -> bool {
     let ctx = context! {
         messages => messages,
         add_generation_prompt => false,
@@ -41,7 +46,7 @@ fn probe_default_raises(env: &Environment, messages: serde_json::Value, tok: &Pr
         eos_token => tok.eos,
         unk_token => tok.unk,
     };
-    match env.get_template("default") {
+    match env.get_template(template_name) {
         Ok(t) => t.render(&ctx).is_err(),
         Err(_) => false,
     }
@@ -52,20 +57,35 @@ fn probe_default_raises(env: &Environment, messages: serde_json::Value, tok: &Pr
 ///
 /// All three shapes are probed because strict families disagree on which they
 /// reject: Qwen3.5 allows consecutive users, Gemma-3 allows a non-leading system.
-fn detect_requires_system_normalization(env: &Environment, tok: &ProbeTokens) -> bool {
+fn detect_requires_system_normalization(
+    env: &Environment,
+    template_name: &str,
+    tok: &ProbeTokens,
+) -> bool {
     let sys = |c: &str| json!({"role": "system", "content": c});
     let usr = |c: &str| json!({"role": "user", "content": c});
     let asst = |c: &str| json!({"role": "assistant", "content": c});
 
     // A template that can't render even this isn't rejecting shape, it's broken.
-    let baseline_ok = !probe_default_raises(env, json!([sys("s"), usr("u")]), tok);
+    let baseline_ok = !probe_template_raises(env, template_name, json!([sys("s"), usr("u")]), tok);
     if !baseline_ok {
         return false;
     }
-    let nonleading_system = probe_default_raises(env, json!([sys("s0"), usr("u"), sys("s1")]), tok);
-    let consecutive_users = probe_default_raises(env, json!([sys("s"), usr("u0"), usr("u1")]), tok);
-    let mid_after_assistant = probe_default_raises(
+    let nonleading_system = probe_template_raises(
         env,
+        template_name,
+        json!([sys("s0"), usr("u"), sys("s1")]),
+        tok,
+    );
+    let consecutive_users = probe_template_raises(
+        env,
+        template_name,
+        json!([sys("s"), usr("u0"), usr("u1")]),
+        tok,
+    );
+    let mid_after_assistant = probe_template_raises(
+        env,
+        template_name,
         json!([sys("s0"), usr("u"), asst("a"), sys("s1"), usr("u2")]),
         tok,
     );
@@ -545,8 +565,10 @@ impl HfTokenizerConfigJsonFormatter {
             eos: config.eos_tok(),
             unk: config.unk_tok(),
         };
-        let requires_system_normalization =
-            detect_requires_system_normalization(&env, &probe_tokens);
+        let default_requires_system_normalization =
+            detect_requires_system_normalization(&env, "default", &probe_tokens);
+        let tool_use_requires_system_normalization =
+            detect_requires_system_normalization(&env, "tool_use", &probe_tokens);
 
         Ok(HfTokenizerConfigJsonFormatter {
             env,
@@ -560,7 +582,8 @@ impl HfTokenizerConfigJsonFormatter {
             image_placeholder_template,
             default_template_handles_tool_calls_arguments_string,
             tool_use_template_handles_tool_calls_arguments_string,
-            requires_system_normalization,
+            default_requires_system_normalization,
+            tool_use_requires_system_normalization,
         })
     }
 }
--- a/renderer/src/template/formatters.rs
+++ b/renderer/src/template/formatters.rs
@@ -23,6 +23,67 @@ fn render_default_probe(env: &Environment, messages: serde_json::Value) -> Strin
         .unwrap_or_default()
 }
 
+/// Special tokens supplied to a probe render, so a template that appends
+/// `eos_token`/`bos_token` doesn't raise for a missing-token reason and get
+/// mistaken for a message-shape rejection.
+struct ProbeTokens {
+    bos: Option<String>,
+    eos: Option<String>,
+    unk: Option<String>,
+}
+
+/// Renders the `default` template and returns whether it raised. Unlike
+/// [`render_default_probe`] (which swallows errors), this surfaces the error so
+/// a `raise_exception` in the template — the signal that it rejects a given
+/// message shape — is observable at load time. bos/eos/unk are supplied so only
+/// message-shape raises (not missing-token raises) are detected.
+fn probe_default_raises(env: &Environment, messages: serde_json::Value, tok: &ProbeTokens) -> bool {
+    let ctx = context! {
+        messages => messages,
+        add_generation_prompt => false,
+        bos_token => tok.bos,
+        eos_token => tok.eos,
+        unk_token => tok.unk,
+    };
+    match env.get_template("default") {
+        Ok(t) => t.render(&ctx).is_err(),
+        // No `default` template compiled: nothing to normalize against.
+        Err(_) => false,
+    }
+}
+
+/// Detects at load time whether this template rejects the message streams agent
+/// clients produce, so `render` can conditionally normalize only when needed.
+///
+/// Different strict families fail on different shapes: Qwen3.5 rejects a
+/// non-leading `system` but accepts consecutive users; Gemma-3 accepts a
+/// non-leading `system` right after the first user yet rejects consecutive
+/// users and a `system` after an assistant turn; Mistral rejects both. So we
+/// probe several shapes and flag the template if ANY that a normalized stream
+/// would fix currently raises — while a plain `[system, user]` baseline still
+/// renders, to avoid flagging a genuinely broken template.
+fn detect_requires_system_normalization(env: &Environment, tok: &ProbeTokens) -> bool {
+    let sys = |c: &str| json!({"role": "system", "content": c});
+    let usr = |c: &str| json!({"role": "user", "content": c});
+    let asst = |c: &str| json!({"role": "assistant", "content": c});
+
+    let baseline_ok = !probe_default_raises(env, json!([sys("s"), usr("u")]), tok);
+    if !baseline_ok {
+        return false;
+    }
+    // Non-leading system, consecutive users, and a system after an assistant
+    // turn: the union catches strict-leading (Qwen3.5), alternation (Gemma,
+    // Mistral) and their combination.
+    let nonleading_system = probe_default_raises(env, json!([sys("s0"), usr("u"), sys("s1")]), tok);
+    let consecutive_users = probe_default_raises(env, json!([sys("s"), usr("u0"), usr("u1")]), tok);
+    let mid_after_assistant = probe_default_raises(
+        env,
+        json!([sys("s0"), usr("u"), asst("a"), sys("s1"), usr("u2")]),
+        tok,
+    );
+    nonleading_system || consecutive_users || mid_after_assistant
+}
+
 /// Detects if a template requires content as arrays (multimodal) vs strings (text-only).
 /// Returns true if the template only works with array format.
 fn detect_content_array_usage(env: &Environment) -> bool {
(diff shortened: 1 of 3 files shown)


KrishnanPrash:
@bloodeagle40234 tagging you here since this implements the approach we discussed in ai-dynamo/dynamo#11762.
The two things most worth your eyes, given you originally raised the tradeoff:

1. The demotion is in place rather than folded, for the KV cache reason you flagged.
2. The load-time probe is what decides whether to transform at all, so permissive templates are untouched.

renderer/src/template/formatters.rs
```diff
@@ -486,6 +547,19 @@ impl HfTokenizerConfigJsonFormatter {
...
+        // (non-leading system / consecutive users). Only when it does will
+        // `render` normalize; permissive templates render untouched. Supply the
+        // model's special tokens so a missing-token raise isn't misread as a
+        // message-shape rejection.
+        let probe_tokens = ProbeTokens {
+            bos: config.bos_tok(),
+            eos: config.eos_tok(),
+            unk: config.unk_tok(),
+        };
+        let requires_system_normalization =
```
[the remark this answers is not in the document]
KrishnanPrash: a68dac0 - Computes separate normalization flags for the default and tool-use templates and applies the selected flag during rendering.

renderer/src/template/oai.rs
```diff
@@ -428,6 +429,76 @@ impl OAIChatLikeRequest for dynamo_protocols::types::CreateChatCompletionRequest
...
+    }
+
+    // 3. Coalesce consecutive user turns into one.
+    let mut coalesced: Vec<serde_json::Value> = Vec::with_capacity(list.len());
+    for m in list.drain(..) {
+        if role_is(&m, "user") && coalesced.last().is_some_and(|p| role_is(p, "user")) {
+            let cur_text = content_of(&m);
+            let prev = coalesced.last_mut().unwrap();
+            let merged = format!("{}\n\n{}", content_of(prev), cur_text);
+            *prev = json!({"role": "user", "content": merged});
```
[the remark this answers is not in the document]
KrishnanPrash: a68dac0 - Preserves the original user message, multimodal content, and extra fields while appending demoted system text in place.

renderer/src/template/formatters.rs
```diff
@@ -23,6 +23,75 @@ fn render_default_probe(env: &Environment, messages: serde_json::Value) -> Strin
...
+
+/// Unlike [`render_default_probe`], surfaces the error: a template's
+/// `raise_exception` is the signal that it rejects this message shape.
+fn probe_template_raises(
+    env: &Environment,
+    template_name: &str,
+    messages: serde_json::Value,
+    tok: &ProbeTokens,
+) -> bool {
+    let ctx = context! {
```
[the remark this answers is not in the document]
KrishnanPrash: 10bc85c - Probes the default template with no tools and the tool-use template with a non-empty function tool.

renderer/src/template/oai.rs
```diff
@@ -472,26 +585,9 @@ impl OAIPromptFormatter for HfTokenizerConfigJsonFormatter {
...
-                    self.tool_use_template_handles_reasoning,
-                )
-            } else {
-                (
-                    "default",
-                    self.default_template_handles_tool_calls_arguments_string,
-                    self.default_template_handles_reasoning,
-                )
-            };
+        if requires_system_normalization {
```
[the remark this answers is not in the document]
KrishnanPrash: 10bc85c - Normalizes system messages before content-shape conversion so array-only templates receive arrays.

renderer/src/template/oai.rs
```diff
@@ -432,6 +429,95 @@ impl OAIChatLikeRequest for dynamo_protocols::types::CreateChatCompletionRequest
...
+    for m in list.drain(..) {
+        if role_is(&m, "user") && coalesced.last().is_some_and(|p| role_is(p, "user")) {
+            let cur_text = content_of(&m);
+            let prev = coalesced.last_mut().unwrap();
+            append_message_content_text(prev, cur_text);
+        } else {
+            coalesced.push(m);
+        }
+    }
+    *list = coalesced;
```
rmccorm4: just reiterating my understanding 
- for some templates (model examples: ?), multiple `system` messages are not allowed, so we replace them with `user` role
- for some templates (model examples: gemma, mistral), adjacent `user` messages are not allowed, so we join them together?

what do `vllm serve` and `sglang serve` do in these scenarios, if they have any handling for it?

renderer/src/template/formatters.rs
```diff
@@ -23,6 +23,82 @@ fn render_default_probe(env: &Environment, messages: serde_json::Value) -> Strin
...
+        tok,
+    );
+    let mid_after_assistant = probe_template_raises(
+        env,
+        template_name,
+        json!([sys("s0"), usr("u"), asst("a"), sys("s1"), usr("u2")]),
+        tools,
+        tok,
+    );
+    nonleading_system || consecutive_users || mid_after_assistant
```
keivenchang: nit: the three probes get `||`'d into one bit but `normalize_system_messages` does three independent things off it, so Qwen3.5 (rejects non-leading system, accepts consecutive users) also gets its user turns merged — `[system, user u0, user u1]` on your `STRICT_LEADING_TMPL` renders one user block instead of two. how about keeping the bits separate and gating each step on its own?

renderer/src/template/oai.rs
```diff
@@ -428,6 +429,95 @@ impl OAIChatLikeRequest for dynamo_protocols::types::CreateChatCompletionRequest
...
+            *m = json!({"role": "user", "content": content_of(m)});
+        }
+    }
+
+    // Demotion can leave two user turns adjacent, which alternation-enforcing
+    // templates (Gemma, Mistral) reject.
+    let mut coalesced: Vec<serde_json::Value> = Vec::with_capacity(list.len());
+    for m in list.drain(..) {
+        if role_is(&m, "user") && coalesced.last().is_some_and(|p| role_is(p, "user")) {
+            let cur_text = content_of(&m);
```
keivenchang: nit: `content_of` keeps only parts with a `text` key, so coalescing `[user(text), user(text + image_url)]` drops the image silently. you already fixed the mirror of this for the prev message in `a68dac0` — same splice-onto-the-array treatment would cover the current one.

keivenchang (approved):
approved — thanks Krishnan, the load-time probe plus the corpus audit is solid work. two nits inlined, neither blocking.

bloodeagle40234 (commented):
Thanks for driving this — I like the load-time-probe direction. Keying off actual template behavior instead of a model allowlist is the right call, and gating so permissive templates stay byte-identical is exactly what we'd want.

### Validation — Qwen3.6-27B-FP8, aggregated

I built this branch from source (renderer 1.3.8 + this PR, patched into an ai-dynamo 1.3.0 runtime wheel) and ran it in aggregated mode serving Qwen/Qwen3.6-27B-FP8. I then drove it with a real Claude Code session — multi-turn, including tool calls — i.e. the exact /v1/messages flow (agents/skills system block plus the toggling mid-conversation reminder) that returns 500 today. Every request returned 200, both plain-chat and tool-call turns, so the tool_use template path is exercised too. Works well end to end.

### Question — the single flag may over-normalize

requires_system_normalization is set by OR-ing three distinct probe results (nonleading_system || consecutive_users || mid_after_assistant), and when set, normalize_system_messages runs all three steps unconditionally — the individual results aren't preserved. normalize_merges_leading_run_and_coalesces shows this directly: [system, system, user, system] → [system("A\n\nB"), user("hi\n\nreminder")], i.e. the mid-conversation system is demoted and coalesced into the prior user turn.

For a template that rejects only a non-leading system but accepts consecutive users — the Qwen family I tested — that coalesce is stronger than the template requires:

[system, user, system]  --demote-->  [system, user, user]  --coalesce-->  [system, user(merged)]

Qwen would have rendered [system, user, user] fine, so we merge two user turns it would have accepted as-is. It's symmetric, too: a consecutive-users-only-strict template would still get its non-leading systems demoted unnecessarily.

Is the single-boolean-plus-full-bundle intentional for simplicity, or worth preserving the three probe results (a small struct / bitflags) and applying each step only when its own restriction fired — so we normalize exactly what a given template needs and no more? Not a blocker — just want to make sure we're not reshaping prompts (and prefix-cache boundaries) more than necessary for the common Qwen case.

rmccorm4:
@KrishnanPrash let's follow up on this one this week ^

KrishnanPrash:
@bloodeagle40234 you are correct: the initial version OR-ed the three probe results into one flag and ran every rewrite unconditionally. As of `20b47d2` the probe results are preserved as the two restrictions they actually measure (`demote_nonleading_system`, `coalesce_consecutive_users`), and each rewrite runs only when its own probe fired. An audit of a 126-template corpus confirms your case: 5 of the 11 strict templates, the Qwen3.5/3.6 and Ornith families, reject only a non-leading system and accept adjacent `user` turns, so they now get the demote step without coalescing.

renderer/src/template/oai.rs
KrishnanPrash: Close, but the first restriction is about position rather than count: a single leading system message is always accepted, and what raises is any system turn that is not first (rejected by Qwen3.5, Qwen3.6-27B-NVFP4, Qwen3.6-35B-A3B-NVFP4, Ornith-1.0-9B, and Ornith-1.0-35B), which we demote to `user` in place. Your second bullet is correct: Gemma 3 and the Mistral family (7B v0.2/v0.3, Small, Nemo) also reject adjacent `user` turns, so those get coalesced after the demotion. vLLM ([vllm-project/vllm#46025](https://github.com/vllm-project/vllm/pull/46025)) and SGLang ([sgl-project/sglang#28906](https://github.com/sgl-project/sglang/pull/28906)) both added a load-time template probe for this exact problem, but instead of converting to `user` they fold the inline system text into the leading system block and delete the inline turn, which never produces adjacent `user` turns; we keep the turn at its original position to preserve the KV prefix cache.

renderer/src/template/formatters.rs
KrishnanPrash: 20b47d2 - Gates each rewrite on its own probe result, so a template that only rejects a non-leading system keeps the consecutive user turns it accepts.

renderer/src/template/oai.rs
KrishnanPrash: 20b47d2 - Splices content arrays part for part when coalescing, so image parts on the turn being merged away survive.

renderer/src/template/formatters.rs
KrishnanPrash: 20b47d2 - Gates each rewrite on its own probe result, so a template that only rejects a non-leading system keeps the consecutive user turns it accepts.

KrishnanPrash:
@bloodeagle40234 you are correct: the initial version OR-ed the three probe results into one flag and ran every rewrite unconditionally. As of `20b47d2` the probe results are preserved as the two restrictions they actually measure (`demote_nonleading_system`, `coalesce_consecutive_users`), and each rewrite runs only when its own probe fired. An audit of a 126-template corpus confirms your case: 5 of the 11 strict templates, the Qwen3.5/3.6 and Ornith families, reject only a non-leading system and accept adjacent `user` turns, so they now get the demote step without coalescing.

renderer/src/template/oai.rs
```diff
@@ -432,6 +433,113 @@ impl OAIChatLikeRequest for dynamo_protocols::types::CreateChatCompletionRequest
...
+            {
+                m.insert("role".to_string(), json!("user"));
+            }
+        }
+    }
+
+    if rules.coalesce_consecutive_users {
+        let mut coalesced: Vec<serde_json::Value> = Vec::with_capacity(list.len());
+        for mut m in list.drain(..) {
+            if role_is(&m, "user") && coalesced.last().is_some_and(|p| role_is(p, "user")) {
```
rmccorm4, line 532: **Pre-existing behavior — non-blocking for merge:** Mistral-style tool templates validate a filtered role stream: they remove `tool` turns and assistant turns carrying `tool_calls` before enforcing alternation. This coalescer only checks physical adjacency, so after demoting `[system, user, assistant(tool_calls), tool, system]`, the template still sees logical `[user, user]` and raises. I reproduced this at the current head with the vLLM Mistral tool template. A follow-up could coalesce against the template-visible stream or add this tool-cycle shape to the probe.

renderer/src/template/formatters.rs
```diff
@@ -491,6 +571,25 @@ impl HfTokenizerConfigJsonFormatter {
         let tool_use_template_handles_tool_calls_arguments_string =
             template_handles_args_string("tool_use");
 
+        let probe_tokens = ProbeTokens {
+            bos: config.bos_tok(),
+            eos: config.eos_tok(),
+            unk: config.unk_tok(),
+        };
+        let default_probe_tools = Option::<serde_json::Value>::None;
```
rmccorm4, line 579: **Pre-existing behavior — non-blocking for merge:** The default template is probed with `tools=None`, but a valid request with `tools: []` selects `default` because `has_tools` is false while Jinja still receives `Some([])`. Templates using `tools is not none` can therefore enter an unprobed strict branch; I reproduced the current-head failure locally. A follow-up can canonicalize an empty tools array to `None` before both template selection and context construction.

renderer/src/template/oai.rs
KrishnanPrash: Will follow-up with a fix

renderer/src/template/formatters.rs
KrishnanPrash: Will follow-up with a fix

PR #166, opened 2026-07-30
feat(parsers): gemma4 on the v2 UnifiedParser — one ordered stream, with Dynamo request modes

keivenchang:
#### Overview:

This PR moves **Gemma4** onto the v2 `UnifiedParser` so reasoning, visible content, and tool calls stay in one ordered stream. Gemma4 only; every other family keeps its current path.

This is standalone on current `main`. #178, #174, and #194 already provide the shared parser interface, request modes, and guided-streaming contract.

Gemma4 conformance matrix

Case `UNIFIED.12.b` — a tool call opens inside a reasoning span:

```
before: reasoning = "I should check. <|tool_call>"   ← markup leaks into reasoning; the call is lost
after:  reasoning = "I should check. ", tool_call = get_weather, reasoning = " now answer"
```

#### Details:

- Gemma4 supplies its markers, emitter, and reasoning specification to the shared `ScannerUnified`; registration is one family row.
- `ReasoningSpec.start_label` strips Gemma4's structural `thought\n` role label instead of exposing it as reasoning text.
- Native recovery keeps lexical `call:` prose, skips wrapper-shaped data inside argument strings, and resynchronizes after malformed or incomplete wrappers.
- The current capture is `dynamo_v2-0.3.3+pr166`, separate from the released `0.3.3` identity. Historical columns use immutable sparse overlays, and released base hashes remain unchanged.

#### Verification:

- `cargo fmt -- --check`, parser clippy, and conformance clippy pass.
- The v2 parser package passes 450 unit tests plus its integration and doc-test binaries.
- The full conformance package passes. `python3 -m pytest conformance/utils/tests` passes 191 tests.
- The generated v2 matrix contains all 292 unified cases with no `postdates that build` placeholders.
- The archive audit verifies 28 manifest hashes and sizes, four released hashes, 12 sparse counts, and the four removed archive names.

`parsers/v2/src/unified/gemma4.rs`, then `parsers/v2/src/tool_calling/scan.rs` and `parsers/v2/src/tool_calling/v1core/gemma4/parser.rs`

/coderabbit profile chill


parsers/v2/src/tool_calling/gemma4.rs
```diff
@@ -7,299 +7,114 @@
...
+            block_starts: vec![TOOL_CALL_START.to_string()],
+            // Same token as `invoke_end`: closing the call closes the block.
+            block_ends: vec![TOOL_CALL_END.to_string()],
+            invoke_start: CALL_PREFIX.to_string(),
+            invoke_end: TOOL_CALL_END.to_string(),
+            orphan_markers: vec![TOOL_CALL_END.to_string()],
+            // The string delimiter is held back too: a `<|"|>` split across a
+            // chunk boundary must not be released as text, or the value it opens
+            // is read as structure.
+            holdback_markers: vec![
```
rmccorm4, line 299: [P1] Keep lexical holdback prefixes out of the unconditional control-marker set

`WrappedBlockScanner::control_markers()` returns this entire `holdback_markers` list, and the guided path strips every entry as structural markup before the grammar-aware `InvokeScan::opens` decision can reject ordinary prose. Adding the ambiguous `call:` prefix here therefore corrupts user-visible output in guided mode.

I reproduced this through the public registry on this head:

- reasoning `I will call: you` became `I will  you` before a valid guided JSON call;
- recovered visible text split as `I will ca` + `ll: you tomorrow` became `I will  you tomorrow`.

Both native probes are protected by `opens_bare_call`, but guided parsing bypasses that protection via the regular control-marker match. Please separate unconditional structural markers from lexical boundary-holdback candidates and let `InvokeScan` own semantic recognition of `call:`. Add whole-marker and split-marker guided regressions for ordinary prose.

parsers/v2/src/tool_calling/v1core/gemma4/parser.rs
```diff
@@ -201,6 +201,7 @@ fn has_bare_call_body_start(input: &str) -> bool {
 /// adjacency at the real object close, so a bare `<tool_call|>` literal embedded
 /// inside a `<|"|>` string value does not false-trigger a "section complete"
 /// signal here — matches upstream.
+#[allow(dead_code)] // Retained for whole-message consumers while streaming types one invoke.
```
rmccorm4: [P3] Do not retain a private second Gemma orchestration path

After this migration, the active v2 parser uses `WrappedBlockScanner` with `find_leading_tool_call_end_gemma4` and `parse_one_tool_call_gemma4`. This whole-message entry point, `try_tool_call_parse_gemma4`, and their recovery/text-assembly helpers have no production v2 caller. Because `v1core` is behind a private module, making these functions `pub` inside it does not create an external v2 contract.

The new `allow(dead_code)` annotations hide a second implementation of call-span scanning, malformed-call recovery, and surrounding-text assembly—the same responsibilities now owned by `WrappedBlockScanner`. Keeping both makes future fixes liable to land in only one path and silently diverge.

rmccorm4: Recommended resolution: keep only the low-level helpers used by the shared scanner—`detect_tool_call_start_gemma4`, `find_leading_tool_call_end_gemma4`, `is_call_prefix_boundary`, and `parse_one_tool_call_gemma4`—and remove the unused whole-message coordinator plus its recovery/text-gap helpers and tests. If v2 needs a batch API later, expose one deliberately and implement it by passing the complete input through this same scanner followed by `finish()`; that preserves one behavioral owner instead of reviving a parallel parser.

parsers/v2/src/tool_calling/gemma4.rs
```diff
@@ -7,299 +7,114 @@
...
+const GEMMA4_INVOKE_SCAN: InvokeScan = InvokeScan {
+    end: find_leading_tool_call_end_gemma4,
+    opens: opens_bare_call,
+    holdback: partial_bare_opener_suffix_len,
+};
+
+/// Build the Gemma 4 scanner. ONE construction site, shared by the tool-only
+/// parser below and by the unified parser, so the grammar cannot drift between
+/// the two surfaces.
+pub(crate) fn gemma4_scanner(tools: &[Tool]) -> WrappedBlockScanner<Gemma4InvokeEmitter> {
```
rmccorm4, line 284: [P3] Update the shared-scanner adoption map for this migration

This new construction site makes Gemma a `WrappedBlockScanner` family, with `InvokeScan` supplying the balanced/string-aware behavior that static markers cannot express. However, the module-level architecture documentation in `tool_calling/scan.rs:20-25` still lists only qwen3_coder, minimax_m2, minimax_m3, and kimi_k2, and explicitly says Gemma keeps a bespoke drain.

That stale map will send future Gemma fixes to the wrong layer. Please add Gemma to the wrapped-family list and describe `InvokeScan` as the extension point for its grammar-aware boundaries; leave only the families that genuinely retain bespoke drains in the exception list.

parsers/v2/src/tool_calling/gemma4.rs
```diff
@@ -7,299 +7,114 @@
...
+                CALL_PREFIX.to_string(),
+                STRING_DELIM.to_string(),
+            ],
+            // No outer wrapper survives a recovered bare call, so later narration
+            // is the user's text again.
+            bare_recovery_latch: BareRecoveryLatch::Clear,
+            invoke_latch: InvokeLatch::IfEmitted,
+            drop_invoke_crossing_block_end: false,
+            invoke_scan: Some(GEMMA4_INVOKE_SCAN),
+            preserve_special_tokens: true,
```
rmccorm4, line 310: [P3] Delegate special-token policy to the scanner-owned grammar

This spec now owns `preserve_special_tokens: true`, and `WrappedBlockScanner::preserve_special_tokens()` is the intended single policy source used to keep grammar and reasoning requirements aligned. However, `Gemma4ToolStreamParser::preserve_special_tokens` below still independently hard-codes `true`. That creates a second value that can drift from the unified surface—the exact split the `WrappedBlockSpec` documentation says grammar ownership prevents.

Please delegate from the adapter:

```rust
fn preserve_special_tokens(&self) -> bool {
    self.scanner.preserve_special_tokens()
}
```

parsers/v2/src/tool_calling/gemma4.rs
keivenchang: fixed in ca176671 — guided mode now preserves ambiguous `call:` prose, with whole-input and every-split regression coverage

parsers/v2/src/tool_calling/v1core/gemma4/parser.rs
keivenchang: fixed in ca176671 — removed the unused whole-message coordinator and kept only the grammar helpers used by the stream scanner

parsers/v2/src/tool_calling/gemma4.rs
keivenchang: fixed in ca176671 — the architecture map now puts Gemma under `WrappedBlockScanner` with `InvokeScan` for grammar-aware decisions

parsers/v2/src/tool_calling/gemma4.rs
keivenchang: fixed in ca176671 — `preserve_special_tokens()` now delegates to the scanner

conformance/utils/src/gen_unified_golden.py
```diff
@@ -1130,30 +1140,57 @@ def _guided_product():
             golden = ([{"kind": "tool_call", "name": "get_weather",
                         "arguments": want_args}] if dispatches else
                       [{"kind": "text", "text": None}])
+            if sur_name == "bare_opener":
+                # Gemma's `call:` is also ordinary lexical text. It is control
+                # markup only when InvokeScan accepts a native `call:NAME{...}`
+                # body; the guided JSON payload shapes here never satisfy that
+                # grammar, so the rejected prefix stays in the text contract.
+                golden = golden_by_family(
```
rmccorm4: [P1] Do not redefine a valid guided call as text to match the parser

This family override changes the Gemma `bare_opener` contract from a `tool_call` to `text`, but no parser code changed. On this exact head, a public-API probe with `call:[{"name":"get_weather","arguments":{"city":"Paris"}}]` returns the entire stream as visible text and never dispatches the valid call. This makes conformance green by moving the oracle to the regression, so a required guided call can silently become content. Keep `call:` out of unconditional marker stripping, but add a scanner-owned guided-prefix probe (`NoMatch` / `Pending` / `Match`) that recognizes an immediately adjacent guided JSON payload while preserving lexical prose such as `call: you`; then restore the tool-call golden and test every split point independently.

keivenchang: fixed in 38485be6 — the scanner now owns NoMatch/Pending/Match for adjacent guided payloads, and every valid UTF-8 split preserves prose while dispatching valid call: JSON.

parsers/v2/src/unified/mod.rs
```diff
@@ -1146,7 +1147,18 @@ fn guided_holdback_len(
         .unwrap_or(0);
     let pending_invoke = invoke_control
         .map(|(start, scan)| {
-            let partial = (scan.holdback)(input);
+            let guided = input
+                .char_indices()
```
rmccorm4: This probes `guided_prefix` at every character offset. Gemma’s hook rescans `text[..at]`, making guided holdback O(n²); a 32 KiB probe took ~816 ms. 

`Pending` can only occur within the trailing `invoke_start.len()` bytes, so iterate backward over that bounded suffix only. 

The same probe then takes ~0.84 ms without changing parsing semantics.

conformance/utils/lib/parsers/UNIFIED_CASES.md
```diff
@@ -54,11 +54,11 @@ The parser recovers everything it can and NEVER drops valid text, leaks markup,
 
 `MATCH` (green) · `ORDER` / `MERGE` / `LOSS` (the unification gap) · `LEAK` (markup in text, `↯`) · `ARG_MISMATCH` / `WHITESPACE` (version drift) · `ERROR` (engine hard-errored where the spec expects graceful output).
 
-The Dynamo column is a per-family MIXTURE: `qwen3` and `muse_glimmer` run the native `UnifiedParser`; `gemma4` and `kimi_k2` still run the v1-reasoning + v2-tool split and carry the gap. Every remaining red Dynamo cell in this tab is a split-path cell.
+The Dynamo column is a per-family MIXTURE: `qwen3`, `gemma4`, and `muse_glimmer` run the native `UnifiedParser`; `kimi_k2` still runs the v1-reasoning + v2-tool split and carries the gap. Every remaining red Dynamo cell in this tab is a split-path cell.
```
rmccorm4: Conformance architecture documentation still says Kimi uses the split path

[UNIFIED_CASES.md:57](https://github.com/ai-dynamo/frontend-crates/blob/7900f341c1aacf5e1e7a1eea4fbb91f764a02597/conformance/utils/lib/parsers/UNIFIED_CASES.md#L57) contradicts both `parser_families.yaml` and `unified_render.rs`, which correctly state that all four corpus families now use native unified parsers. Update it to prevent misinterpreting future red cells as expected split-path failures.

parsers/v2/tests/vendor_registry.rs
```diff
@@ -192,3 +194,153 @@ fn unknown_family_error_explains_how_to_register() {
...
+        let mut deltas = Vec::new();
+        for chunk in chunks {
+            deltas.extend(parser.push(chunk).expect("push"));
+        }
+        deltas.extend(parser.finish().expect("finish").events);
+        assert_eq!(assemble(&deltas), expected, "split {split:?}");
+    }
+}
+
+/// Gemma's lexical `call:` prefix is only structural when the grammar-aware
```
rmccorm4: [vendor_registry.rs:244](https://github.com/ai-dynamo/frontend-crates/blob/7900f341c1aacf5e1e7a1eea4fbb91f764a02597/parsers/v2/tests/vendor_registry.rs#L244-L308) tests Gemma guided parsing, not registry behavior, and unnecessarily takes the process-global serialization lock. Move these to a dedicated Gemma guided integration test.

parsers/v2/src/unified/mod.rs
keivenchang: fixed in 1ed18a9d — guided holdback now probes only the bounded trailing invoke-prefix window, with deterministic callback-count coverage

conformance/utils/lib/parsers/UNIFIED_CASES.md
keivenchang: fixed in 1ed18a9d — all current taxonomy, generator, renderer, and feed descriptions now identify the four corpus families as native unified

parsers/v2/tests/vendor_registry.rs
keivenchang: fixed in 1ed18a9d — Gemma guided behavior now has its own integration-test binary and no longer takes the registry serialization lock

keivenchang:
Pushed `e3df430a` with the remaining Gemma4 v2 review fixes and capture updates.

The parser now preserves lexical `call:` prose and resynchronizes after malformed or incomplete Gemma wrappers in both tool-only and unified modes at every UTF-8 split. The capture set now keeps released bases immutable, fills historical columns through sparse overlays, and records this PR under `0.3.3+pr166`.

Local validation passes: format and clippy, 450 parser unit tests plus integrations/docs, the full conformance package, 191 Python utility tests, and a 292-case matrix with no historical placeholders. CI for `e3df430a` is still in progress. I published the three prepared replies and left all eight reviewer threads open for Ryan.

parsers/v2/src/tool_calling/v1core/gemma4/parser.rs
```diff
@@ -131,360 +130,148 @@ fn parse_recoverable_call_at(
...
-        for &(start, end) in &removed_spans {
-            // Spans are non-overlapping and sorted; guard against any pathological
-            // overlap so we never slice on a stale offset.
-            if start >= prev_end {
-                push_natural_text_gap(&mut kept, &message[prev_end..start]);
+            let Some(open_brace) = open_brace else {
+                cursor += TOOL_CALL_START.len();
+                continue;
+            };
+            let Some(close_brace) = find_balanced_args_end(rest, open_brace) else {
```
rmccorm4: [P1] Keep Gemma resynchronization linear

Each unmatched wrapper calls `find_balanced_args_end` through the remaining suffix, then advances only past that wrapper and rescans the same bytes for the next candidate. Repeated incomplete wrappers scaled roughly 4× per 2× input; 19 KiB took ~40 ms in one release-mode push. Replace the nested suffix scans with one forward state machine that tracks string state, the latest structural wrapper, and brace depth.

parsers/v2/src/tool_calling/scan.rs
```diff
@@ -975,6 +1003,19 @@ impl<E: InvokeEmitter> WrappedBlockScanner<E> {
                     self.buffer.drain(..start);
                 }
                 let Some(end) = self.invoke_end_at(&self.buffer, flush) else {
+                    if let Some(next_block) = self.resync_block_start(&self.buffer) {
```
rmccorm4: [P1] Preserve Gemma scan state across pushes

`invoke_end_at` restarts Gemma’s end scan from byte zero on every append; while incomplete, this immediately runs a second full-buffer resynchronization scan. In release, a valid 32 KiB string argument took 1.11 s with 4-byte chunks and 255 ms with 16-byte chunks. An unfinished 32 KiB `call:NAME` took 1.74 s and 489 ms respectively. Carry the active invoke’s cursor, brace depth, string state, and header phase across pushes so each appended byte is scanned once.

keivenchang: fixed in `8fcd1dfd` — `WrappedBlockScanner` now retains Gemma’s call-header, brace-depth, string-delimiter, and completed-body state across `push()` calls, so it only scans newly appended bytes instead of restarting from byte zero. Recovery checks only the new closer-marker window and runs the grammar-aware recovery scan once a later call can actually be complete. The 32 KiB 4-byte reproduction dropped from 565 ms to 8 ms; the unfinished shape dropped from 572 ms to 10 ms.

parsers/v2/src/tool_calling/v1core/gemma4/parser.rs
keivenchang: fixed in `8fcd1dfd` — `find_complete_wrapped_call_after_gemma4` is now a single forward walker that tracks quoted strings and the current candidate’s brace depth. A later structural wrapper replaces an unfinished candidate, so repeated malformed wrappers no longer call `find_balanced_args_end` over overlapping suffixes.

parsers/v2/src/tool_calling/scan.rs
```diff
@@ -543,10 +655,18 @@ pub(crate) struct WrappedBlockScanner<E: InvokeEmitter> {
...
+    gemma_invoke: Option<GemmaInvokeProgress>,
+    /// Bytes already checked for a new Gemma recovery closer. A later wrapped
+    /// call cannot be complete until its closer arrives, so this avoids a
+    /// suffix-wide recovery probe for ordinary argument chunks.
+    gemma_resync_checked: usize,
 }
 
 impl<E: InvokeEmitter> WrappedBlockScanner<E> {
     pub(crate) fn new(spec: WrappedBlockSpec, emitter: E) -> Self {
+        let gemma_invoke = (spec.family == "gemma4").then(GemmaInvokeProgress::default);
```
indrajit96: ## Move the Gemma-specific streaming parser out of the shared scanner

`family` is documented as a logging label, but this line uses the literal `"gemma4"` to enable parsing behavior. Renaming that label—for example, to `"gemma-4"`—would silently disable Gemma’s incremental parser even though the grammar did not change.

The shared module also contains `GemmaInvokeProgress::end`, which duplicates the Gemma boundary logic already exposed through `GEMMA4_INVOKE_SCAN.end`. For example, both implementations must understand that `<tool_call|>` inside a `<|"|>` string is data, while the same token after the closing `}` ends the call. Normal streaming uses the shared implementation, while fallback paths can use the family hook. A future fix applied to only one implementation could therefore make parsing depend on how the response was chunked.

## Suggested fix

Make incremental boundary state a capability supplied by each grammar. For example, the shared module could define a family-neutral interface:

```rust
trait InvokeBoundaryState {
    fn end(
        &mut self,
        text: &str,
        flush: bool,
        tool_index: usize,
    ) -> Option<usize>;

    fn resync(
        &mut self,
        text: &str,
        tool_index: usize,
    ) -> Option<usize> {
        None
    }

    fn reset(&mut self);
}
```

`WrappedBlockSpec` could provide a factory because each stream needs fresh mutable state:

```rust
type InvokeBoundaryFactory = fn() -> Box<dyn InvokeBoundaryState>;

struct WrappedBlockSpec {
    // Existing grammar configuration...
    invoke_boundary: Option<InvokeBoundaryFactory>,
}
```

The shared scanner would store and use that interface generically:

```rust
struct WrappedBlockScanner<E> {
    // Existing shared state...
    invoke_boundary: Option<Box<dyn InvokeBoundaryState>>,
}
```

Then `gemma4.rs` would own the concrete implementation:

```rust
#[derive(Default)]
struct GemmaInvokeBoundary {
    cursor: usize,
    name_started: bool,
    depth: usize,
    in_string: bool,
    body_end: Option<usize>,
    invalid: bool,
}

impl InvokeBoundaryState for GemmaInvokeBoundary {
    // Gemma-specific end/resync/reset behavior
}
```

The Gemma specification would explicitly supply it:

```rust
invoke_boundary: Some(|| {
    Box::new(GemmaInvokeBoundary::default())
}),
```

With this structure, `WrappedBlockScanner` only creates, calls, and resets the supplied boundary state. It no longer needs `GemmaInvokeProgress`, `family == "gemma4"`, or knowledge of `call:`, `<|"|>`, and `<tool_call|>`. Gemma’s grammar remains in `gemma4.rs`, and there is one authoritative implementation shared by normal streaming and fallback paths.

keivenchang: Moved the mutable Gemma boundary state into `Gemma4InvokeEmitter` in `deb6f85f`. `WrappedBlockScanner` now owns only generic lifecycle and invokes family hooks for incremental boundaries, resync readiness, and reset. The Gemma emitter owns the cursor, brace depth, string state, and closer window; tool-only and unified adapters still share that one scanner. Full Rust CI-equivalent validation passes locally.

parsers/v2/src/unified/mod.rs
```diff
@@ -1607,22 +1646,37 @@ fn guided_holdback_len(
...
-                })
-                .max()
-                .unwrap_or(0);
-            partial.max(body)
+            let guided = guided_pending_suffix_len(input, start, scan.guided_prefix);
+            // A lexical prefix directly before an incomplete reasoning marker is
+            // undecided until the marker finishes. Releasing it first makes a
+            // whole-input parse strip malformed `call:` syntax while a split parse
+            // exposes it as visible text.
+            let before_reasoning_marker = if start == "call:" {
```
indrajit96: ### Keep Gemma’s `call:` policy out of the generic guided parser

`GuidedState` is described as grammar-independent, but the shared guided path now checks the literal Gemma prefix `"call:"` here and again in `control_marker_at`. It also contains Gemma-specific rules for identifier characters and for deciding when only the `call:` portion of malformed syntax should be stripped.

These are valid Gemma behaviors, but placing them in `unified/mod.rs` makes the common state machine depend on one family’s token spelling and malformed-input policy. A future family with a similar lexical prefix would require another family-specific branch in this shared function.

### Suggested fix

Extend the family hook so it classifies a guided prefix using context supplied by the common parser:

```rust
struct GuidedPrefixContext {
    at_payload_boundary: bool,
    in_reasoning: bool,
    followed_by_reasoning_marker: bool,
}

enum GuidedPrefixAction {
    Text,
    Pending,
    Consume,
    Strip,
}

type GuidedPrefixPolicy =
    fn(text: &str, at: usize, context: GuidedPrefixContext) -> GuidedPrefixAction;
```

Gemma can implement this policy in `gemma4.rs`. The shared guided parser would calculate the context and apply the returned action without comparing `invoke_start` to `"call:"` or checking Gemma’s allowed name characters.

parsers/v2/src/unified/mod.rs
```diff
@@ -1607,22 +1646,37 @@ fn guided_holdback_len(
...
+                0
+            };
+            // Native invocation prefixes are a separate grammar concern. Keep
+            // their existing holdback, but let `guided_prefix` exclusively own
+            // the guided prefix's Match/Pending distinction.
+            let native = (scan.holdback)(input);
+            // `flush: false` always makes `tool_index` inert (Kimi's EOF-only
+            // recovery gate can only fire when `flush` is true). Zero preserves
+            // the pre-index behavior of this holdback-only probe.
+            let body = pending_native_invoke_len(input, start, scan.opens, scan.end);
```
indrajit96: ### Reuse the stateful invoke scanner in guided mode

The native path now keeps `Gemma4InvokeProgress` between chunks, but the guided path receives only a copied `InvokeScan` containing stateless function pointers. This call therefore runs `scan.end` over the entire retained native invoke whenever another chunk arrives.

On the current head, a diagnostic probe streaming a native Gemma invoke in 4-byte chunks took approximately 0.43, 1.66, and 6.44 seconds for 8, 16, and 32 KiB argument values. Doubling the payload causes approximately four times the work, which shows that this path is quadratic.

### Example

For an invoke arriving four bytes at a time:

```text
call:get_weather{city:<|"|>a large value...<|"|>}<tool_call|>
```

the guided path effectively scans:

```text
push 1: first 4 bytes
push 2: first 8 bytes
push 3: first 12 bytes
...
```

The native path received incremental state specifically to avoid this repeated prefix parsing, but `GuidedState` cannot use that state.

### Suggested fix

Expose a request-local boundary-state factory through the grammar and give each `GuidedState` a fresh instance of the same incremental scanner used by native mode:

```rust
struct GuidedState {
    // Existing request state...
    invoke_boundary: Option<Box<dyn InvokeBoundaryState>>,
}
```

Create it during `GuidedState::new`, advance it as new bytes arrive, and reset it when the candidate is consumed, rejected, or the request is reset. `pending_native_invoke_len` should no longer call a stateless `end` function over the complete growing suffix.

The current native implementation stores its progress on `Gemma4InvokeEmitter`. That is sufficient for native mode, but guided mode bypasses the emitter, so the incremental state must be factored into a boundary object that both paths can instantiate.

Please also add a regression that measures total bytes examined across a sequence of pushes. A timing assertion is unnecessary; an instrumented boundary implementation can prove that total work grows linearly with the input.

parsers/v2/src/unified/gemma4.rs
```diff
@@ -0,0 +1,718 @@
...
+        let mut deltas = Vec::new();
+        for chunk in chunks {
+            deltas.extend(parser.push(chunk).expect("push"));
+        }
+        deltas.extend(parser.finish().expect("finish").events);
+        assemble(&deltas)
+    }
+
+    fn reasoning(text: &str) -> UnifiedEvent {
+        UnifiedEvent::Reasoning { text: text.into() }
```
indrajit96, line 185: ### Make Gemma resynchronization stateful and string-aware

Moving the recovery gate into `gemma4.rs` fixes the shared-layer ownership concern, but `resync_ready` still searches newly appended bytes for any literal `<tool_call|>` and then causes the stateless `find_complete_wrapped_call_after_gemma4` hook to scan the retained buffer from the beginning.

The readiness check is not string-aware. A `<tool_call|>` occurrence inside a `<|"|>` string is data, but it still triggers another complete resynchronization scan. On the current head, inputs of approximately 3.3, 6.7, 13.3, and 26.7 KiB containing repeated string-data closers took approximately 11, 40, 160, and 622 ms. This is quadratic scaling.

The readiness hook also has no `flush` input. Consequently, it cannot activate Gemma’s missing-end EOF recovery for a later balanced candidate after an earlier malformed call.

### Example

In this shape, every `<tool_call|>` is data inside an unterminated string, but every occurrence triggers another full recovery scan:

```text
<|tool_call>call:broken{note:<|"|>
x<tool_call|>x<tool_call|>x<tool_call|>...
```

In this shape, the later call has a balanced body and is eligible for Gemma’s missing-close recovery, but the current parser emits no call because no later literal closer makes `resync_ready` return `true`:

```text
<|tool_call>call:broken{note:<|"|>closed<|"|>
<|tool_call>call:get_weather{city:<|"|>NYC<|"|>}
```

### Suggested fix

Fold readiness and recovery into one stateful Gemma operation instead of combining a boolean marker gate with a stateless full-buffer callback:

```rust
trait InvokeBoundaryState {
    fn resync(
        &mut self,
        text: &str,
        flush: bool,
        tool_index: usize,
    ) -> Option<usize>;
}
```

`Gemma4InvokeProgress` can retain a recovery cursor, string state, candidate start, and brace depth. It can then inspect only newly appended bytes, ignore closer-shaped data while inside a string, and consider a balanced missing-close candidate when `flush` is true.

Please add deterministic regressions for both properties: total bytes examined must remain linear when a string contains repeated closer tokens, and a later balanced call after a malformed block must receive the same missing-end recovery at EOF as a leading call.

parsers/v2/src/unified/mod.rs
```diff
@@ -3399,6 +3513,51 @@ mod tests {
...
+    }
+
+    #[test]
+    fn native_pending_probe_stops_at_the_first_qualifying_opener() {
+        let input = "call:".repeat(32 * 1024);
+        let open_probes = Cell::new(0);
+        let end_probes = Cell::new(0);
+
+        assert_eq!(
+            pending_native_invoke_len(
```
indrajit96: ### Test total streaming work rather than callback count

This test proves that `pending_native_invoke_len` invokes `end` once during one call. It does not bound the amount of input examined by that callback, and it does not model the caller invoking this helper again after every streamed chunk.

The one callback receives the entire retained suffix. As a result, the guided path can perform quadratic total work while this test continues to pass; that is exactly what happens with large Gemma invokes on the current head.

### Example

The test currently checks only one row:

```text
one helper call -> one end callback
```

The real streaming behavior is:

```text
4-byte buffer     -> scan 4 bytes
8-byte buffer     -> scan 8 bytes
12-byte buffer    -> scan 12 bytes
...
32-KiB buffer     -> scan the whole 32 KiB again
```

Counting callbacks inside one helper invocation cannot distinguish those behaviors.

### Suggested fix

Exercise a sequence of growing pushes and count the total number of bytes examined by the boundary scanner. For example:

```rust
let examined = Cell::new(0usize);

for end in (4..=input.len()).step_by(4) {
    pending_native_invoke_len(
        &input[..end],
        "call:",
        |_, _| true,
        |suffix, _, _| {
            examined.set(examined.get() + suffix.len());
            None
        },
    );
}

assert!(examined.get() <= LINEAR_FACTOR * input.len());
```

After introducing stateful guided scanning, it would be better to instrument that boundary state through `GuidedState` directly. The important assertion is amortized linear work over the stream, not one callback per individual holdback calculation.

parsers/v2/src/tool_calling/kimi_k2.rs
```diff
@@ -458,10 +458,20 @@ fn kimi_invoke_holdback(_text: &str) -> usize {
     0
 }
 
+/// Kimi's markers are unambiguous structural tokens with no ordinary-prose
+/// collision (unlike Gemma 4's lexical `call:` prefix), so it has no guided
+/// prefix of its own: always `NoMatch`, which the guided drain treats
+/// identically to not having `invoke_scan` set at all.
+fn kimi_no_guided_prefix(_text: &str, _at: usize) -> GuidedPrefix {
```
indrajit96: ### Split optional family capabilities instead of requiring no-op implementations

Adding Gemma’s lexical guided-prefix and malformed-block recovery requirements forces Kimi to provide an always-`NoMatch` `guided_prefix` function and `resync: None`. This is evidence that `InvokeScan` is combining independent capabilities rather than representing one cohesive contract.

This makes the common interface grow whenever one family needs a new behavior, and every unrelated family must then add another no-op implementation.

### Example

Kimi has a real invoke boundary scanner, but no lexical guided prefix and no Gemma-style resynchronization policy:

```rust
fn kimi_no_guided_prefix(_text: &str, _at: usize) -> GuidedPrefix {
    GuidedPrefix::NoMatch
}

const KIMI_INVOKE_SCAN: InvokeScan = InvokeScan {
    // Real Kimi boundary operations...
    guided_prefix: kimi_no_guided_prefix,
    resync: None,
};
```

The no-op function exists only because the combined struct requires it.

### Suggested fix

Separate the required boundary behavior from optional guided policy:

```rust
struct InvokeCapabilities {
    boundary_factory: InvokeBoundaryFactory,
    guided_policy: Option<GuidedPrefixPolicy>,
}
```

The boundary state can provide `end`, optional `resync`, and `reset`. Gemma can additionally provide `guided_policy: Some(...)`, while Kimi can use `guided_policy: None` without implementing Gemma-driven no-op hooks.

The exact representation can remain static or generic if allocation is a concern; the important part is that unrelated families do not have to implement capabilities they do not use.

indrajit96 (commented):
@keivenchang thanks for adding this
Left a few comments on streaming changes with a possible perf issue.
SOme gemma specific behavior is still present in `unified/mod.rs`
The existing tests all pass.

parsers/v2/src/unified/mod.rs
keivenchang: Moved the lexical prefix decision into Gemma’s optional guided policy in `bb56d9a3`. The shared guided parser now passes family-neutral context and applies the returned action; it no longer compares against `call:` or applies Gemma’s identifier rules.

parsers/v2/src/unified/mod.rs
keivenchang: `bb56d9a3` gives guided parsing its own request-local Gemma boundary from the same factory as native parsing. It advances only appended candidate bytes, and the new one-byte 8/16/32 KiB guided regression asserts the boundary examines at most `2N` bytes.

parsers/v2/src/unified/gemma4.rs
keivenchang: Folded Gemma recovery into the stateful boundary in `bb56d9a3`. It now retains recovery cursor, string state, and candidate depth, ignores closer-shaped string data, and recovers a later balanced missing-close candidate at EOF. The public native recovery tests cover the malformed-first and later-balanced EOF path at chunk sizes 1, 4, and 16.

parsers/v2/src/unified/mod.rs
keivenchang: Replaced the callback-count coverage in `bb56d9a3` with public one-byte streaming tests at 8, 16, and 32 KiB. They count Gemma boundary bytes examined across the complete push sequence for tool-only native, unified native, and guided parsing, with a `2N` bound.

parsers/v2/src/tool_calling/kimi_k2.rs
keivenchang: Split guided prefix behavior from the invoke boundary contract in `bb56d9a3`. Kimi now supplies its real stateless boundary only and omits the Gemma-only guided policy; its existing guided split behavior remains covered by the public factory regression.

parsers/v2/src/unified/gemma4.rs
indrajit96: ### [P2] Preserve a balanced recovery candidate across trailing whitespace

Thanks—`bb56d9a3` fixes the later balanced missing-close case when the closing `}` is the final byte. The current public regression ends exactly there:

```rust
let later_balanced =
    "<|\"|>}<|tool_call>call:get_weather{city:<|\"|>NYC<|\"|>}";
```

In that case, the loop exits with `resync_candidate = Some((start, 0))`, so the final `flush` returns it. However, the same candidate is still lost if whitespace follows the balanced body before EOF.

### Example

```text
<|tool_call>call:broken{note:<|"|>unterminated<|"|>}<|tool_call>call:get_weather{city:<|"|>NYC<|"|>}   \n\t
```

After the later call’s `}`, the candidate has depth zero. The first following space causes another loop iteration; because it is neither a complete nor partial `<tool_call|>`, the unconditional assignment clears `resync_candidate`. The final `flush` then returns `None`.

On `c0591ba`, both public adapters consequently emit zero calls. Removing only the trailing whitespace makes the existing test pass, so this is an uncovered suffix of the recovery implemented in `bb56d9a3`, not a duplicate of the fixed case. The leading-call path already treats whitespace-only EOF suffixes as recoverable through `after.trim().is_empty()`.

### Suggested fix

Preserve a depth-zero candidate while consuming whitespace and clear it only when a non-whitespace byte cannot continue the closer:

```rust
if depth == 0 {
    if rest.starts_with(TOOL_CALL_END) {
        return Some(start);
    }
    if TOOL_CALL_END.starts_with(rest) {
        return None;
    }
    if !rest.chars().next()?.is_whitespace() {
        self.resync_candidate = None;
    }
}
```

The existing flush path will then return the retained candidate at EOF. Please add:

```rust
#[test]
fn later_balanced_call_with_trailing_whitespace_recovers_at_eof() {
    let malformed =
        "<|tool_call>call:broken{note:<|\"|>unterminated";
    let later_balanced =
        "<|\"|>}<|tool_call>call:get_weather{city:<|\"|>NYC<|\"|>}   \n\t";
    let input = format!("{malformed}{later_balanced}");

    assert_both_adapters_at_chunk_sizes(&input, &[1, 4, 16], 1);
}
```

I validated this exact change locally: the regression fails with zero calls on the reviewed head and passes through both adapters at chunk sizes 1, 4, and 16 after the fix.

indrajit96 (approved):
LGTM!
Conditional approval with one corner case comment here https://github.com/ai-dynamo/frontend-crates/pull/166#discussion_r3907330171

parsers/v2/src/unified/gemma4.rs
keivenchang: fixed and pushed in 2a6d2006 - Gemma now preserves a balanced recovery candidate across trailing whitespace at EOF, with native and unified coverage at chunk sizes 1, 4, and 16.

keivenchang:
Indrajit’s conditional approval is addressed in 2a6d2006 — the parser now preserves a balanced recovery candidate across trailing whitespace at EOF, with native and unified coverage at chunk sizes 1, 4, and 16.

PR #173, opened 2026-08-04
perf(conformance): resolve fixtures in memory, 2m36s render -> 10s — NO PARSER CHANGE

keivenchang:
## Overview

The conformance render took 2m35.8s and the pytest suite 5m13s. Both spent their time redoing the same work: the fixture resolver ran 20 times per render, each run re-parsing all 1732 source YAMLs and round-tripping every output file through YAML on every version layer. The render was 97% one core on a 32-core box.

| | before | after |
|---|---|---|
| `render_table_v2.sh` | 2m35.8s | 9.5-12.7s |
| `python3 -m pytest tests` | 313.9s | 40-44s |
| `check.sh ci` (this PR's CI gate) | ~2m50s | 16.0s |

No parallelism was used — this is redundant work removed, so that lever is still free if it is ever needed.

## Details

- **`yaml_fast.py` (new)** routes `yaml.safe_load` **and** `safe_dump` through libyaml. Four files had each grown their own copy of the CSafeLoader shim and none patched the dump side, so the pure-Python emitter was 72% of a resolver run.
- **`fixture_corpus.py` (new)** is the shared parent for `load` / `version_key` / the `<impl>-<version>` splitter — one copy each in the three resolvers — plus the new `load_corpus`.
- **`resolve_docs()`** keeps the fold's accumulator in memory instead of on disk, so each source file is parsed once and each output dumped at most once (was: load staged file, merge, dump, write — per file per version layer).
- **The 20 `subprocess.run` resolves are gone.** `load_all_cases(mode, docs=...)` takes the resolved docs directly, so a selection is no longer staged to a tempdir just to be parsed straight back.
- **`test_browser_smoke.py`** dropped its module-scoped render fixture, a duplicate of conftest's session-scoped one, so the suite renders once instead of twice.
- **`check.sh ci`** runs the whole `tests/` dir rather than a three-file list that silently excluded the resolver, render-invariant and colorize tests: 77 -> 95 tests in CI. Browser tests skip themselves without Selenium, which CI does not install.

## Verification

Byte-identical output, not just "it still renders":

- The rendered HTML matches the pre-change output once the render timestamp is normalised, and the parsed JSON page model compares equal.
- The three resolvers produce identical staged trees — `diff -r` clean over 2188 files across 5 batch + 4 stream selections plus reasoning.
- `check.sh ci` re-run in a fresh venv with only `pyyaml jinja2 pytest` (what CI installs): rc=0, 95 passed + 2 skipped.

Deep-copying each grafted value separately expanded the YAML anchors the fixtures use for repeated `expected` blocks; copying the overlay doc in one `deepcopy` call preserves the sharing via the memo. `test_resolve_corpus_reuse.py` pins that plus the no-aliasing-between-selections property, and both were mutation-tested — reintroducing either defect fails the suite.

## Where

`conformance/utils/src/{yaml_fast,fixture_corpus,resolve_fixtures,resolve_stream_fixtures,resolve_reasoning_fixtures,fixtures,generate_conformance_table}.py`, `conformance/utils/{check.sh,src/_common.sh}`, `conformance/utils/tests/{conftest,test_browser_smoke,test_resolve_corpus_reuse}.py`

## Notes for the reviewer

- `_raw_chunk_counts` used to swallow YAML parse errors with a blanket `except Exception: continue`. Parsing now happens up front in `load_corpus`, so a malformed fixture fails the render instead of silently dropping its chunk counts. Intended direction, but it is a behavior change.
- `test_browser_smoke.py::test_every_wired_element_stays_pinned[normal]` fails in a full-suite run and passes in isolation. It fails the same way on untouched `main`, so it is pre-existing and untouched here.

### DIFF
--- a/conformance/utils/check.sh
+++ b/conformance/utils/check.sh
@@ -66,10 +66,12 @@ run_ci() {  # the conformance-table CI gate; fail-fast, also runnable locally
   # The uploaded CI artifact name predates the v1/v2 split; keep it stable.
   \cp -f "$out" "$UTILS/CONFORMITY.html"
   run_coverage
-  python3 -m pytest \
-    "$UTILS/tests/test_model.py" \
-    "$UTILS/tests/test_stream_on_batch.py" \
-    "$UTILS/tests/test_family_coverage.py" -q
+  # The WHOLE tests/ dir, not a hand-maintained file list: the old list named three
+  # files, so the resolver-fold, render-invariant and colorize tests never ran in CI
+  # and a new test file was gated only by remembering to add it here. The browser
+  # tests skip themselves when Selenium isn't installed (CI installs pyyaml/jinja2/
+  # pytest only), so this stays a no-engine, no-browser gate.
+  python3 -m pytest "$UTILS/tests" -q
 }
 
 engine="${1:-}"; shift || true
--- a/conformance/utils/src/_common.sh
+++ b/conformance/utils/src/_common.sh
@@ -76,6 +76,9 @@ _build_stage_base() {
   \cp -f "$TOOLS/model.py" "$STAGE/tests/parity/model.py"
   # markers.py (comparison/facts semantics) + impls.py (identity) are the single
   # source the tables import instead of forking their own comparison copies.
+  # yaml_fast.py is the one place PyYAML gets routed through libyaml; the staged
+  # generator imports it by name, so it has to sit beside it in tests/parity/.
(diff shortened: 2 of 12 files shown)


rmccorm4 (approved):

rmccorm4:
Rendered table looks good: https://github.com/ai-dynamo/frontend-crates/pull/173#issuecomment-5182429796

Thanks Keiven

PR #174, opened 2026-08-04
feat(parsers): add Qwen3 unified request modes

keivenchang:
#### Overview:

**Second half of the split from #178, now based directly on `main`.** #178 landed the peer-trait alignment, vendor registry, and shared GUI/conformance tooling. This PR carries everything that depends on **Qwen3 request modes**, plus the corpus evidence that has to move atomically with it.

Qwen3 gains the two request modes the reasoning/tool split cannot honor: a prompt-prefilled channel (`UnifiedParserStartingState`) and guided-JSON tool output (`UnifiedToolOutputMode`). Other model families are unchanged.

#### Where this sits

```mermaid
flowchart TB
    A["#178 &nbsp;·&nbsp; MERGED FOUNDATION &nbsp;·&nbsp; ordered parser API, error-safe recovery, vendor registry"]
    B["#174 &nbsp;·&nbsp; THIS PR &nbsp;·&nbsp; Qwen3 native + guided-JSON request modes &nbsp;·&nbsp; corpus 99 → 240 cases"]
    C["#166 &nbsp;·&nbsp; NEXT FAMILY &nbsp;·&nbsp; Gemma 4 reuses the same contract and the same 240 cases"]
    A ==> B ==> C
    classDef landed fill:#eef1f4,stroke:#57606a,stroke-width:2px,color:#000
    classDef here fill:#fff4ce,stroke:#b06000,stroke-width:4px,color:#000
    classDef next fill:#f6f8fa,stroke:#8c959f,stroke-width:2px,color:#000
    class A landed
    class B here
    class C next
```

image

The request-mode contract, failure behavior, evidence, and exclusions are tracked in [DIS-2544](https://linear.app/nvidia/issue/DIS-2544/v2-add-qwen3-unified-request-modes). #178 owns the reusable surface; this PR consumes it for Qwen3.

Input: `<think>Look it up.</think><tool_call>…</tool_call><think>Now answer.</think>It's 18C.`

```
before:  reasoning = "Look it up.Now answer."   <- second thought hoisted ahead of the call it followed, fused with the first
         content   = "It's 18C."
after:   reasoning("Look it up.") | tool_call(get_weather) | reasoning("Now answer.") | text("It's 18C.")
```

Ordering is not a field the split can add; it is lost at the seam between the two parsers.

#### Details:

- **Request modes** — prefilled channels and guided JSON, resolved once through an owned `UnifiedParserInit`, with bounded prefix-header ownership, checkpoints and rollback, guided payload draining, and native-markup fallback that stays control-only across every stream split. Invalid guided payloads default to a typed caller-actionable error; the conformance harness explicitly selects text recovery.
- **Corpus** — the authored feed goes from main's 33 scenarios / 99 cases to 80 / 240. 45 of the 47 added scenarios are request-scoped (groups 30/31/40/41/50/51); the other two are generic group-4 cases that travel here because the generator and LFS corpus have one owner. Generator, taxonomy, docs, schema consumers, archives, and manifest pins move together.
- **Capture contract** — split families keep their incremental chunk deltas as raw streaming evidence, but their assembled result follows the serving contract: whole-input reasoning first, then tool parsing over the remaining text. Folding those raw rows invented an interleaving the split path cannot represent. Qwen3 uses the native unified parser for both views, so its ordered `reasoning → tool_call → reasoning → text` stream remains intact.
- **Doc and corpus drift fixed** — before this PR, `UNIFIED_CASES.md` said 63 scenarios / 189 cases, claimed Gemma 4 ran the native `UnifiedParser` when the registry had only Qwen3, and described the 0.1.22 archive as holding 162 entries. The current generator emits 80 scenarios / 240 cases; the corrected 0.1.22 and 0.1.23 archives each contain all 240 cases.
- **LFS hygiene** — seven unrelated archive changes dropped: four tarballs whose extracted payloads are *identical* to the merge base (only gzip bytes differed), one whose only change is `captured_with.dynamo_v2` `0.1.23→0.1.24`, and two new tool-calling shards. None is request-mode evidence.

#### Verification:

```
cargo test --workspace --all-targets --locked   1451 passed, 33 suites, 0 failed
golden unified feed, 240 cases                  sha256 7e032176…  generated from the authored corpus
cargo clippy --workspace --all-targets          0 warnings
```

`parsers/v2/src/unified/mod.rs` (request modes and guided state), then `parsers/v2/src/unified/qwen3.rs`, then the authored cases in `conformance/utils/src/gen_unified_golden.py`. Read the generated archives last, by count and hash rather than line by line.

/coderabbit profile chill

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring parser starting state, output mode, and named tools.
  * Added guided JSON tool-output handling, including malformed-input recovery and raw argument preservation.
  * Expanded unified conformance coverage with new reasoning, tool-calling, streaming, schema, and recovery scenarios.
  * Added live capture validation and richer rendered conformance results.

* **Bug Fixes**
  * Improved marker handling, buffering, reset behavior, and recovery across chunk boundaries.
  * Empty outputs are now supported for control-only turns when documented.

* **Documentation**
  * Updated parser integration, migration, fixture, taxonomy, and conformance authoring guidance.

### DIFF
--- a/conformance/fixtures-manifest.json
+++ b/conformance/fixtures-manifest.json
@@ -1,9 +1,9 @@
 {
-  "snapshot": "20260807_193005",
-  "created_pt": "2026-08-07 19:30:05 America/Los_Angeles",
+  "snapshot": "20260814_105421",
+  "created_pt": "2026-08-14 10:54:21 America/Los_Angeles",
   "crates": {
-    "dynamo-parsers": "7.1.0",
-    "dynamo-parsers-v2": "0.1.27"
+    "dynamo-parsers": "7.1.1",
+    "dynamo-parsers-v2": "0.2.0"
   },
   "peers": {
     "vllm": "0.26.0",
@@ -171,19 +171,39 @@
       "size": 4308
     },
     {
+      "path": "unified/dynamo_v2-0.1.22.tar.gz",
+      "sha256": "a7e7f1d161719e8f6caaa1f3fb07f169718308cc13f6a687916975c0ab144de3",
+      "size": 10008
+    },
+    {
       "path": "unified/dynamo_v2-0.1.23.tar.gz",
-      "sha256": "7d8ec2aa4fd17b47afb2107bb3b90ad8461031eac575e88424c811845ef651b3",
-      "size": 5130
+      "sha256": "69dd3be5a0f3a108e513c34a8b3669b405b521061872bfab5b1fb9885ac76880",
+      "size": 10164
+    },
+    {
+      "path": "unified/dynamo_v2-0.1.25.tar.gz",
+      "sha256": "c3113143c757e206d719538092213268cdfbae1f6df0954ac73a72b6da55a5a1",
+      "size": 9660
+    },
+    {
+      "path": "unified/dynamo_v2-0.1.27.tar.gz",
+      "sha256": "64c76851f11dae7c3aeda35f746f64355e535b6c81c77b60f70c214f367dfa8c",
+      "size": 9785
+    },
+    {
+      "path": "unified/dynamo_v2-0.2.0.tar.gz",
+      "sha256": "3e0f153218785fc1a4285d90f19a7a54cc69066462848ba5eda61e25a522772e",
+      "size": 10039
     },
     {
       "path": "unified/golden.tar.gz",
-      "sha256": "43db28380b26c57e01ca90d1d63cf6674c27d41eed0f4603ac4ddda05f6ec8a4",
-      "size": 3644
+      "sha256": "fef673ea851d87236f33e8a989dfe22f1d73d7a1f2c7710bb31583bba71c21ba",
+      "size": 7074
     },
     {
       "path": "unified/inputs.tar.gz",
-      "sha256": "70fda09d7e2f2b2add94b075d3c64207ea0bde7513d426940ff570e35422a1b1",
-      "size": 12937
+      "sha256": "b074f0fe37564cbfb659f958a90c9d4ae45b38a161f6037cd8026ec94d2cf47c",
+      "size": 34876
     },
     {
       "path": "unified/sglang_python-0.5.14.tar.gz",
@@ -191,14 +211,19 @@
       "size": 4363
     },
     {
+      "path": "unified/sglang_python-0.5.16.tar.gz",
+      "sha256": "68b20e272da148503f2626342fbe7977ac259a0a391c5fc4b902ce59c9584d35",
+      "size": 8409
+    },
+    {
       "path": "unified/vllm_python-0.25.1.tar.gz",
-      "sha256": "001076878fe7649be3ff0ae44352c45112b5e7cbba86399afa9d8cf3993540ad",
-      "size": 4742
+      "sha256": "ac45a91593eb85a694321e86216eceeeb6134108a0aed7e352231910c930aa2a",
+      "size": 9365
     },
     {
       "path": "unified/vllm_rust-0.25.1.tar.gz",
-      "sha256": "6c0a80957217c5246e6a7269dfe22e4c7f018ac42b625d30cb1f3502eaca8fdd",
-      "size": 5367
+      "sha256": "0ef327ee87ac8df71dd8bee069ba33561cd4fc06486acb99298e73f1cac8ff9a",
+      "size": 10472
     }
   ]
 }
--- a/conformance/tests/unified_parity.rs
+++ b/conformance/tests/unified_parity.rs
@@ -17,6 +17,8 @@
 
 mod common;
 
+use common::Init;
+
 use std::collections::BTreeMap;
 
 use dynamo_parsers_v2::{
@@ -36,6 +38,11 @@ struct GoldenFile {
 struct GoldenCase {
     input: String,
     golden: Vec<UnifiedEvent>,
+    /// Request-scoped parser configuration, declared by the case. Shared with
+    /// `unified_render` via `common::Init` so both harnesses configure a case
+    /// identically; see that type for why it is declared and not inferred.
+    #[serde(default)]
+    init: Init,
 }
 
 /// Tool schemas the corpus is written against (string params, so a value like
@@ -53,6 +60,7 @@ fn tools() -> Vec<Tool> {
         mk("f", "x"),
         mk("g", "y"),
         mk("run", "cmd"),
+        mk("sum_values", "values"),
         mk("log", "note"),
     ]
 }
@@ -78,9 +86,11 @@ fn has_unified_parser(family: &str) -> bool {
     create_unified_parser_for_family(family, &[]).is_ok()
 }
 
-fn events(family: &str, chunks: &[String]) -> Vec<UnifiedEvent> {
+fn events(family: &str, chunks: &[String], init: &Init) -> Vec<UnifiedEvent> {
     let mut parser = create_unified_parser_for_family(family, &tools())
         .unwrap_or_else(|e| panic!("create unified parser for `{family}`: {e}"));
+    init.apply(&mut parser, family);
+
     let mut deltas = Vec::new();
     for chunk in chunks {
         deltas.extend(parser.push(chunk).unwrap_or_else(|e| panic!("push: {e}")));
@@ -167,7 +177,7 @@ fn unified_parser_matches_the_golden_oracle() {
     for file in &covered {
         for (id, case) in &file.cases {
             checked += 1;
-            let got = events(&file.family, &chunk_markers(&case.input));
+            let got = events(&file.family, &chunk_markers(&case.input), &case.init);
             if got != case.golden {
                 failures.push(format!(
                     "{id}\n     input: {:?}\n    golden: {}\n   unified: {}",
@@ -197,9 +207,9 @@ fn unified_parser_is_chunk_invariant() {
         .filter(|f| has_unified_parser(&f.family))
     {
         for (id, case) in &file.cases {
-            let baseline = events(&file.family, std::slice::from_ref(&case.input));
+            let baseline = events(&file.family, std::slice::from_ref(&case.input), &case.init);
             for (label, chunks) in splittings(&case.input) {
-                let got = events(&file.family, &chunks);
+                let got = events(&file.family, &chunks, &case.init);
                 if got != baseline {
                     failures.push(format!(
                         "{id} [{label}, {} chunks]\n  whole: {}\n    got: {}",
@@ -227,8 +237,10 @@ fn unified_parser_has_stream_batch_parity() {
         .filter(|f| has_unified_parser(&f.family))
     {
         for (id, case) in &file.cases {
-            let streamed = events(&file.family, &chunk_markers(&case.input));
+            let streamed = events(&file.family, &chunk_markers(&case.input), &case.init);
             let mut parser = create_unified_parser_for_family(&file.family, &tools()).unwrap();
+            case.init.apply(&mut parser, id);
+
             let batch = parser
                 .parse_complete(&case.input)
                 .unwrap_or_else(|e| panic!("{id}: parse_complete: {e}"));
@@ -257,11 +269,14 @@ fn unified_parsers_are_isolated_per_stream() {
                 continue;
             };
             let (ca, cb) = (chunk_markers(&a.input), chunk_markers(&b.input));
-            let solo_a = events(&file.family, &ca);
-            let solo_b = events(&file.family, &cb);
+            let solo_a = events(&file.family, &ca, &a.init);
+            let solo_b = events(&file.family, &cb, &b.init);
 
             let mut pa = create_unified_parser_for_family(&file.family, &tools()).unwrap();
             let mut pb = create_unified_parser_for_family(&file.family, &tools()).unwrap();
+            a.init.apply(&mut pa, id_a);
+            b.init.apply(&mut pb, id_b);
+
             let (mut da, mut db) = (Vec::new(), Vec::new());
             for i in 0..ca.len().max(cb.len()) {
                 if let Some(c) = ca.get(i) {
@@ -271,8 +286,8 @@ fn unified_parsers_are_isolated_per_stream() {
                     db.extend(pb.push(c).unwrap());
                 }
             }
-            da.extend(pa.finish().unwrap());
-            db.extend(pb.finish().unwrap());
+            da.extend(pa.finish().unwrap().events);
+            db.extend(pb.finish().unwrap().events);
 
             assert_eq!(
                 assemble(&da),
--- a/conformance/tests/unified_render.rs
+++ b/conformance/tests/unified_render.rs
@@ -31,6 +31,8 @@ use serde_json::{Value, json};
 
 mod common;
 
+use common::Init;
+
 #[derive(Deserialize)]
 struct GoldenFile {
     family: String,
@@ -42,6 +44,10 @@ struct GoldenCase {
     description: String,
     #[serde(default)]
     policy: Vec<String>,
+    #[serde(default)]
+    init: Init,
+    #[serde(default)]
+    finish_reason: Option<String>,
     input: String,
     golden: Vec<Ev>,
     expect: BTreeMap<String, Expect>,
@@ -103,6 +109,8 @@ fn tools() -> Vec<Tool> {
         mk("f", "x"),
         mk("g", "y"),
         mk("run", "cmd"),
+        mk("log", "note"),
+        mk("sum_values", "values"),
     ]
 }
 
@@ -179,8 +187,10 @@ fn unified_delta_json(d: &dynamo_parsers_v2::UnifiedParserEvent) -> Value {
 ///
 /// Both paths are driven from the SAME chunking as `dynamo_chunks`, so the
 /// assembled row and the per-chunk rows in the popup describe one run.
-fn dynamo_events(family: &str, input: &str) -> Vec<Ev> {
+fn dynamo_events(family: &str, input: &str, init: &Init) -> Vec<Ev> {
     if let Ok(mut parser) = create_unified_parser_for_family(family, &tools()) {
+        init.apply(&mut parser, family);
+
         let mut deltas = Vec::new();
         for chunk in chunk_input(input) {
             deltas.extend(
@@ -292,9 +302,11 @@ fn tool_deltas(res: &dynamo_parsers_v2::ToolParseResult, out: &mut Vec<Value>) {
 /// Stream `input` through Dynamo's split pipeline CHUNK BY CHUNK, recording the
 /// real per-chunk emitted deltas (v1 reasoning streaming incremental -> v2 tool
 /// streaming push on the leftover content).
-fn dynamo_chunks(family: &str, input: &str) -> Vec<ChunkRow> {
+fn dynamo_chunks(family: &str, input: &str, init: &Init) -> Vec<ChunkRow> {
     // Unified families: ONE parser, so a chunk's deltas are simply what it emitted.
     if let Ok(mut parser) = create_unified_parser_for_family(family, &tools()) {
+        init.apply(&mut parser, family);
+
         let mut rows = Vec::new();
         for chunk in chunk_input(input) {
             let deltas = parser.push(&chunk).unwrap_or_default();
@@ -309,12 +321,10 @@ fn dynamo_chunks(family: &str, input: &str) -> Vec<ChunkRow> {
             .iter()
             .map(unified_delta_json)
             .collect();
-        if !tail.is_empty() {
-            rows.push(ChunkRow {
-                delta_text: "‹finish›".to_string(),
-                deltas: tail,
-            });
-        }
+        rows.push(ChunkRow {
+            delta_text: "‹finish›".to_string(),
+            deltas: tail,
+        });
         return rows;
     }
 
@@ -350,13 +360,22 @@ fn dynamo_chunks(family: &str, input: &str) -> Vec<ChunkRow> {
         tool_deltas(&tr, &mut tail);
     }
     tool_deltas(&tp.finish().unwrap_or_default(), &mut tail);
-    if !tail.is_empty() {
-        rows.push(ChunkRow {
-            delta_text: "‹finish›".to_string(),
-            deltas: tail,
-        });
-    }
+    rows.push(ChunkRow {
+        delta_text: "‹finish›".to_string(),
+        deltas: tail,
+    });
     rows
+}
+
+#[test]
+fn finish_is_part_of_the_stream_schedule_even_when_it_emits_nothing() {
+    let rows = dynamo_chunks("qwen3", "plain response", &Init::default());
+    let finish = rows.last().expect("finish row");
+    assert_eq!(finish.delta_text, "‹finish›");
+    assert!(
+        finish.deltas.is_empty(),
+        "fixture must exercise an empty finish"
+    );
 }
 
 /// Classify a Dynamo divergence from the golden.
@@ -496,6 +515,44 @@ fn cell(
 
 #[test]
 fn render_unified_conformance_html() {
+    // The vLLM column is LIVE, not an expectation. `capture_vllm_rust_unified.py`
+    // records the `vllm-parser` crate against this same corpus; reading it here is
+    // what makes the column evidence instead of a claim.
+    let vllm_live: BTreeMap<(String, String), Vec<Ev>> = {
+        let froot = common::ensure_fixtures().join("unified");
+        let mut m = BTreeMap::new();
+        // Shards key by TAXONOMY id (`UNIFIED.30.a`); the golden keys by SCENARIO
+        // (`UNIFIED.guided_json_named_tool.qwen3`). The inputs shard carries both, so
+        // it is the bridge — without it every cell reads NO-DATA while the capture
+        // sits right there, which is how this first went wrong.
+        let mut by_tax: BTreeMap<(String, String), String> = BTreeMap::new();
+        for entry in glob_yaml(&froot.join("inputs")) {
+            if let Ok(doc) = serde_yaml::from_str::<InputDoc>(
+                &std::fs::read_to_string(&entry).unwrap_or_default(),
+            ) {
+                for (cid, c) in doc.cases {
+                    by_tax.insert((doc.family.clone(), cid), c.scenario);
+                }
+            }
+        }
+        if let Some(d) = common::version_dirs_ascending(&froot, "vllm_rust-").pop() {
+            for entry in glob_yaml(&d) {
+                if let Ok(doc) = serde_yaml::from_str::<CaptureDoc>(
+                    &std::fs::read_to_string(&entry).unwrap_or_default(),
+                ) {
+                    for (cid, c) in doc.cases {
+                        if let Some(sc) = by_tax.get(&(doc.family.clone(), cid)) {
+                            m.insert(
+                                (doc.family.clone(), format!("UNIFIED.{sc}.{}", doc.family)),
+                                c.assembled,
+                            );
+                        }
+                    }
+                }
+            }
+        }
+        m
+    };
     let dir = common::ensure_unified_golden();
     let mut files: Vec<GoldenFile> = Vec::new();
     for entry in std::fs::read_dir(&dir).unwrap() {
@@ -528,13 +585,13 @@ fn render_unified_conformance_html() {
             total += 1;
 
             // Dynamo: live.
-            let got = dynamo_events(&file.family, &case.input);
+            let got = dynamo_events(&file.family, &case.input, &case.init);
             let dclass = classify(&file.family, &case.golden, &got);
             eprintln!(
                 "{id:44} dynamo={dclass:6} :: {}",
                 got.iter().map(Ev::render).collect::<Vec<_>>().join("  |  ")
             );
-            let chunk_feed: Vec<Value> = dynamo_chunks(&file.family, &case.input)
+            let chunk_feed: Vec<Value> = dynamo_chunks(&file.family, &case.input, &case.init)
                 .into_iter()
                 .map(|r| json!({"delta_text": r.delta_text, "dynamo": r.deltas}))
                 .collect();
@@ -550,6 +607,8 @@ fn render_unified_conformance_html() {
                 "scenario": scenario,
                 "description": case.description,
                 "policy": case.policy,
+                "init": case.init.applied(),
+                "finish_reason": case.finish_reason.clone().unwrap_or_else(|| "stop".to_string()),
                 "input": case.input,
                 "golden": case.golden,
                 "dynamo": got,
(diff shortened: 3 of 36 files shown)


parsers/v2/src/unified/mod.rs
```diff
@@ -86,12 +88,57 @@ pub enum UnifiedEvent {
...
+    /// contains one call object or an array of call objects.
+    GuidedJson { named_tool: Option<&'a str> },
+}
+
 /// A parser that owns reasoning + content + tool calls for one stream.
 ///
 /// Streaming-first, like [`crate::ToolParser`]: `push` per decoded delta,
 /// `finish` once at end of stream. One instance parses exactly one choice of
 /// one request, which is what gives per-stream isolation (`I4`) by construction.
 pub trait UnifiedParser: Send {
```
indrajit96, line 210: There are two initialization paths: `initialize_from_prompt()` can derive the starting channel but implicitly selects native output, while `initialize_with_output_mode()` supports guided JSON but requires the caller to resolve the channel separately. 
How should a guided request with a prompt-opened reasoning channel be initialized without duplicating prompt detection? 
Could we use one resolved `UnifiedParserInit { starting_state, tool_output_mode }`, calculated after prompt rendering and backend configuration?

parsers/v2/src/unified/mod.rs
```diff
@@ -219,34 +392,1159 @@ impl ToolParseResult {
...
+                }
+            }
+        }
+        output
+    }
+
+    /// Parse the accumulated payload. Anything that does not parse as the
+    /// expected call shape is surfaced as visible text rather than dropped
+    /// (policy P2 — best-effort recovery, never silent loss).
+    fn finish_json(&mut self) -> Result<Vec<UnifiedDelta>> {
```
indrajit96: In named/required guided mode, malformed JSON or the wrong call shape is converted into visible Text. 
A tracing warning is useful operationally, but the caller cannot enforce policy from it. 
Could `UnifiedParserOutput` carry structured diagnostics such as InvalidGuidedPayload, allowing the serving layer to fail or quarantine the request instead of exposing raw tool payload bytes as an assistant answer?

parsers/v2/src/unified/mod.rs
```diff
@@ -219,34 +392,1159 @@ impl ToolParseResult {
...
+            self.json.clear();
+            return Ok(Vec::new());
+        }
+
+        let raw_payload = self.json.clone();
+        let calls = match &self.named_tool {
+            // A named choice constrains output to that tool's ARGUMENTS alone,
+            // so the payload is the argument object and the name is known.
+            // Arguments are an OBJECT. A bare string / number / null / array is
+            // syntactically valid JSON but is not an argument set, and dispatching it
```
indrajit96: I verified that native, guided-required, and guided-named modes can all emit a ToolCall whose name is absent from the supplied tools. 
If this parser is intentionally syntax-only, can we document that contract clearly and ensure the serving integration validates the offered tool names, tool_choice, argument schema, and parallel-call policy before execution? 
The comments currently say “dispatch,” which sounds as though this validation already happened.

parsers/v2/src/unified/mod.rs
```diff
@@ -86,12 +88,57 @@ pub enum UnifiedEvent {
...
+        starting_state: UnifiedParserStartingState,
+        tool_output_mode: UnifiedToolOutputMode<'_>,
+    ) -> Result<()> {
+        if tool_output_mode != UnifiedToolOutputMode::Native {
+            anyhow::bail!("this unified parser does not support guided tool output");
+        }
+        self.initialize(starting_state)
+    }
+
     /// Feed one decoded text delta; returns the updates it completed, in order.
```
indrajit96: Qwen’s tool markers are special tokens, but push(&str) loses whether marker-shaped text came from a control token or ordinary text tokens. `preserve_special_tokens=true` prevents removal but does not preserve that identity. 
Since native mode can promote <tool_call>... inside text/reasoning into an executable call, should UnifiedParser accept token IDs or a structured decoder signal before becoming the default?

parsers/v2/src/tool_calling/scan.rs
```diff
@@ -293,20 +334,71 @@ impl<E: InvokeEmitter> WrappedBlockScanner<E> {
...
+    ///
+    /// NOT part of [`Self::control_markers`]: a BARE `</function>` with no invoke
+    /// open is ordinary text to the native scanner, and measured identical on both
+    /// paths, so stripping it unconditionally would create a divergence rather than
+    /// remove one. It is consumed only as the tail of an invoke already stripped.
+    pub(crate) fn invoke_end(&self) -> &str {
+        &self.spec.invoke_end
+    }
+
+    /// Select whether this stream interprets reasoning markers, without
```
indrajit96, line 432: For a family without ReasoningSpec, an explicit starting_state=Reasoning silently becomes reasoning-disabled, so leading private reasoning would be emitted as visible text. 
A no-op is reasonable for the neutral state, but should an explicitly requested reasoning state fail initialization when the parser cannot represent it?

indrajit96 (commented):
Overall, this PR delivers what it claims at the parser-library level for Qwen3. 
It introduces one parser that owns reasoning, visible text, and tool calls, and emits them as an ordered event stream. 
This fixes the fundamental split-parser ownership problem: a tool call can interrupt reasoning without being swallowed or leaked, while reasoning-like markers inside tool arguments remain argument data. 

I also compared this architecture against the parser/tool-leak fixes from the last two months. 
It structurally addresses the most important class: reasoning/tool ownership, event ordering, and many chunk-boundary marker leaks for correctly ported marker-based families. 
 
In my understanding What remains(Follow Up PRs) :
- Production serving does not consume UnifiedParser yet, so this is a parser implementation and registration—not a runtime rollout.
- Malformed assembled arguments currently become {}. That can make a corrupted side-effecting call indistinguishable from a legitimate no-argument call; I would prefer failing closed.
- Guided parse failures become visible text plus tracing warnings. A typed diagnostic/error result would let serving distinguish assistant text from invalid constrained output.
- Tool name, schema, tool_choice, and parallel-call validation remain outside the parser. That boundary should be explicit and enforced before dispatch.
- Native and guided modes share the ordered API but use separate internal state machines, so their parity is test-enforced rather than guaranteed by construction.
 - Correct prompt-prefilled state depends on caller plumbing. That needs boundary-level serving tests when integration happens.
- The current reasoning grammar is sufficient for Qwen3 and other simple paired-marker families, but not yet a universal replacement for Granite/Inkling/Harmony-style parsers.
 
 Thanks a lot @keivenchang for adding this critical peice!

parsers/v2/src/unified/mod.rs
keivenchang: Fixed in `eb78c72e`.

The behaviour is intentional — `parsers/v2/README.md` goal 7: parsing is separate from validation, emit the call and let serving reject unknown tools. But you are right that "dispatch" implied the validation had already happened; four comments now say "emit" (the parser emits, downstream dispatches).

The trait now carries an explicit **syntax-only** section: an emitted `ToolCall` guarantees nothing about tool availability, `tool_choice`, argument schema or parallel-call policy, and the serving layer MUST validate before executing. Agreed a serving-boundary test should land with the integration — there is no serving consumer of `UnifiedParser` in this repo yet, so this PR cannot prove that boundary.

parsers/v2/src/tool_calling/scan.rs
keivenchang: Fixed in `eb78c72e`.

`ScannerUnified::initialize_request` now rejects `starting_state == Reasoning` when the family has no `ReasoningSpec`, before anything is mutated. Deliberately NOT in `set_reasoning_mode`: its generic `enabled=true` cannot distinguish an explicit demand from the neutral default, which is exactly the conflation you identified — a no-op is right for `None` and wrong for `Reasoning`.

Worth recording that my first regression test for this was a false positive: it exercised built-in `qwen3`, which always installs a reasoning spec, so it asserted that `Reasoning` *succeeds* and could never have caught the defect. It is replaced by tests that build the real qwen3 scanner without `.with_reasoning(..)` and prove `Reasoning` rejects, `None`/`Response` still accept, and a rejected initialize leaves the parser reusable. Negative-controlled: revert the guard and 2 of 3 fail.

The same pass found a sibling you did not see — unsupported guided mode was mutating `starting_state`, resetting the scanner and setting the reasoning mode *before* discovering the scanner had no reasoning spec, so a rejected initialize left the parser half-configured. All prerequisites are now resolved before any mutation.

parsers/v2/src/unified/mod.rs
keivenchang: fixed in 23981e78 — one owned `UnifiedParserInit` now carries prompt IDs, starting state, output mode, and failure policy; `initialize(&[u32])` is the only peer-shaped adapter.

parsers/v2/src/unified/mod.rs
keivenchang: fixed in 23981e78 — invalid guided payloads now default to a typed `InvalidGuidedPayload` error; only the conformance harness explicitly selects byte-preserving `RecoverAsText`.

parsers/v2/src/unified/mod.rs
keivenchang: addressed in 23981e78 — this stays a syntax-only text parser matching the peer contract, and the trait now requires serving-side availability, authorization, schema, and tool-choice validation before execution.

keivenchang:
@indrajit96 — heads up that this PR was **split** after your review, so the code you commented on has moved and the history was rewritten.

- **#178** now carries the peer-trait alignment, the vendor registry, the GUI/hover fixes and the conformance tooling. It targets `main` and is independently green.
- **This PR** keeps the qwen3 request modes and their corpus, and is now stacked on #178.

Your five comments were against `514f102f`, which no longer exists on this branch. Where each one went:

| Your comment | Where it is now |
|---|---|
| unknown tool names / "dispatch" wording | fixed here — trait now states the syntax-only contract explicitly, and "dispatch" is reworded to "emit" |
| explicit `Reasoning` on a family that cannot represent it | fixed here — `initialize_request` rejects it before any mutation, with tests on a scanner built without `.with_reasoning(..)` |
| two initialization paths cannot be combined | still open, still here. Agreed design is one `UnifiedParserInit` owner with peer-shaped `initialize(&[u32])` as an adapter |
| malformed guided payload has no caller signal | still open, still here. Planned shape is an explicit policy in `UnifiedToolOutputMode::GuidedJson` rather than a new event variant |
| `push(&str)` loses special-token provenance | still open, still here, tracked as an integration gate before unified becomes the default |

The two fixed threads are resolved; the three open ones are genuinely open, not stale. I re-verified each against the rewritten branch rather than letting the force-push make them look outdated.

Two further things your review led to, which you did not see: an independent audit found that the vendor registry I added afterwards had an alias gap (registering `qwen3` did not shadow `qwen3_coder`, so the same family silently ran two different parsers depending on routing), and that my first regression test for your `Reasoning` finding was a false positive — it exercised built-in qwen3, which always has a reasoning channel, so it asserted the opposite of what it claimed. Both are fixed and negative-controlled.

keivenchang:
@coderabbitai review

@devin review

This is the qwen3 request-modes half of the split (stacked on #178). Note it is a Draft — if that suppresses review, say so and I will mark it ready.

Highest-value areas:

- `parsers/v2/src/unified/mod.rs` and `qwen3.rs` — the guided-JSON state machine and the prefilled-channel starting states. Malformed or wrong-shape guided payloads currently become visible text with no caller-actionable signal; that is a known open finding, not an oversight.
- The trait has four initialization entry points (`initialize`, `initialize_with_state`, `initialize_with_output_mode`, `initialize_from_prompt`). Consolidating them is a known open finding too.
- `parsers/v2/src/tool_calling/scan.rs` — bounded prefix-header ownership and checkpoint/rollback across chunk splits. Split-invariance is the property most likely to be wrong here.

Three of Indrajit's five findings are still open below and are tracked, not forgotten.

conformance/fixtures-manifest.json
```diff
@@ -1,9 +1,9 @@
 {
-  "snapshot": "20260729_121209",
-  "created_pt": "2026-07-29 12:12:09 America/Los_Angeles",
+  "snapshot": "20260804_141402",
+  "created_pt": "2026-08-04 14:14:02 America/Los_Angeles",
   "crates": {
-    "dynamo-parsers": "6.0.4",
-    "dynamo-parsers-v2": "0.1.23"
+    "dynamo-parsers": "7.0.1",
+    "dynamo-parsers-v2": "0.1.25"
```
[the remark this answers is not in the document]
keivenchang: fixed in 23981e78 — the regenerated manifest records parser v2 0.2.0 and preserves the historical 0.1.25 shard. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

conformance/tests/unified_parity.rs
```diff
@@ -35,6 +37,11 @@ struct GoldenFile {
 struct GoldenCase {
     input: String,
     golden: Vec<UnifiedEvent>,
+    /// Request-scoped parser configuration, declared by the case. Shared with
+    /// `unified_render` via `common::Init` so both harnesses configure a case
+    /// identically; see that type for why it is declared and not inferred.
+    #[serde(default)]
+    init: Init,
```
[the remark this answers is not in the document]
keivenchang: fixed in 23981e78 — `GoldenCase` models `init` and `finish_reason`, `Init` serializes, and unknown fields are denied. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

conformance/tests/unified_render.rs
```diff
@@ -308,12 +320,10 @@ fn dynamo_chunks(family: &str, input: &str) -> Vec<ChunkRow> {
...
-        if !tail.is_empty() {
-            rows.push(ChunkRow {
-                delta_text: "‹finish›".to_string(),
-                deltas: tail,
-            });
-        }
+        rows.push(ChunkRow {
+            delta_text: "‹finish›".to_string(),
+            deltas: tail,
+        });
```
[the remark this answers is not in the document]
keivenchang: fixed in 23981e78 — the current 0.2.0 capture was regenerated and the live-capture drift guard passes all 237 cases. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

conformance/tests/unified_schema_roundtrip.rs
```diff
@@ -120,10 +120,19 @@ fn every_case_is_well_formed() {
...
+            // every case against the live parser, so an empty golden passes only when
+            // the parser genuinely produces no events. Require the description to SAY
+            // so, else a reader cannot tell a deliberate no-op from a missing one.
+            if case.golden.is_empty() {
+                assert!(
+                    case.description.contains("emit"),
+                    "{path}: `{id}` has an empty golden but its description does not say \
+                     the turn emits nothing"
+                );
+            }
```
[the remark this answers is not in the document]
keivenchang: fixed in 23981e78 — empty goldens now require an explicit emits-nothing or emits-no-events declaration. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

conformance/utils/lib/parsers/UNIFIED_CASES.md
```diff
@@ -113,24 +113,161 @@ Case IDs use short `group.sub` labels (`1.a`, `2.b`, …) like the other suites;
...
+- **`30.f`** (`guided_json_after_reasoning`) A normal thought, THEN the constrained payload. Every other guided case starts at the payload, so nothing pinned the ordinary shape where the model reasons first and the backend constrains only the call. This is the baseline group 31's surroundings cases contrast with.
+- **`30.g`** (`guided_json_marker_inside_argument`) A control marker of the family's OWN grammar inside a guided argument VALUE. Once the payload has opened, a marker is argument DATA and must survive byte-exact (`I7`); re-reading it as a channel token corrupts the call the tool receives while still looking like a successful dispatch. The golden argument is the family's own marker, not a placeholder — a stand-in would pass whatever the parser did.
 
-### Group 31 — Guided decoding, malformed
+### Group 31 — Guided decoding, malformed / recovery
 - **`31.a`** (`guided_json_invalid_call`) Valid JSON that is not a call (no `name`). Surfaces as text (P2); no call dispatched.
```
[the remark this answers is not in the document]
keivenchang: fixed in 23981e78 — 31.a now names the guided malformed-payload policy, not P2. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

conformance/utils/src/e2e_cases.json
```diff
@@ -0,0 +1,655 @@
...
+      "mode": "stream",
+      "steps": 2,
+      "unified": [
+        "30.b"
+      ],
+      "artifacts": [
+        "case-0145-lifecycle_chained_calculation__stream-budget_capped.json",
+        "case-0145-lifecycle_chained_calculation__stream-budget_unlimited.json"
+      ]
+    },
```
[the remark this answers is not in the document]
keivenchang: fixed in 23981e78 — the chained lifecycle entry maps both 10.b and 30.b. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

conformance/utils/src/gen_unified_golden.py
```diff
@@ -377,27 +478,608 @@ def D(cls, note):
...
+      {"kind": "tool_call", "name": "get_weather", "arguments": {"city": "Paris"}}],
+     {"starting_state": "Reasoning", "tool_output_mode": "GuidedJson", "named_tool": None},
+     {"finish_reason": "stop"},
+     {
+        "qwen3": ("checking weather</think>[{\"name\": \"get_weather\", \"arguments\": {\"city\": \"Paris\"}}]",
+                  D("UNSUPPORTED", "vLLM base case doesn't set a starting channel state or use guided JSON"),
+                  {"verdict": "match", "note": "Dynamo v2 unified parser with starting_state=Reasoning and tool_output_mode=GuidedJson{named_tool=None}"}),
+        "gemma4": ("checking weather<channel|><|tool_call>call:get_weather{city:<|\"|>Paris<|\"|>}<tool_call|>", M, M),
+        "kimi_k2": ("checking weather</think><|tool_calls_section_begin|><|tool_call_begin|>functions.get_weather:0<|tool_call_argument_begin|>{\"city\": \"Paris\"}<|tool_call_end|><|tool_calls_section_end|>", M, M),
+     }),
```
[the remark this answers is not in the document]
keivenchang: fixed in 23981e78 — the prefilled guided case uses the shared guided-surroundings path, and the request-mode gate corrects the vLLM verdict too. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

conformance/utils/src/gen_unified_golden.py
```diff
@@ -377,27 +478,608 @@ def D(cls, note):
...
+# trailing marker is handed back byte-identical — trailing space and all. The
+# corpus records that difference instead of each case guessing at it.
+GUIDED_SURROUNDS = {
+    "clean": (lambda pay, fam: pay, "no surrounding grammar", False),
+    "trailing_close": (lambda pay, fam: f"{pay}{control_tokens(fam)[3]}",
+                       "a stray tool CLOSE after the payload", True),
+    "wrapped": (lambda pay, fam: f"{control_tokens(fam)[2]}{pay}{control_tokens(fam)[3]}",
+                "the payload wrapped in native tool markup", True),
+    "bare_opener": (lambda pay, fam: f"{control_tokens(fam)[2]}{pay}",
+                    "a bare tool OPENER before the payload, never terminated", False),
```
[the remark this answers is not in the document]
keivenchang: fixed in 23981e78 — `invoke_header_prefix` now supplies the unterminated inner header for all three marker-ownership scenarios. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

conformance/utils/src/gen_unified_golden.py
```diff
@@ -409,24 +1091,59 @@ def build_cases(fam):
...
+        # Nothing asserts this field (the Dynamo column is computed live), so a
+        # false `match` never fails; it just tells a reader two engines handle
+        # request modes they cannot see. One gate every case passes through is the
+        # only way an authoring shortcut cannot route around it.
+        if fam not in UNIFIED_FAMILIES and _init_is_request_scoped(init):
+            dynamo = D(
+                "UNSUPPORTED",
+                "no native unified parser in this build, so the split path ignores "
+                "`init` and cannot honour this request mode",
+            )
```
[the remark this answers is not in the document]
keivenchang: fixed in 23981e78 — every request-scoped case now receives the matching unsupported vLLM verdict. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

conformance/utils/src/generate_conformance_table.py
```diff
@@ -2232,7 +2232,7 @@ def _read_dir(name):
             "id": cid, "scenario": scenario, "family": fam,
             "description": inp.get("description", ""),
             "policy": inp.get("policy") or [], "policy_tags": inp.get("policy") or [],
-            "init": inp.get("init") or {"prefill": "None", "tool_output_mode": "Native", "named_tool": None},
+            "init": inp.get("init") or {"starting_state": "None", "tool_output_mode": "Native", "named_tool": None},
```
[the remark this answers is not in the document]
keivenchang: fixed in 23981e78 — table generation normalizes each omitted init field to the Rust defaults. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

parsers/v2/src/tool_calling/scan.rs
```diff
@@ -289,20 +330,71 @@ impl<E: InvokeEmitter> WrappedBlockScanner<E> {
...
+        // invites. The check could not tell "this family has no reasoning" from
+        // "this scanner was built wrong" anyway; it only ever caught the former.
+        let enabled = enabled && self.reasoning.is_some();
+        self.reasoning_enabled = enabled;
+        if let Some(reasoning) = self.reasoning.as_mut() {
+            reasoning.forced_start = forced_start;
+        }
+        self.in_reasoning = enabled && forced_start;
+        self.accept_redundant_reasoning_start = self.in_reasoning;
+    }
```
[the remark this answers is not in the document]
keivenchang: fixed in 23981e78 — request-scoped forced start is stored separately and falls back to the immutable family declaration. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

parsers/v2/src/tool_calling/scan.rs
```diff
@@ -432,10 +610,9 @@ impl<E: InvokeEmitter> WrappedBlockScanner<E> {
         let mut out: Vec<UnifiedParserEvent> = Vec::new();
 
         loop {
-            // Reasoning yields to tool structure: while a thought is open, its
-            // closer OR a nested tool opener ends the current reasoning run (see
-            // `drain_reasoning`), so a call emitted inside a thought is extracted
-            // and the thought resumes after it.
+            // Reasoning is the innermost scope: while it is open, ONLY its
+            // closer ends it, so a `<tool_call>` inside a thought stays part of
+            // the thought instead of being executed.
```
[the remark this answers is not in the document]
keivenchang: fixed in 23981e78 — the comment now matches the extraction/resume behavior and documents the reverse nesting rule. @coderabbitai does this address the issue, and is this conversation resolvable? if so, please resolve conversation.

parsers/v2/src/unified/mod.rs
```diff
@@ -464,45 +641,1096 @@ impl ToolParseResult {
...
+                        &self.input,
+                        &self.control_markers,
+                        &self.invoke_end,
+                        // Nor past the start of the payload itself.
+                        self.input.find(['{', '[']),
+                        // A thought marker ahead also ends the header: a stray
+                        // `<function=` must not borrow the `>` from `<think>` and
+                        // swallow the thought — that put private reasoning in the
+                        // user's answer.
+                        &[start, end],
```
indrajit96, line 1351: This call only supplies the reasoning markers as competitors. A later tool marker such as <tool_call> is therefore not allowed to end a held <function=...> header.
Now consider
For x<function=f><tool_call>[{"name":"f","arguments":{"x":"y"}}], one push emits Text("x"), ToolCall(f), while character-sized pushes emit Text("x<function=f>"), ToolCall(f).
can we include the other tool-control markers in the prefix-header boundary and holdback calculations. The marker at the current position should be excluded or ignored so it does not compete with itself?

Suggested Fix

Build the competitor list from both reasoning and tool-control markers:
```
  let competing_markers = self
      .control_markers
      .iter()
      .map(String::as_str)
      .chain([start, end])
      .collect::<Vec<_>>();
```
  Pass that list to control_marker_at() instead of only:
```
  &[start, end]
```
  Apply the same rule inside guided_holdback_len(), where pending prefix-form markers are evaluated. prefix_header_end() must ignore the marker at offset zero, otherwise the prefix competes with itself:
```
  if let Some(rel) = haystack[at..bound].find(marker)
      && rel > 0
  {
      bound = at + rel;
  }
```
 A cleaner production implementation could centralize this competitor calculation so the scanner and holdback logic cannot drift apart.

 Add a regression test using the exact reproducer. Compare one-push parsing with character-sized parsing and assert that neither output contains <function=f>.

I tried this fix locally: both new regressions and all 216 dynamo-parsers-v2 tests passed.

parsers/v2/src/unified/mod.rs
```diff
@@ -464,45 +641,1096 @@ impl ToolParseResult {
...
+                            &self.input,
+                            reasoning_markers,
+                            &self.control_markers,
+                            &self.invoke_end,
+                            flush,
+                        )
+                    };
+                    let visible_len = self.input.len().saturating_sub(keep);
+                    if visible_len > 0 {
+                        if self.payload_emitted && self.input[..visible_len].trim().is_empty() {
```
indrajit96: Required guided mode loses whitespace when it arrives in its own chunk. 
For [{"name":"f","arguments":{"x":"y"}}]x , one push emits Text("x "), while character-sized pushes emit only Text("x").
After x has been emitted as post-payload text, the following space is visible content. 
This branch buffers the whitespace and then discards it during finish() because named_tool is None. Please track whether post-payload text has started. Before text starts, whitespace may still be treated as a JSON separator. After text starts, whitespace must be emitted as Text.

Suggested fix :

  Can we add a request-scoped state:
```
  payload_emitted: bool,
  post_payload_text_started: bool,
```
  Then change the condition around line 1409:
```
  if self.payload_emitted
      && !self.post_payload_text_started
      && self.input[..visible_len].trim().is_empty()
  {
      // Existing JSON-separator handling.
  } else if self.payload_emitted {
      push_run(&mut output, Kind::Text, &self.input[..visible_len]);
      self.post_payload_text_started = true;
  }
```
 Initialize and reset the new field to false. 
Also set it to true in every other branch that emits Kind::Text after the payload.

Add a regression test beside `guided_json_is_chunk_boundary_independent` that parses the reproducer once as a single push and once character-by-character, then asserts both produce ToolCall(f), Text("x ").

parsers/v2/src/unified/mod.rs
indrajit96: Verified on current head. 
`UnifiedParserInit` now owns prompt token IDs, starting state, output mode, and invalid-payload policy, and request configuration goes through `initialize_request`. This addresses the split initialization paths I was concerned about. Thanks!

parsers/v2/src/unified/mod.rs
indrajit96: Verified on current head. 
Invalid guided output now defaults to a typed `InvalidGuidedPayload` error, while only callers that explicitly select `RecoverAsText` preserve it as visible text. 
This gives serving an enforceable failure policy. Thanks!

parsers/v2/src/unified/mod.rs
indrajit96: Verified. The trait now explicitly documents that ToolCall emission is syntax-only and requires downstream availability, authorization, schema, tool_choice, and parallel-call validation. Since this repository has no serving consumer yet, I agree the boundary test belongs with that integration. Thanks!

parsers/v2/src/unified/mod.rs
```diff
@@ -464,45 +641,1096 @@ impl ToolParseResult {
...
+/// unbounded, then it was bounded only by the payload start — which let a stray
+/// `<function=` BORROW the `>` from a later `<think>`, so
+/// `<function=<think>secret</think>[…]` consumed through the thought opener and
+/// emitted the model's private reasoning as visible text (`I3`).
+///
+/// The boundary is the earliest of: the payload start, and any competing control or
+/// reasoning marker. A header that has no `>` before that point is not a complete
+/// header — it is literal text the model happened to write.
+fn prefix_header_end(haystack: &str, at: usize, limit: Option<usize>, competing: &[&str]) -> usize {
+    let mut bound = limit.unwrap_or(haystack.len()).max(at);
```
indrajit96: A complete `<function=...>` marker is handled differently depending on chunking.
In guided RecoverAsText mode, `limit` is the position of the first `{` or `[`. That delimiter can occur before the `<function=` marker.

For example:

`x{}<function=f>`

Here, `limit` points to `{`, while `at` points to `<function=`. This line:
```
  let mut bound = limit.unwrap_or(haystack.len()).max(at);
```
  sets `bound` to exactly `at`. The parser then searches an empty range and cannot see the closing `>`.

This produces different output for the same input:

  - One push: Text("x{}f>")
  - Character chunks: Text("x{}")

The one-push path removes only `<function=` and leaks `f>`. The character-chunk path waits for the complete marker and removes all of `<function=f>`.

A limit before the current marker should not restrict this marker's header search. 

Suggested fix:
```
  let mut bound = limit
      .filter(|&end| end > at)
      .unwrap_or(haystack.len());

  for marker in competing {
      if let Some(rel) = haystack[at..bound].find(marker)
          && rel > 0
      {
          bound = at + rel;
      }
  }
```

indrajit96 (commented):
The overall design looks good, but I found three chunk-boundary correctness issues in guided mode. 
In each case, identical input can produce different visible output depending on how the stream is chunked. 
Happy to re-review once updated.

parsers/v2/src/unified/mod.rs
keivenchang: confirmed — `prefix_header_end` is called with only `&[start, end]`, so no tool-control marker can end a held header. i also checked the competitor loop at mod.rs:915 and it has no `rel > 0` guard, so your second point holds too: the marker competes with itself. taking your approach of building the competitor list from control markers chained with start/end, and centralizing it so the scanner and holdback cannot drift. thanks for the exact reproducer and for running it locally — that saved a round trip.

parsers/v2/src/unified/mod.rs
keivenchang: agreed. buffering the whitespace and then dropping it in finish() because named_tool is None is the bug — after post-payload text has started, a space is visible content, not a JSON separator. tracking payload_emitted / post_payload_text_started as you suggest.

parsers/v2/src/unified/mod.rs
keivenchang: confirmed this one line by line: `let mut bound = limit.unwrap_or(haystack.len()).max(at);` — when limit (first `{` or `[`) sits before `at`, the `.max(at)` makes bound == at and `haystack[at..at]` is empty, so the closing `>` is unreachable and one-push leaks `f>`. your `filter(|&end| end > at)` is the right shape.

worth noting all three of these are the same class, and we already have 13 split-invariance tests in qwen3.rs including two that iterate every cut — they missed these because the tested INPUT set does not contain these shapes. so rather than adding three fixed cases, i am going to generate inputs over the marker alphabet and assert one-push == every-split across them, otherwise the next three shapes slip through the same way. will cite the SHA on all three threads when it lands.

parsers/v2/src/unified/mod.rs
keivenchang: fixed in 3a88dac1 — competitor list now built from `control_markers()` chained with start/end and passed through to `control_marker_at()` and the holdback path, so a tool marker can end a held header. the self-competition guard is in too (`find(|&rel| rel > 0)`).

parsers/v2/src/unified/mod.rs
keivenchang: fixed in 3a88dac1 — added `post_payload_text_started`, so whitespace is a JSON separator only before visible text starts and is emitted as Text after. regression asserts one-push and char-sized both give ToolCall(f), Text("x ").

parsers/v2/src/unified/mod.rs
keivenchang: fixed in 3a88dac1 — `limit.filter(|&end| end > at)` so a payload boundary before the marker no longer collapses the header search.

on the class point: rather than three fixed regressions, there is now `generated_control_marker_pairs_are_split_invariant`, which builds inputs from the marker alphabet (`<function=`, `<tool_call>`, `<think>`, `</think>`, `{`, `[`, `>`), takes all 49 ordered pairs, and asserts one-push output equals every char-boundary split for each — ~2k assertions. your three reproducers are in there as specific pairs, and so are the ones none of us thought of. thanks, this was a good catch and the generated version came directly out of it.

leaving all three threads open for you to close.

indrajit96: Verified on 3a88dac. 
I reran the exact `x{}<function=f>` reproducer as one push, every two-chunk split, and character-by-character delivery. 
All paths now produce `Text("x{}")` without leaking `f>`.

indrajit96 (approved):
LGTM!
Thanks for adding this @keivenchang

rmccorm4:
/ok to test [3a88dac](https://github.com/ai-dynamo/frontend-crates/pull/174/commits/3a88dac10c5cb8bd62264d1cd2e982503cec3b7a)

PR #178, opened 2026-08-08
refactor(parsers): align the unified parser surface with peer traits, and let vendors supply their own

keivenchang:
#### Overview:

This PR aligns the unified parser's ordered-event and incremental-parse surface with the peer traits, and lets a third party supply their own unified parser — including replacing one this crate ships — without forking.

Two peer capabilities are deliberately NOT in this PR, so "aligned" should not be read as "drop-in adoptable": the peer's `create` takes a tokenizer as well as the tools, while the vendor factory here takes only tools, so a peer parser that resolves its markers from a tokenizer cannot be constructed through this registry yet; and `structural_tag_model`, which constrains generation for strict tool calling, has no type or consumer in this crate. Both need a design, and neither is closed by the follow-up request-mode PR — that one selects between native markup and guided JSON in already-generated bytes, which is a different question. It also fixes hover popups in the conformance report, which had stopped opening entirely.

Split out of [DIS-2544](https://linear.app/nvidia/issue/DIS-2544) / #174, which bundled several unrelated concerns into ~7k lines across 56 files and stalled in review. This half does **not** depend on qwen3 request modes, so it lands on its own; the request modes follow stacked on top in #174.

**No corpus change.** The unified golden feed is byte-identical to `main`.

#### Where this sits

```mermaid
flowchart TB
    M["main &nbsp;·&nbsp; 99 golden cases"]
    A["#178 &nbsp;·&nbsp; THE SURFACE &nbsp;·&nbsp; UnifiedParser trait + output type + vendor registry"]
    B["#174 &nbsp;·&nbsp; REQUEST MODES &nbsp;·&nbsp; StartingState + ToolOutputMode &nbsp;·&nbsp; corpus 99 → 237"]
    C["#166 &nbsp;·&nbsp; SECOND FAMILY &nbsp;·&nbsp; gemma4 on the surface, graded by #174's corpus"]
    M --> A --> B --> C
    classDef here fill:#fff4ce,stroke:#b06000,stroke-width:3px,color:#000
    classDef other fill:#f6f8fa,stroke:#8c959f,color:#000
    classDef base fill:#eef1f4,stroke:#8c959f,color:#000
    class M base
    class A here
    class B,C other
```

The full graph — every trait, type and registry entry each layer adds, and which layer consumes it — is in [DIS-2644](https://linear.app/nvidia/issue/DIS-2644/v2-align-the-unified-parser-surface-with-peer-traits-and-let-vendors), kept there because it is too wide to read inline on GitHub.

#### Review aid: the lifecycle this PR pins down

```mermaid
stateDiagram-v2
    direction LR
    [*] --> Fresh
    Fresh --> Streaming : parse_into
    Streaming --> Streaming : parse_into
    Streaming --> Errored : Err
    Errored --> Fresh : reset()
    Streaming --> Ended : finish()
    Ended --> [*]
```

What each edge guarantees, and where to check it in the diff:

| edge | guarantee | where |
|---|---|---|
| `parse_into` | committed events land in the **caller's** output, so a later `Err` in the same advance cannot retract them | `ScannerUnified::parse_into` writes straight through `EventSink` |
| `Err` | the failing invoke was **never consumed** — buffer intact, `next_index` not advanced | both invoke sites call the emitter *before* `buffer.drain(..)` |
| `reset()` | returns the unconsumed text and clears **every** field carrying stream position: `in_block`, `in_reasoning`, `resume_reasoning`, `suppress_normal_text`, `next_index` | `reset_stream` in `scan.rs` |
| `finish()` | flushes into a fresh output; open reasoning is promoted rather than leaked | `ScannerUnified::finish` |

`push()` is the one spelling that CANNOT honour the first row — it owns its buffer and returns `Result<Vec<_>>`, which has nowhere to carry partial output. A parser that may commit and then fail must be driven through `parse_into`.

#### Review aid: what goes in, what comes out

Every line below is marked with where it comes from. Verified against the vLLM Rust parser crate at `rust/src/parser/src/unified/mod.rs` (v0.26.0).

```rust
// ---- IN ----
// (FROM vLLM — identical signature)
fn parse_into(&mut self, delta: &str, output: &mut UnifiedParserOutput) -> Result<()>;
// (FROM vLLM — identical signature)
fn initialize(&mut self, prompt_token_ids: &[u32]) -> Result<()>;

// ---- OUT, streaming ----
// (FROM vLLM — identical struct, identical public field)
pub struct UnifiedParserOutput { pub events: Vec<UnifiedParserEvent> }

// (FROM vLLM — identical variants, identical tuple shapes)
pub enum UnifiedParserEvent {
    Text(String),
    Reasoning(String),
    ToolCall(ToolCallDelta),
}

// (FROM vLLM — identical fields)
pub struct ToolCallDelta { pub tool_index: usize, pub name: Option<String>, pub arguments: String }

// ---- OUT, assembled ----
// (NOT vLLM — ours. vLLM has no assembled-event type and no assemble().)
pub enum UnifiedEvent {
    Reasoning { text: String },
    Text      { text: String },
    ToolCall  { name: String, arguments: serde_json::Value },
}
```

**Provenance of every trait method:**

| method | provenance |
|---|---|
| `parse_into` · `initialize` · `reset` · `preserve_special_tokens` · `tool_call_id` | **FROM vLLM** — identical signatures |
| `finish` | **FROM vLLM**, one deliberate difference: vLLM gives it a default, we make it REQUIRED. A family that forgets to flush silently drops the tail of every stream, which is not a failure worth inheriting for symmetry |
| `push(chunk) -> Result<Vec<UnifiedParserEvent>>` | **NOT vLLM** — ours, additive convenience. The conformance corpus asserts against this spelling |
| `parse_complete(output) -> Result<Vec<UnifiedEvent>>` | **NOT vLLM** — ours. vLLM has no batch entry point; routing batch through `parse_into`/`finish` makes stream/batch parity structural |
| `register_unified_parser` / `unregister` / aliases | **NOT vLLM** — ours. vLLM has no vendor registry for unified parsers |

**Accumulation helpers** `push_text` · `push_reasoning` · `push_call` · `append` are **FROM vLLM** by name and semantics, including that `append` coalesces across the seam. One difference: vLLM's `append(other: Self)` consumes; ours takes `&mut Self`.

**In vLLM, deliberately NOT here:** `create(tools, tokenizer)` — their construction takes a tokenizer, ours takes only `&[Tool]`, so a peer parser that resolves markers from a tokenizer cannot be built through this registry yet. `structural_tag_model()` — constrains generation for strict tool calling; no type or consumer in this crate. `append_tool_output()` — bridges their separate tool-only output type.

**Why a flat `Vec` and not `{reasoning_text, content, tool_calls}`.** A bundle of parallel fields physically cannot say whether text came before or after a call — that is why the split path today hoists every thought to the front and merges them. Ordering is not a field you can add; it has to be the shape. This is the part vLLM already got right, and the reason aligning to it was worth doing.

Two distinctions that are easy to misread in the diff:

- `ToolCallDelta.arguments` is a **`String` fragment** that concatenates across deltas; `UnifiedEvent::ToolCall.arguments` is a **parsed `serde_json::Value`**. `assemble()` is the single fold between them — and both the fold and the assembled type are ours, not vLLM's.
- `UnifiedParserEvent` uses **tuple variants** (vLLM's shape); `UnifiedEvent` uses **struct variants with a serde tag**, because it serializes to the golden-corpus schema.

`Result` is `anyhow::Result` throughout.

#### Review aid: what an emitter failure costs, after this PR

The ordering below is the fix. Before it, the buffer was drained *before* the fallible call and the drain loop collected into a local vector that `?` dropped — so this same input lost `prefix`, the `ok` call that had already succeeded, and both invoke bodies, leaving `reset()` with only `</tool_call>suffix`.

```mermaid
sequenceDiagram
    autonumber
    participant C as Caller (owns output)
    participant S as Scanner drain loop
    participant E as Emitter (fallible)

    C->>S: parse_into("prefix<tool_call>ok…boom…</tool_call>suffix")
    S-->>C: push_text("prefix") — COMMITTED
    S->>E: parse_invoke(ok)
    E-->>S: Ok(ToolCall)
    S-->>C: push_call(ok) — COMMITTED
    S->>S: drain the ok bytes (only after success)
    S->>E: parse_invoke(boom)
    E-->>S: Err
    Note over S: boom bytes were NOT drained
    S-->>C: Err
    C->>S: reset()
    S-->>C: "<function=boom></function></tool_call>suffix"
```

Hovering a cell in the conformance report opened nothing on a normal desktop:

```
before:  (hover: hover) reports false -> pointerenter listener never registered
         browser delivers real hover, cell matches CSS :hover, popup never opens
         keyboard focus also dead, because focusin sat behind the same gate
after:   listeners always registered; only tap-to-pin consults PointerEvent.pointerType
         real mouse, media queries false -> popup with 1334 chars, opacity 1, on top
```

#### Details:

- **Peer alignment** — `UnifiedParserEvent` in peer variant order and payload shapes, `parse_into` as the required method, `finish -> UnifiedParserOutput`, `initialize(&[u32])`, plus `reset` / `preserve_special_tokens` / `tool_call_id` and the peer accumulation helpers. `parse_complete` and the assembled `UnifiedEvent` view stay additive, so a peer-shaped caller never sees them.
- **Vendor registry** — `register_unified_parser` / `unregister_unified_parser`, keyed on the **canonical** family name so registering `qwen3` also shadows its `qwen3_coder` alias. Shadowing is non-destructive; unregistering restores the built-in exactly. `builtin_unified_families()` excludes vendor entries, because the conformance suite iterates it and a vendor family has no cases here. `CUSTOM_PARSERS.md` documents the contract, and its worked example exists as a compiled integration test (`tests/vendor_parser_example.rs`) using only the public API, with a guard asserting the guide's `impl UnifiedParser` methods match that file's. The guard is scoped to those method bodies — it does not compile the Markdown block, so a doc-only change to imports or struct fields would still slip past it.
- **GUI + tooling** — hover listeners no longer gated on a media query, capture-result semantics, ORDER/MERGE sequence display, render fixes.

#### Verification:

```
cargo test --workspace --all-targets --locked   1386 passed, 33 suites, 0 failed
python3 -m pytest conformance/utils/tests/      125 passed
cargo clippy --workspace --all-targets          0 warnings
piece-1 coupling gate                           PASS
golden.tar.gz / inputs.tar.gz                   43db2838… / 70fda09d… unchanged
```

`conformance/utils/piece1_coupling_gate.sh` is the falsifiable proof that no request-mode work leaked in: it rejects every request-mode symbol from the diff, proves `conformance/fixtures` and the manifest are byte-identical to `main`, pins both LFS pointers, and confirms the generator still emits main's 33 scenarios / 99 cases. It was negative-controlled against the combined branch, where it fails for eight independent reasons.

`parsers/v2/src/unified/mod.rs` (the trait and registry), then `parsers/v2/CUSTOM_PARSERS.md`.

/coderabbit profile chill

## Summary by CodeRabbit

* **New Features**
  * Added support for custom unified parsers, including registration, overrides, lifecycle controls, and vendor family management.
  * Improved conformance views with clearer event-order comparisons, compact tables, formatted tooltips, and sequence-divergence indicators.
  * Enhanced tooltip interactions across mouse, keyboard, touch, and pen input.

* **Bug Fixes**
  * Updated parser output handling for consistent reasoning, text, and tool-call events.

* **Documentation**
  * Added guidance for implementing, registering, testing, and integrating custom unified parsers.

* **Tests**
  * Expanded browser, parser, registry, and layout coverage for the updated behavior.

### DIFF
--- a/conformance/utils/src/assets/conformance.js
+++ b/conformance/utils/src/assets/conformance.js
@@ -697,22 +697,36 @@
   // still read `touch`, call `preventDefault()` and pin — suppressing the link the
   // keyboard user asked for. `pointercancel` clears it too, so an aborted gesture
   // cannot arm a later click.
+  // Both the KIND of pointer and WHERE it went down. The kind alone is a document-wide
+  // global: a tap that lands anywhere else — a header, a scroll, any element that is not a
+  // tooltip host — leaves `touch` armed, because only a tooltip host's click consumes it.
+  // The next MOUSE click on a host would then read that stale `touch`, pin, and
+  // `preventDefault()` the parser-source link the mouse user actually asked for. Requiring
+  // the pointerdown to have landed inside the same host makes the provenance belong to
+  // this interaction rather than to whatever touched the page last.
   let pendingPointerType = '';
+  let pendingPointerTarget = null;
   document.addEventListener('pointerdown', function (e) {
     pendingPointerType = e.pointerType || '';
+    pendingPointerTarget = e.target;
   }, true);
   document.addEventListener('pointercancel', function () {
     pendingPointerType = '';
+    pendingPointerTarget = null;
   }, true);
   // A click that came from a finger or stylus has no hover behind it, so it must open the
   // popup and pin it. A mouse click must not: hover already showed the popup, and pinning
   // is modal (see `hoverAllowed`), so a mouse pin would switch hover off page-wide.
   // Unknown/empty pointerType (synthetic `.click()`, keyboard activation) is treated as
   // NOT touch, matching the mouse/keyboard contract.
-  const consumeClickShouldPin = function () {
+  const consumeClickShouldPin = function (cell) {
     const t = pendingPointerType;
+    const target = pendingPointerTarget;
     pendingPointerType = '';
-    return t === 'touch' || t === 'pen';
+    pendingPointerTarget = null;
+    if (t !== 'touch' && t !== 'pen') { return false; }
+    // The gesture must have STARTED in this host, or it belongs to something else.
+    return !!(cell && target && cell.contains(target));
   };
   let pinnedCell = null;
   function unpinCell(c) { if (c && c._ttipUnpin) { c._ttipUnpin(); } }
@@ -839,7 +853,7 @@
       // so leaving the provenance unconsumed on any of them hands a stale `touch` to the
       // NEXT click — the defect consume-on-use exists to prevent, one branch over. A tap
       // on the popup's own links or its ✕ takes this return, and used to leave it armed.
-      const shouldPin = consumeClickShouldPin();
+      const shouldPin = consumeClickShouldPin(cell);
       if (e.target.closest('.ttip')) { return; }
       // Decided per-interaction, not per-page: only a finger or stylus pins. A mouse click
       // falls through to the cell's own parser-source link, and never enters modal pin mode.
--- a/conformance/utils/src/assets/conformance.js
+++ b/conformance/utils/src/assets/conformance.js
@@ -674,10 +674,46 @@
   }
 
   // Touch devices have no hover, so the tooltip is opened by TAP and pinned open
-  // (with an ✕ to close) rather than shown on pointerenter. Desktop keeps hover;
-  // a click there also pins (handy for reading a big tooltip without holding the
-  // mouse still). Only one tooltip is pinned at a time.
-  const hoverCapable = window.matchMedia('(hover: hover)').matches;
+  // (with an ✕ to close) rather than shown on pointerenter. Where hover EXISTS, hover is
+  // the only affordance — clicking never pins, because pinning is modal and one stray
+  // click otherwise disabled hover page-wide. So pinning is a touch-only mechanism now.
+  // Only one tooltip is pinned at a time.
+  // NEVER gate event REGISTRATION on a media query. `(hover: hover)` and
+  // `(any-hover: hover)` are advisory descriptions of a device; they are not a statement
+  // about which events the browser will deliver. Chrome on Keiven's desktop reports BOTH
+  // false while delivering real hover — the cell matched CSS `:hover` and the lazy content
+  // builder ran on `pointerover`, but no `pointerenter` listener existed to open the popup,
+  // so hovering did nothing at all. The same gate silently removed keyboard access, because
+  // `focusin`/`focusout` sat behind it too.
+  //
+  // So: listeners are always attached (below), and the ONE thing that genuinely needs to
+  // know the input device — whether a tap should pin — is decided from the actual
+  // `PointerEvent.pointerType` of the interaction in hand, which is a fact rather than a
+  // guess. `pendingPointerType` is written on `pointerdown`, which always precedes the click
+  // it belongs to.
+  // CONSUMED by the click it belongs to, then cleared. Holding it until the next
+  // `pointerdown` made it stale rather than click-scoped: a keyboard activation issues
+  // no `pointerdown` at all, so after any touch the next Enter on a focused link would
+  // still read `touch`, call `preventDefault()` and pin — suppressing the link the
+  // keyboard user asked for. `pointercancel` clears it too, so an aborted gesture
+  // cannot arm a later click.
+  let pendingPointerType = '';
+  document.addEventListener('pointerdown', function (e) {
+    pendingPointerType = e.pointerType || '';
+  }, true);
+  document.addEventListener('pointercancel', function () {
+    pendingPointerType = '';
+  }, true);
+  // A click that came from a finger or stylus has no hover behind it, so it must open the
+  // popup and pin it. A mouse click must not: hover already showed the popup, and pinning
+  // is modal (see `hoverAllowed`), so a mouse pin would switch hover off page-wide.
+  // Unknown/empty pointerType (synthetic `.click()`, keyboard activation) is treated as
+  // NOT touch, matching the mouse/keyboard contract.
+  const consumeClickShouldPin = function () {
+    const t = pendingPointerType;
+    pendingPointerType = '';
+    return t === 'touch' || t === 'pen';
+  };
   let pinnedCell = null;
   function unpinCell(c) { if (c && c._ttipUnpin) { c._ttipUnpin(); } }
   // "Inside a pinnable element" must mean EXACTLY the set `attachTooltip` wired, so ask
@@ -786,18 +822,29 @@
       }, hideDelayMs);
     }
 
-    // Tap (or click) toggles a pinned tooltip. Taps INSIDE the tooltip (its links,
-    // the ✕) behave normally. On touch there's no hover, so a tap on the cell must
-    // open the tooltip instead of following the cell's own parser-source link —
-    // preventDefault blocks that navigation. On desktop, a click on the cell link
-    // still navigates; a click on the cell body pins.
+    // TOUCH ONLY. Tap toggles a pinned tooltip; taps INSIDE the tooltip (its links, the ✕)
+    // behave normally. Touch has no hover, so a tap on the cell must open the tooltip
+    // instead of following the cell's own parser-source link — preventDefault blocks that
+    // navigation. Without this, a touch user could not read a popup at all.
+    //
+    // There is deliberately NO click-to-pin where hover exists. Pinning is MODAL
+    // (`hoverAllowed()` below): while one cell is pinned, hover opens nothing anywhere
+    // else. That made a single stray click kill hover for the whole page until the pin
+    // was released, and in details view "click somewhere else to dismiss" almost always
+    // lands on another wired cell, which just moves the pin — so hover never came back.
+    // On a hover-capable device hover is the only affordance, and a click on a cell now
+    // does what the cell says it does: follow its parser-source link.
     cell.addEventListener('click', function (e) {
+      // Consume FIRST, unconditionally. Every early return below still ends this click,
+      // so leaving the provenance unconsumed on any of them hands a stale `touch` to the
+      // NEXT click — the defect consume-on-use exists to prevent, one branch over. A tap
+      // on the popup's own links or its ✕ takes this return, and used to leave it armed.
+      const shouldPin = consumeClickShouldPin();
       if (e.target.closest('.ttip')) { return; }
-      if (!hoverCapable) {
-        e.preventDefault();
-      } else if (e.target.closest('a, button, input, label')) {
-        return;
-      }
+      // Decided per-interaction, not per-page: only a finger or stylus pins. A mouse click
+      // falls through to the cell's own parser-source link, and never enters modal pin mode.
+      if (!shouldPin) { return; }
+      e.preventDefault();
       if (ttip.classList.contains('ttip-pinned')) { unpin(); } else { pin(); }
     });
     // Hover show/hide only where hover exists; on touch it just flickers. A pinned
@@ -810,20 +857,32 @@
     // their own. Hover resumes when the pin is released (✕, clicking the pinned cell
     // again, or clicking outside).
     const hoverAllowed = function () { return pinnedCell === null || pinnedCell === cell; };
-    if (hoverCapable) {
-      cell.addEventListener('pointerenter', function () {
-        if (hoverAllowed() && !ttip.classList.contains('ttip-pinned')) { scheduleShow(); }
-      });
-      cell.addEventListener('pointerleave', function () {
-        if (!ttip.classList.contains('ttip-pinned')) { scheduleHide(); }
-      });
-      cell.addEventListener('focusin', function () {
-        if (hoverAllowed()) { scheduleShow(); }
-      });
-      cell.addEventListener('focusout', function () {
-        if (!ttip.classList.contains('ttip-pinned')) { scheduleHide(); }
-      });
-    }
+    // Registered UNCONDITIONALLY — see the `pendingPointerType` note above. If a device has no
+    // hover these simply never fire; if it does, they work regardless of what the media
+    // queries claim. `focusin`/`focusout` additionally give keyboard users the popup.
+    const onEnter = function () {
+      if (hoverAllowed() && !ttip.classList.contains('ttip-pinned')) { scheduleShow(); }
+    };
+    const onLeave = function () {
+      if (!ttip.classList.contains('ttip-pinned')) { scheduleHide(); }
+    };
+    cell.addEventListener('pointerenter', onEnter);
+    cell.addEventListener('pointerleave', onLeave);
+    // ALSO the classic mouse events, deliberately not instead. `pointerenter` is the
+    // modern spelling, but it does not fire on movement when the browser classifies the
+    // input as a touch pointer — and a device can report no hover capability, deliver a
+    // real mouse, and still surprise us about which family of events it sends. These two
+    // are what every browser has emitted for a moving mouse for twenty years. Both paths
+    // funnel into the same scheduleShow/scheduleHide, which are idempotent (they clear
+    // their own timers), so a browser sending BOTH opens exactly one popup.
+    cell.addEventListener('mouseenter', onEnter);
+    cell.addEventListener('mouseleave', onLeave);
+    cell.addEventListener('focusin', function () {
+      if (hoverAllowed()) { scheduleShow(); }
+    });
+    cell.addEventListener('focusout', function () {
+      if (!ttip.classList.contains('ttip-pinned')) { scheduleHide(); }
+    });
   }
   // The elements present at load. `th.case-sub` carries the per-column grammar popup (the
   // same case in every family's grammar); it uses the identical hover/pin machinery as a
--- a/parsers/v2/CUSTOM_PARSERS.md
+++ b/parsers/v2/CUSTOM_PARSERS.md
@@ -52,7 +52,9 @@ impl UnifiedParser for AcmeParser {
 
 This example is compiled and run as an integration test — [`tests/vendor_parser_example.rs`](tests/vendor_parser_example.rs). It uses only this crate's public API, exactly as your crate would.
 
-The block above and that file are compared **by a test**, not by hand: `doc_example_matches_compiled_example` parses the first Rust block out of this file and asserts its `parse_into` and `finish` bodies equal the compiled ones. That check exists because the claim it replaces was false — the two drifted, the documented example double-emitted its buffer, and nothing caught it.
+The block above and that file are compared **by a test**, not by hand: `doc_trait_method_bodies_match_compiled_example` parses the first Rust block out of this file and asserts that the two implement the same trait METHODS with the same bodies. That check exists because the claim it replaces was false — the two drifted, the documented example double-emitted its buffer, and nothing caught it.
+
+Know its limit: the test does not COMPILE this Markdown block, and it compares only the `impl UnifiedParser` methods. Renaming a struct field or changing an import here alone would break the documented example while the test stays green. The compiled file is the authority; treat this block as a copy that is guarded at the method level.
 
 `UnifiedParserOutput` gives you `push_text`, `push_reasoning` and `push_call`. The first two coalesce: appending text onto a trailing text event extends it rather than adding a second one.
 
--- a/parsers/v2/CUSTOM_PARSERS.md
+++ b/parsers/v2/CUSTOM_PARSERS.md
@@ -43,6 +43,10 @@ impl UnifiedParser for AcmeParser {
         out.push_text(std::mem::take(&mut self.buffered));
         Ok(out)
     }
+
+    fn reset(&mut self) -> String {
+        std::mem::take(&mut self.buffered)
+    }
 }
 ```
 
--- a/parsers/v2/CUSTOM_PARSERS.md
+++ b/parsers/v2/CUSTOM_PARSERS.md
@@ -10,7 +10,7 @@ One method is required. `finish` is required too, because a parser that forgets 
 
 ```rust
 use dynamo_parsers_v2::{
-    Tool, UnifiedParser, UnifiedParserEvent, UnifiedParserOutput,
+    Tool, UnifiedParser, UnifiedParserEvent, UnifiedParserExt, UnifiedParserOutput,
 };
 use anyhow::Result;
 
@@ -58,6 +58,8 @@ Know its limit: the test does not COMPILE this Markdown block, and it compares o
 
 `UnifiedParserOutput` gives you `push_text`, `push_reasoning` and `push_call`. The first two coalesce: appending text onto a trailing text event extends it rather than adding a second one.
 
+The required lifecycle is `initialize` → `parse_into` → `finish`, with `reset` for recovery. `push` and `parse_complete` come from the blanket `UnifiedParserExt` trait: import it when you want those allocation conveniences. They cannot be overridden through `impl UnifiedParser`, so `parse_complete` always runs the same `parse_into` + `finish` path as streaming.
+
 ## 2. Register it
 
 ```rust
@@ -97,7 +99,7 @@ These are the responsibilities a unified parser carries, and the reasons they ex
 | **Split-invariance** | The same bytes must produce the same ASSEMBLED output regardless of where the transport split them. Test every split point around your markers, not just a few. Raw event boundaries may legitimately differ between splits — one chunking can commit `Text("alpha ")` + `Text("rest")` where another commits `Text("alpha rest")` — which is why the check is `assembled()`, not the event vector. |
 | **Never leak your own markup** | Bytes you consumed as structure must not reappear in visible text. |
 | **Arguments keep their meaning and order** | Do not drop, duplicate, or corrupt argument values, and preserve the model's key order where the family contract depends on it. If you emit incremental fragments, the fragments you chose must concatenate into the intended arguments JSON. Verbatim BYTES are only required where the model already emits API-shaped JSON — the shipped XML families deliberately schema-type and re-serialize, then restore source key order. |
-| **Recover, do not panic** | Malformed or truncated output is normal. Emit what you can and drop what you cannot; returning an error should be reserved for genuinely unusable input. When driven through `parse_into`, whatever you already appended to the caller's output stays committed on `Err`; `push` owns its buffer and returns `Result<Vec<_>>`, which has nowhere to carry partial output, so it cannot honour that guarantee. |
+| **Recover, do not panic** | Malformed or truncated output is normal. Emit what you can and drop what you cannot; returning an error should be reserved for genuinely unusable input. When driven through `parse_into`, whatever you already appended to the caller's output stays committed on `Err`; the `UnifiedParserExt::push` helper owns its buffer and returns `Result<Vec<_>>`, which has nowhere to carry partial output, so it cannot honour that guarantee. |
 | **Flush on `finish`** | Do not silently drop the tail of a stream. What to DO with an unterminated channel is your family's policy — the shipped Qwen parser promotes open reasoning rather than leaking it as text, but a grammar with no reasoning channel, or one that treats an unterminated opener as visible recovery text, is making a different and equally valid call. State yours. |
 | **Override `reset` if you buffer** | The default returns an empty string and clears nothing. A parser that holds bytes back MUST override it, or a caller following the documented recovery path after a `parse_into` error resumes on your stale buffer and mis-numbers tool indices. Reset every field that carries stream position, tool index included — the text you hand back is a NEW stream. |
 
@@ -125,9 +127,9 @@ A family with no cases would otherwise report as covered while nothing measured 
 
 ## 7. Alignment with peer traits
 
-This trait is aligned with the peer streaming-parser traits other serving engines expose: `parse_into` is the required method, `finish` returns an output buffer, the event type carries the same variants in the same order, and `initialize` takes prompt token IDs. A parser written against a peer trait ports here mostly by renaming.
+The required `UnifiedParser` lifecycle is aligned with the peer streaming-parser traits other serving engines expose: `parse_into` is the required advance method, `finish` returns an output buffer, the event type carries the same variants in the same order, and `initialize` takes prompt token IDs. A parser written against a peer trait ports here mostly by renaming.
 
 Two caveats worth knowing before you plan on a literal drop-in:
 
 - Rust is nominally typed, so an identically-shaped type in another crate is still a different type. Porting is a mechanical translation, not a recompile.
-- This crate adds surface the peer traits do not have — `parse_complete` and the assembled `UnifiedEvent` view. Both are additive, so a peer-shaped caller never sees them, and a peer-shaped parser that does not provide them falls back to their defaults.
+- This crate adds surface the peer traits do not have — the blanket `UnifiedParserExt` helpers (`push` and `parse_complete`) and the assembled `UnifiedEvent` view. They are additive conveniences, not vendor-overridable lifecycle methods.
--- a/parsers/v2/CUSTOM_PARSERS.md
+++ b/parsers/v2/CUSTOM_PARSERS.md
@@ -86,17 +86,17 @@ This is deliberately supported. Disagreeing with one of our families should cost
 
 ## 4. What the contract requires
 
-These are the properties the conformance corpus checks, and the reasons it checks them. A parser that violates one will look correct in a demo and fail in production.
+These are the responsibilities a unified parser carries, and the reasons they exist. A parser that violates one will look correct in a demo and fail in production. The built-in conformance corpus exercises several of them for the families it ships; it cannot check factory/global-state isolation or your parser's own `reset` and error behaviour, and it does not run your parser at all until you enrol it (see below).
 
 | | Requirement |
 |---|---|
 | **One parser per stream** | A factory builds one parser for one choice of one request. Keep per-stream state in the parser, never in the factory or a global. |
 | **Order is the output** | Emit events in the order the model produced them. Do not hoist all reasoning to the front; that is precisely the defect the ordered event stream exists to remove. |
-| **Split-invariance** | The same bytes must produce the same events regardless of where the transport split them. Test every split point around your markers, not just a few. |
+| **Split-invariance** | The same bytes must produce the same ASSEMBLED output regardless of where the transport split them. Test every split point around your markers, not just a few. Raw event boundaries may legitimately differ between splits — one chunking can commit `Text("alpha ")` + `Text("rest")` where another commits `Text("alpha rest")` — which is why the check is `assembled()`, not the event vector. |
 | **Never leak your own markup** | Bytes you consumed as structure must not reappear in visible text. |
-| **Argument bytes are verbatim** | Forward argument fragments exactly as the model emitted them. Do not reformat, reorder keys, or re-serialize. |
-| **Recover, do not panic** | Malformed or truncated output is normal. Emit what you can and drop what you cannot; returning an error should be reserved for genuinely unusable input. On `Err`, whatever you already appended stays committed. |
-| **Flush on `finish`** | Open reasoning is promoted, not dropped and not leaked as text. |
+| **Arguments keep their meaning and order** | Do not drop, duplicate, or corrupt argument values, and preserve the model's key order where the family contract depends on it. If you emit incremental fragments, the fragments you chose must concatenate into the intended arguments JSON. Verbatim BYTES are only required where the model already emits API-shaped JSON — the shipped XML families deliberately schema-type and re-serialize, then restore source key order. |
+| **Recover, do not panic** | Malformed or truncated output is normal. Emit what you can and drop what you cannot; returning an error should be reserved for genuinely unusable input. When driven through `parse_into`, whatever you already appended to the caller's output stays committed on `Err`; `push` owns its buffer and returns `Result<Vec<_>>`, which has nowhere to carry partial output, so it cannot honour that guarantee. |
+| **Flush on `finish`** | Do not silently drop the tail of a stream. What to DO with an unterminated channel is your family's policy — the shipped Qwen parser promotes open reasoning rather than leaking it as text, but a grammar with no reasoning channel, or one that treats an unterminated opener as visible recovery text, is making a different and equally valid call. State yours. |
 | **Override `reset` if you buffer** | The default returns an empty string and clears nothing. A parser that holds bytes back MUST override it, or a caller following the documented recovery path after a `parse_into` error resumes on your stale buffer and mis-numbers tool indices. Reset every field that carries stream position, tool index included — the text you hand back is a NEW stream. |
 
 ## 5. Optional capabilities
@@ -111,13 +111,15 @@ All have defaults, so implement only what your grammar needs.
 
 ## 6. Check it against the corpus
 
-The conformance corpus is the useful part of this repo. Point it at your parser to find the cases you have not thought of: malformed envelopes, markers split mid-token, marker-looking text inside a JSON string, reasoning interleaved with calls, guided payloads that fail schema rather than syntax.
+The conformance corpus is the useful part of this repo — the cases you have not thought of: malformed envelopes, markers split mid-token, marker-looking text inside a JSON string, reasoning interleaved with calls.
+
+Running the workspace command below does NOT exercise a vendor parser. There is no flag that points the suite at a registered family: the suite iterates `builtin_unified_families()`, and a vendor registration deliberately does not enrol itself there. Adding your family means adding cases to the corpus and wiring it into the harness. Until you do, a green run says nothing about your parser.
 
 ```bash
 cargo test --workspace --all-targets --locked
 ```
 
-`builtin_unified_families()` is what the suite iterates, and a vendor registration deliberately does NOT enrol itself there — a family with no cases would otherwise report as covered while nothing measured it. Add cases for your family before trusting a green run.
+A family with no cases would otherwise report as covered while nothing measured it, which is why enrolment is deliberate rather than automatic.
 
 ## 7. Alignment with peer traits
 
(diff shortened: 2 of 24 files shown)


keivenchang:
@coderabbitai review

@devin review

Context for both: this is the peer-trait + vendor-registry half of a two-way split of #174. It is deliberately behaviour-neutral — main's unified golden feed and this branch's are the same sha256 (`169880fb…`), so the review surface is API shape and the new process-global registry, not parser output.

Worth the most scrutiny:

- `parsers/v2/src/unified/mod.rs` — the registry is process-global mutable state consulted on every parser construction. Registration is documented as startup-only and is NOT linearizable against concurrent construction; I want that assumption challenged.
- Vendor registrations canonicalise family aliases, so registering `qwen3` also shadows `qwen3_coder`. An earlier version keyed on the caller's spelling and silently ran two different parsers for one family depending on routing.
- `parsers/v2/tests/vendor_parser_example.rs` is the compiled form of the `CUSTOM_PARSERS.md` example. If those two drift, the docs are wrong and nothing else catches it.

keivenchang:
@coderabbitai full review

The head you analysed (`8c785de7`) no longer exists — this branch was rebased onto main after #140 landed, and force-pushed. Current head is `a086bd3d`. Your earlier pass posted an analysis chain but no findings, so please re-review from scratch rather than treating prior commits as reviewed.

What changed in the rebase, since it is material:

- Three conflicts in `build_stream_fixtures.py`, `fixtures.py` and `resolve_stream_fixtures.py`, all the same class: main renamed a case-schema key to `exception`, this branch still used `errors`. Resolved by taking main's side — main's version is a superset (normalisation on both mappings, mutual exclusivity documented) and is used consistently across five files.
- Seven conformance-tooling files reverted to main entirely, because #140 landed a better version of the same work and this PR has no hover or parser content in them. The PR shrank 24 files -> 14.
- A CSS assertion still expecting `44px` after the cell narrowing to `26px`.

Worth the most scrutiny on this head:

- `parsers/v2/src/unified/mod.rs` — the vendor registry is process-global mutable state consulted on every parser construction. Registration is documented as startup-only and is NOT linearizable against concurrent construction. Challenge that assumption.
- Alias canonicalisation: registering `qwen3` must also shadow `qwen3_coder`. An earlier version keyed on the caller's spelling and silently ran two different parsers for one family depending on routing.
- `parsers/v2/tests/vendor_parser_example.rs` is the compiled form of the `CUSTOM_PARSERS.md` example. If those drift, the docs are wrong and nothing else catches it.

This PR is deliberately behaviour-neutral: main's unified golden feed and this branch's are the same sha256, so parser output is not the review surface — API shape and the registry are.

parsers/v2/src/unified/mod.rs
```diff
@@ -92,42 +96,200 @@ pub enum UnifiedEvent {
...
+
+    /// Return the parser to a FRESH-STREAM state and hand back any unconsumed text.
+    ///
+    /// This is not a mid-turn continuation hook. Everything restarts, including the
+    /// tool index, so the returned text must be re-parsed as a NEW stream and any
+    /// calls already emitted belong to the abandoned one — feeding the remainder back
+    /// into the same turn would re-number from index 0 and collide with them.
+    fn reset(&mut self) -> String {
+        String::new()
+    }
```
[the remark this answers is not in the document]
keivenchang: Valid, fixed in `2fd0eba6`.

Confirmed before fixing: `impl UnifiedParser for ScannerUnified` implemented only `parse_into` and `finish`, so the trait default applied — returning `""` while the scanner kept `buffer`, `in_block`, `in_reasoning` and `next_index`. Exactly as you describe, a caller following the documented recovery path after a `parse_into` error would resume on stale state and mis-number indices.

Added `WrappedBlockScanner::reset_stream`, which drains the buffer and resets every field carrying stream position. `next_index` resets too, deliberately: the returned text is a NEW stream, so a call parsed out of it must not reuse an index already emitted from the abandoned one. `in_reasoning` restores to the family-declared `forced_start` rather than plain `false`, so a family whose prompt opens reasoning resets to its own default rather than to a state it never starts in.

Kept `reset` as a defaulted method rather than making it required: it is part of the peer trait surface this PR aligns to, and a vendor parser with no buffering has nothing to hand back.

conformance/utils/src/assets/conformance.js
```diff
@@ -674,10 +674,33 @@
...
+  let lastPointerType = '';
+  document.addEventListener('pointerdown', function (e) {
+    lastPointerType = e.pointerType || '';
+  }, true);
+  // A click that came from a finger or stylus has no hover behind it, so it must open the
+  // popup and pin it. A mouse click must not: hover already showed the popup, and pinning
+  // is modal (see `hoverAllowed`), so a mouse pin would switch hover off page-wide.
+  // Unknown/empty pointerType (synthetic `.click()`, keyboard activation) is treated as
+  // NOT touch, matching the mouse/keyboard contract.
+  const clickShouldPin = function () { return lastPointerType === 'touch' || lastPointerType === 'pen'; };
```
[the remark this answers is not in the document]
keivenchang: Valid, fixed in `2fd0eba6`.

You are right and my comment claimed the opposite — it said unknown/empty `pointerType` (synthetic click, keyboard activation) is treated as not-touch, which only held while the variable was empty. After any touch it stayed `touch` indefinitely, so the next keyboard activation would `preventDefault()` and pin, suppressing the navigation.

Provenance is now consumed by the click it belongs to and cleared, so it cannot arm a later activation. `pointercancel` clears it as well, so an aborted gesture cannot either.

Added `test_touch_then_keyboard_does_not_pin`: touch tap, then a keyboard-style click with no `pointerdown`, asserting the event was not `defaultPrevented` and no pin was entered. Negative-controlled — with the clear removed it fails, with it restored it passes.

parsers/v2/CUSTOM_PARSERS.md
```diff
@@ -0,0 +1,115 @@
...
+        // ... decide what is reasoning, what is visible text, what is a tool call ...
+        output.push_text(delta);
+        Ok(())
+    }
+
+    /// Called once at end of stream. Flush anything still buffered.
+    fn finish(&mut self) -> Result<UnifiedParserOutput> {
+        let mut out = UnifiedParserOutput::default();
+        // Whatever is still held back is ordinary text once the stream has ended.
+        out.push_text(std::mem::take(&mut self.buffered));
```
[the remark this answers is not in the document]
keivenchang: Valid, fixed in `2fd0eba6`.

Traced it: `push("hello")` then `finish()` did produce `hellohello`. The example appended each delta to the buffer, emitted the same delta, then emitted the buffer again at finish.

Your second point is the sharper one. That example exists specifically so the docs cannot rot, and it failed at exactly that — the compiled test used different buffering, so nothing compared them. The Markdown block is now the compiled implementation verbatim; both `parse_into` and `finish` bodies compare equal after whitespace and comment normalisation.

parsers/v2/src/unified/mod.rs
```diff
@@ -92,42 +96,200 @@ pub enum UnifiedEvent {
...
+    /// Feed one decoded text delta; returns the events it committed, in order.
+    ///
+    /// Additive convenience over [`UnifiedParser::parse_into`] — allocates a fresh
+    /// vector per advance, which is why a serving loop prefers `parse_into`. The
+    /// conformance corpus asserts against this spelling.
+    fn push(&mut self, chunk: &str) -> Result<Vec<UnifiedParserEvent>> {
+        let mut out = UnifiedParserOutput::default();
+        self.parse_into(chunk, &mut out)?;
+        Ok(out.events)
+    }
```
KrishnanPrash: **AI Generated Comment**

```rust
fn push(&mut self, chunk: &str)
  -> Result<Vec<UnifiedParserEvent>> {
    let mut out =
      UnifiedParserOutput::default();
    self.parse_into(chunk, &mut out)?;
    Ok(out.events)
}
```

Line 116 documents the error contract as "on `Err`, whatever was already appended stays committed"; `CUSTOM_PARSERS.md:92` repeats it to vendors. The `?` drops `out`, and the committed events go with it. Latent today because `ScannerUnified` appends nothing on its error path, but `parse_complete` and the conformance corpus both route through `push`. Fix `push`, or narrow the promise to `parse_into`.

```
vendor parser doing the documented
thing:  out.push_text(d); bail!()

parse_into -> Err, out holds
              [Text("hello")]
push       -> Err, "hello" gone
```

parsers/v2/src/unified/mod.rs
```diff
@@ -222,40 +384,241 @@ pub(crate) struct ScannerUnified<E: InvokeEmitter> {
...
+    /// recovery path after a `parse_into` error would then resume on stale state and
+    /// mis-number tool indices. The only shipped family must honour the contract.
+    fn reset(&mut self) -> String {
+        self.scanner.reset_stream()
+    }
+
+    fn parse_into(&mut self, delta: &str, output: &mut UnifiedParserOutput) -> Result<()> {
+        output.events.extend(self.scanner.push_ordered(delta)?);
+        Ok(())
     }
```
KrishnanPrash, line 497: **AI Generated Comment**

```rust
// built-in (line 401): appends, no coalesce
output.events.extend(
  self.scanner.push_ordered(delta)?);

// documented helper (line 202): coalesces
// into the PREVIOUS event, in place
if let Some(Text(last)) =
    self.events.last_mut() {
    last.push_str(d.as_ref());
    return;
}
```

Per-advance or cumulative is never stated, and the two shipped implementations answer differently. The answer also decides whether the watermark loop the docs steer serving code toward is legal. Neither effect is visible today: `assemble` coalesces afterwards, and every in-repo caller starts from a fresh buffer. Pin it on the `UnifiedParserOutput` doc comment, then make the two agree. `append` at line 192 has the same gap.

```
two advances, "hel" then "lo",
into ONE buffer:

builtin qwen3 -> [Text("hel"),
                  Text("lo")]
vendor helper -> [Text("hello")]

the watermark loop:
let n = out.len();
parse_into("lo", &mut out)?;
&out[n..]  ->  []    // "lo" landed
                     // in out[n-1]
```

parsers/v2/src/unified/mod.rs
KrishnanPrash, line 147: **AI Generated Comment**

```rust
fn reset(&mut self) -> String {
    String::new()
}
```

The `ScannerUnified` override at line 392 states the hazard: the default reports nothing buffered while the scanner still holds a partial marker, `in_block` and a used `next_index`, so a caller following the documented recovery path resumes on stale state and mis-numbers tool indices. That applies to every vendor parser that buffers and does not override. `CUSTOM_PARSERS.md:104` lists `reset()` as optional while the trait doc presents it as the recovery path after a `parse_into` error.

```
minimum: a row in the section 4
contract table, "a parser that
buffers MUST override reset"

airtight:
fn reset(&mut self)
  -> Option<String>
  None     = not supported
  Some("") = nothing buffered
```

parsers/v2/CUSTOM_PARSERS.md
```diff
@@ -0,0 +1,123 @@
...
+    /// once the stream has ended.
+    fn finish(&mut self) -> Result<UnifiedParserOutput> {
+        let mut out = UnifiedParserOutput::default();
+        out.push_text(std::mem::take(&mut self.buffered));
+        Ok(out)
+    }
+}
+```
+
+This example is compiled and run as an integration test — [`tests/vendor_parser_example.rs`](tests/vendor_parser_example.rs). It uses only this crate's public API, exactly as your crate would, so if these instructions ever stop being true the build fails here first. Copy from that file if you want something that definitely compiles.
```
KrishnanPrash: **AI Generated Comment**

Line 49 claims "if these instructions ever stop being true the build fails here first". They stopped being true on this PR and the build did not fail. Matching the two files again restores the invariant but leaves it resting on a manual comparison, which is exactly what failed.

```
the defect that shipped:
  push("hello") + finish()
  -> "hellohello"
the compiled example buffered
differently, so nothing compared
the two.

structural fix, ~20 lines:
let md = include_str!(
    "../CUSTOM_PARSERS.md");
assert_eq!(
  normalize(first_rust_block(md)),
  normalize(impl_of(include_str!(
    "vendor_parser_example.rs"))));

or let the doctest harness do it:
#![doc = include_str!(
    "../CUSTOM_PARSERS.md")]
```

parsers/v2/src/unified/mod.rs
```diff
@@ -222,40 +384,241 @@ pub(crate) struct ScannerUnified<E: InvokeEmitter> {
...
+                     dynamo_parsers_v2::register_unified_parser(\"{other}\", your_factory) \
+                     before serving.",
+                    REGISTERED_UNIFIED_FAMILIES,
+                    vendor_unified_families(),
+                ),
+            };
+            let canonical = match family {
+                $($family $(| $alias)* => $family,)+
+                _ => unreachable!("matched above"),
+            };
```
KrishnanPrash: **AI Generated Comment**

BEFORE:
```rust
let canonical = match family {
    $($family $(| $alias)* =>
        $family,)+
    _ => unreachable!("matched above"),
};
```

AFTER:
```rust
let canonical =
    canonical_unified_family(family)
        .unwrap_or(family);
```

The vendor branch at line 569 already resolves the canonical name with that helper. The second copy costs an arm per family plus a panic that exists only because the compiler cannot see the `bail!` above it. Raised by CodeRabbit, still open.

parsers/v2/src/unified/mod.rs
```diff
@@ -222,40 +384,241 @@ pub(crate) struct ScannerUnified<E: InvokeEmitter> {
...
+    // than one routing name (`qwen3` and `qwen3_coder` are one grammar), and keying
+    // on the caller's spelling shadowed only the spelling they happened to use:
+    // `register_unified_parser("qwen3", ..)` left `qwen3_coder` on the built-in, so
+    // the same family silently ran two different parsers depending on how the
+    // request was routed. Canonicalizing on both sides is what makes "replace a
+    // family this crate ships" true for every name that family answers to.
+    let key = canonical_unified_family(family).unwrap_or(family);
+    let previous = VENDOR_PARSERS
+        .write()
+        .expect("vendor parser registry poisoned")
```
KrishnanPrash: **AI Generated Comment**

Also at lines 487, 495 and 511.

BEFORE:
```rust
.expect(
  "vendor parser registry poisoned")
```

AFTER:
```rust
.unwrap_or_else(|e| e.into_inner())
```

The map holds `String` keys and `fn` pointers; neither can be left half-written by a panic, so poisoning carries no information here and propagating it turns a table read into a process kill. `tests/vendor_registry.rs:54` already uses the tolerant form.

KrishnanPrash (commented):
**AI Generated Comment**

Six comments, all in `parsers/v2`. Four are the same shape: the contract this PR publishes and the code implementing it disagree, in ways `assemble` currently hides, which is why the suite is green. Since handing this surface to people outside the crate is the whole point, those seem worth settling before the API is out. The last two are small.

Verified against head `2fd0eba` locally: `cargo test --workspace --all-targets --locked` gives 1372 passed, `cargo clippy --workspace --all-targets` gives 0 warnings, and the `conformance/fixtures` tree plus both LFS pointers are byte-identical to `main`. The evidence in the description holds.

parsers/v2/src/unified/mod.rs
keivenchang: Valid, and the sharpest of the six. Fixed in `12a16a27`.

You are right that the answer was never stated and the two shipped implementations disagreed. Pinned it as **cumulative, and appending coalesces**, then made the built-in agree: `parse_into` now routes through `push_text`/`push_reasoning`/`push_call` instead of `events.extend`.

The `UnifiedParserOutput` doc now states both consequences you identified — that a watermark window can legally observe nothing because the bytes merged into `events[n - 1]`, with `push` named as the spelling that answers "what did THIS advance produce"; and that appending must go through the helpers, since `extend` bypasses the merge. `append` is documented as the exception: it joins two independently-built buffers and does not coalesce across the seam.

Golden feed is byte-identical after the change, as expected — `assemble` performed the same fold downstream, which is exactly why this was invisible.

parsers/v2/src/unified/mod.rs
keivenchang: Valid. Fixed in `12a16a27`, though not the way I first tried.

Confirmed the mechanism: `?` drops `out` and the committed events go with it, while line 116 and `CUSTOM_PARSERS.md` both promise they stay committed.

I first tried attaching the partial output to the error, then backed it out — `Result<Vec<_>>` genuinely has nowhere to carry it, and inventing an error type to smuggle it through would have been worse than being honest. So I took your second option: the guarantee is now stated on `parse_into`, whose buffer the CALLER owns and can still read after an error, and `push` says explicitly that a parser which may commit events and then fail must be driven through `parse_into`.

parsers/v2/src/unified/mod.rs
keivenchang: Valid. Fixed in `12a16a27`.

You are right that the hazard I documented on the `ScannerUnified` override applies to every vendor parser that buffers, and that listing `reset()` under optional capabilities contradicted the trait doc presenting it as the post-error recovery path.

Took the minimum you specified: it is now a row in the section 4 contract table — a parser that buffers MUST override `reset`, and must reset every field carrying stream position, tool index included, because the text handed back is a new stream. Removed it from the optional table so the two no longer disagree.

Did not take the `Option<String>` signature. It would distinguish not-supported from nothing-buffered, but `reset` is part of the peer trait surface this PR exists to align to, and changing its shape would defeat that.

parsers/v2/CUSTOM_PARSERS.md
keivenchang: Valid, and the criticism lands: the claim was falsified by this very PR.

Fixed in `12a16a27` with the structural check rather than another hand-match. `doc_example_matches_compiled_example` parses the first Rust block out of `CUSTOM_PARSERS.md` and asserts its `parse_into` and `finish` bodies equal the compiled ones, whitespace- and comment-insensitive so formatting cannot fail it spuriously.

Negative-controlled: reintroducing the double-emit in the Markdown only makes it fail with "CUSTOM_PARSERS.md and vendor_parser_example.rs disagree on `parse_into`"; restoring it passes. The prose now names the test instead of asserting a build-failure that was not happening.

parsers/v2/src/unified/mod.rs
keivenchang: Valid. Fixed in `12a16a27` — replaced with `canonical_unified_family(family).unwrap_or(family)`, exactly as written.

You are right that the vendor branch already resolved it that way, and that the generated match cost an arm per family plus an `unreachable!` that existed only because the compiler cannot see the `bail!` above it.

parsers/v2/src/unified/mod.rs
keivenchang: Valid. Fixed in `12a16a27` at all four sites.

The map holds `String` keys and `fn` pointers, so a panic cannot leave it half-written and the poison carries no information — propagating it turned a table read into a process kill. Now `.unwrap_or_else(|e| e.into_inner())`, matching what `tests/vendor_registry.rs:54` already did.

parsers/v2/src/lib.rs
```diff
@@ -26,6 +26,8 @@ pub use tool_calling::{CalledFunctionStream, ToolCallResponseChunk, ToolCallType
 // One state machine per stream owning reasoning + content + tool calls, emitting
 // ONE ordered event stream (see `unified`).
 pub use unified::{
-    REGISTERED_UNIFIED_FAMILIES, UnifiedDelta, UnifiedEvent, UnifiedParser, assemble,
-    create_unified_parser_for_family,
+    REGISTERED_UNIFIED_FAMILIES, UnifiedEvent, UnifiedParser, UnifiedParserEvent,
```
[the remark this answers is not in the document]
keivenchang: Valid. Fixed in `ddaa264c` — bumped `dynamo-parsers-v2` 0.1.27 -> 0.2.0, with `conformance` and `parsers/v2-py` path dependencies and `Cargo.lock` updated to match.

You are right that this is breaking: the public `UnifiedDelta` export is gone and the `UnifiedParser` trait surface changed (`parse_into` is now the required method, `finish` returns `UnifiedParserOutput`). Under Cargo's 0.x rules the MINOR position is the breaking one, so 0.2.0 is the honest version rather than another patch.

Took the version bump rather than deprecated compatibility wrappers, for two reasons. The enum's variants changed SHAPE, not just name — `Text { text }` became `Text(String)` to match the peer traits — so a type alias or wrapper could not preserve pattern matching at a call site, which is where downstream code actually breaks. And the crate has no external consumers today: every in-repo dependant is updated in this change, and `UnifiedDelta` now appears nowhere in the tree. Shipping deprecated shims for a migration nobody has to perform would add surface to a crate this PR is trying to make adoptable.

The commit message records that rationale so the next reader does not have to reconstruct it.

conformance/utils/src/assets/conformance.js
```diff
@@ -786,18 +822,24 @@
       }, hideDelayMs);
     }
 
-    // Tap (or click) toggles a pinned tooltip. Taps INSIDE the tooltip (its links,
-    // the ✕) behave normally. On touch there's no hover, so a tap on the cell must
-    // open the tooltip instead of following the cell's own parser-source link —
-    // preventDefault blocks that navigation. On desktop, a click on the cell link
-    // still navigates; a click on the cell body pins.
+    // TOUCH ONLY. Tap toggles a pinned tooltip; taps INSIDE the tooltip (its links, the ✕)
+    // behave normally. Touch has no hover, so a tap on the cell must open the tooltip
```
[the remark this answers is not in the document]
keivenchang: Valid, and a hole in my own fix. Fixed in `ddaa264c`.

You are right that the `.ttip` early return ran before the consume, so a touch tap on the popup's own links or its ✕ left `pendingPointerType` armed for the next click — the exact defect consume-on-use exists to prevent, one branch over from where I fixed it.

Moved the consume ahead of every early return rather than adding a clear to that one branch: the value is now taken unconditionally at the top of the handler and the branches test the resulting boolean, so a future early return cannot reintroduce this by forgetting to clear.

Extended `test_touch_then_keyboard_does_not_pin` to cover the path: tap inside a tooltip, then a keyboard-style click with no `pointerdown`, asserting it was not `defaultPrevented`. It failed before this fix and passes after — I had the test in place before the JS change landed, so that ordering was verified rather than assumed.

parsers/v2/src/tool_calling/scan.rs
```diff
@@ -432,8 +428,23 @@ impl<E: InvokeEmitter> WrappedBlockScanner<E> {
...
 
-    fn drain(&mut self, flush: bool) -> anyhow::Result<Vec<UnifiedDelta>> {
-        let mut out: Vec<UnifiedDelta> = Vec::new();
+    /// Return to a FRESH-STREAM state and hand back whatever was still buffered.
+    ///
+    /// Every field that carries stream position resets, `next_index` included: the
+    /// returned text is a NEW stream, so a call parsed out of it must not reuse an
+    /// index already dispatched from the abandoned one.
+    pub(crate) fn reset_stream(&mut self) -> String {
+        let carried = std::mem::take(&mut self.buffer);
```
indrajit96, line 526: reset_stream() only returns bytes still in buffer, but drain() removes bytes before calling the fallible emitter. For example, with:
`prefix<tool_call><function=f>...</function></tool_call>suffix`
if `parse_invoke()` fails, prefix is trapped in the discarded local output and the invoke has already been removed. reset_stream() can return only the remaining </tool_call>suffix.
This breaks the new recovery contract. 
Please restore the scanner to its pre-call state on emitter errors and add failing-emitter tests for wrapped and bare invokes.

parsers/v2/src/unified/mod.rs
```diff
@@ -222,40 +422,250 @@ pub(crate) struct ScannerUnified<E: InvokeEmitter> {
 }
 
 impl<E: InvokeEmitter + Send> UnifiedParser for ScannerUnified<E> {
-    fn push(&mut self, chunk: &str) -> Result<Vec<UnifiedDelta>> {
-        self.scanner.push_ordered(chunk)
+    // No `preserve_special_tokens` override here: the trait METHOD is the peer surface
```
indrajit96: Deferring this makes the unified and tool-only Qwen adapters report contradictory decoding requirements: `Qwen3CoderToolStreamParser::preserve_special_tokens()` returns true, while this adapter inherits false despite using the same grammar.

parsers/v2/src/unified/mod.rs
```diff
@@ -222,40 +422,250 @@ pub(crate) struct ScannerUnified<E: InvokeEmitter> {
...
+        self.scanner.reset_stream()
     }
 
-    fn finish(&mut self) -> Result<Vec<UnifiedDelta>> {
-        self.scanner.finish_ordered()
+    fn parse_into(&mut self, delta: &str, output: &mut UnifiedParserOutput) -> Result<()> {
+        // Through the helpers, NOT `events.extend`: the helpers coalesce into a trailing
+        // same-kind event and `extend` does not, so extending made the only shipped
+        // family disagree with the documented vendor contract for identical bytes.
+        for event in self.scanner.push_ordered(delta)? {
```
indrajit96: This creates two vectors for each push():
```
 push()
    -> UnifiedParserOutput.events
    -> push_ordered() creates another Vec
 ```
The scanner’s events are then moved from one vector to the other. 
Please let the scanner write directly into UnifiedParserOutput, or override ScannerUnified::push() to return push_ordered() directly.

**Review by indrajit96 — COMMENTED** [review]
Some higher level commetns
### 1. The parser factory receives too little information

  Today the factory looks like:

  fn create_parser(tools: &[Tool]) -> Box<dyn UnifiedParser>

  It receives only the available tools.

  That works for Qwen because its parser mostly needs the tool schemas. Future parsers may also need:

  - Which tokenizer/model variant is being used?
  - Did the prompt already open the reasoning channel?
  - Is the model producing native markup or guided JSON?
  - Must special tokens be preserved?
  - Does this deployment have vendor-specific configuration?

  For example, Kimi K3 might start in reasoning mode:

  Prompt already opened:
  <|open|>think<|sep|>

  Generated output starts directly with:
  I need to call the weather tool...

  The parser must know that those first generated bytes are reasoning. tools alone cannot tell it that.

  A richer factory could receive:
```
  ParserCreateContext {
      tools,
      model_config,
      tokenizer_info,
      starting_channel: Reasoning,
      output_mode: NativeMarkup,
  }
```
  ### 2. The scanner structure may be too specific to Qwen-like formats

  The current scanner appears to expect a mostly flat structure:
```
  BLOCK
    INVOKE
    INVOKE
  BLOCK_END
```
  Kimi K3 may have nested channels or sections. A simplified example is:
```
  reasoning
  └── tool section
      └── invoke
```
  A flat scanner may incorrectly close the outer block when it encounters the inner section, or emit events in the wrong order.

  Before adding more model families, could we clarify whether ScannerUnified is intended to support nested formats? If it is, the scanner likely needs an explicit stack or model-specific state-machine extension point rather than only
  configurable delimiters.

  ### 3. Kimi K3 (OR any NEW Model) needs a real incremental parser, not only a registry entry

  Adding Kimi K3 support ideally will involve more than?
```
  register_unified_parser("kimi_k3", create_kimi_parser);
```
  The parser must handle arbitrary streaming boundaries. For example, these chunks must behave exactly like one complete string:
```
  Chunk 1: "<|open|>to"
  Chunk 2: "ol<|sep|>{\"name\":"
  Chunk 3: "\"weather\"}..."
```
  We should avoid implementing this logic separately in the v2 tool parser and unified parser. Otherwise, fixes for partial tokens, nesting, or malformed output must be made twice.

  I suggest:
```
  One incremental Kimi K3 state machine
  ├── tool-only adapter
  └── unified-parser adapter
```
  Both adapters would then share identical buffering, token recognition, validation, and error handling.

Is that the design for new models?

File: parsers/v2/src/tool_calling/scan.rs (hunk 11 above)
Comment (keivenchang) [reply parsers/v2/src/tool_calling/scan.rs:526]: youre right, and it was worse than the single-invoke case — with `prefix<tool_call><function=ok></function><function=boom></function></tool_call>suffix` we also lost `prefix` and the `ok` call that had already succeeded, because drain() collected into a local vec that `?` dropped. fixed in 0b84b647: the drain loop writes through a caller-owned sink and each invoke is consumed only after its emitter succeeds, so committed events stay and reset() returns `<function=boom></function></tool_call>suffix`. injected-emitter regressions for first/later x wrapped/bare, driven through the public parse_into. i checked the peer while doing this and it behaves exactly as you described, so thanks — that was the deciding evidence.

File: parsers/v2/src/unified/mod.rs (hunk 12 above)
Comment (keivenchang) [reply parsers/v2/src/unified/mod.rs]: agreed, two surfaces over one scanner cant answer differently. fixed in 0b84b647 — preserve_special_tokens now lives on the grammar spec and both adapters read it, aggregated as the OR over tool+reasoning components (same rule as the peer CombinedParser). kept the shipped `true` rather than flipping it: peer qwen3 and the cached tokenizer both suggest `false` may be the correct value, but one tokenizer isnt an inventory and that flip is an observable behaviour change, so its a separate PR. parity tests for canonical, alias and tool-only.

File: parsers/v2/src/unified/mod.rs (hunk 13 above)
Comment (keivenchang) [reply parsers/v2/src/unified/mod.rs]: real, and fixed in 0b84b647 — though not by overriding push(). parse_into now writes straight into the caller-owned UnifiedParserOutput via the same sink the recovery fix needed, so the intermediate vec is gone from the serving path without adding a second advance implementation. worth noting the peer signature is identical to ours and their native parsers also write directly, so the extra vec was ours, not inherent to the shape.

File: parsers/v2/src/unified/mod.rs [hunk 14]
```diff
@@ -92,24 +96,251 @@ pub enum UnifiedEvent {
...
+    type IntoIter = std::slice::Iter<'a, UnifiedParserEvent>;
+    fn into_iter(self) -> Self::IntoIter {
+        self.events.iter()
+    }
+}
+
+impl FromIterator<UnifiedParserEvent> for UnifiedParserOutput {
+    fn from_iter<T: IntoIterator<Item = UnifiedParserEvent>>(iter: T) -> Self {
+        Self {
+            events: iter.into_iter().collect(),
```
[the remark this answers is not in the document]
keivenchang: good catch — fixed in ee2bd536. you're right that collect() bypassed the merge that append and the push helpers enforce. i'd fixed append earlier in this same PR and didn't audit the other constructors, which is exactly how this one survived. from_iter now routes through push_text/push_reasoning/push_call, and i added construction-parity tests asserting collect() and append both produce the same result as pushing the same events, so a future constructor can't drift on its own. also made ScannerUnified::finish write directly into the output like parse_into while i was in there. golden feed regenerated from scratch: byte-identical.

conformance/utils/src/assets/conformance.js
```diff
@@ -674,10 +674,46 @@
...
+  // `PointerEvent.pointerType` of the interaction in hand, which is a fact rather than a
+  // guess. `pendingPointerType` is written on `pointerdown`, which always precedes the click
+  // it belongs to.
+  // CONSUMED by the click it belongs to, then cleared. Holding it until the next
+  // `pointerdown` made it stale rather than click-scoped: a keyboard activation issues
+  // no `pointerdown` at all, so after any touch the next Enter on a focused link would
+  // still read `touch`, call `preventDefault()` and pin — suppressing the link the
+  // keyboard user asked for. `pointercancel` clears it too, so an aborted gesture
+  // cannot arm a later click.
+  let pendingPointerType = '';
```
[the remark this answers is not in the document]
keivenchang: real, fixed in 8be500ab. provenance now records the pointerdown target too and only pins when the gesture started inside the same host. worth noting the case that actually bites isnt a following mouse click — that carries its own pointerdown which overwrites the stale value — its an activation with no pointerdown at all (keyboard Enter, synthetic click). my first regression drove a mouse pointerdown and passed with the fix reverted, so i rewrote it; the committed one fails on the pre-fix asset. this is the sibling of the consume-on-use fix earlier in this PR: that stopped provenance outliving its click, this stops it belonging to a different element. 17 browser tests green.

keivenchang:
Yes — that is the design boundary. #178 only aligns the trait and adds startup registration for parsers that already exist; it intentionally does not claim the tools-only factory can construct tokenizer/model-dependent parsers or that `ScannerUnified` supports nested grammars. For a new family with distinct nesting, I would implement one model-specific incremental state machine and expose tool-only and unified adapters over it; request state/output mode belongs to per-request initialization, while tokenizer/model/vendor construction context needs a separate factory-context design before such a parser can register. I added both limitations to the Overview so this API is not presented as drop-in adoptable.

keivenchang:
@indrajit96 — all three are fixed, here's where. Heads up that your three threads show as resolved under my account; I didn't resolve them and I'd have left them for you, so please reopen any you're not happy with.

**1. `reset_stream()` losing bytes on emitter error — fixed in `0b84b647`.**

You were right, and it was worse than the single-invoke case. With `prefix<tool_call><function=ok></function><function=boom></function></tool_call>suffix` we also lost `prefix` and the `ok` call that had already succeeded, because `drain()` collected into a local vec that `?` dropped on the way out.

Two changes: the drain loop now writes through a sink the **caller** owns, so a committed event can't be retracted; and both invoke sites call the emitter **before** `buffer.drain(..)`, so a failing invoke stays recoverable. `reset()` now returns `<function=boom></function></tool_call>suffix`.

Regressions for first/later x wrapped/bare, driven through the public `parse_into` with exact output and exact reset bytes: `first_wrapped_failure_keeps_committed_text_and_recovers_the_invoke`, `later_wrapped_failure_keeps_the_call_that_already_succeeded`, and the two bare equivalents.

I checked vLLM while fixing this — their tool trait documents exactly this contract and `CombinedParser` tests it, and I reproduced their parser keeping `prefix` and returning the failed body from `reset()`. So the contract was achievable, not aspirational. Thanks, that was the deciding evidence.

**2. Contradictory `preserve_special_tokens` — fixed in `0b84b647`.**

Agreed, two surfaces over one scanner can't answer differently. The value now lives on the grammar spec and both adapters read it, aggregated as the OR over tool + reasoning components — same rule as vLLM's `CombinedParser`, which ORs its wrapped parsers. Parity tests for the canonical family, the alias, and the tool-only adapter.

**Partly deferred, deliberately:** I kept the shipped `true` rather than flipping it. vLLM's own Qwen parser returns `false`, and the cached Nemotron tokenizer marks `<tool_call>` as `special:false` — so `true` may itself be wrong. But one tokenizer isn't an inventory, and that flip is an observable behaviour change. **Upcoming PR:** pin the value with a repo-owned tokenizer inventory across every supported Qwen/Nemotron alias, changing both adapters together.

**3. Two vectors per `push()` — fixed in `0b84b647`, though not by overriding `push()`.**

Real. `parse_into` now writes straight into the caller's `UnifiedParserOutput` using the same sink the recovery fix needed, so the intermediate vec is gone from the serving path without adding a second advance implementation.

I went the other way on your second suggestion on purpose: overriding `ScannerUnified::push()` would only help the convenience spelling, leave `parse_into` unchanged, and create a second advance path that can drift. Worth noting vLLM's `parse_into` signature is identical to ours and their native parsers also write directly, so the extra vec was ours, not inherent to the shape.

Current head is `ad88face`; the golden feed is byte-identical to `main` throughout, so none of this changes parsed output.

parsers/v2/src/unified/mod.rs
```diff
@@ -92,24 +96,271 @@ pub enum UnifiedEvent {
...
-        let mut deltas = self.push(output)?;
-        deltas.append(&mut self.finish()?);
-        Ok(assemble(&deltas))
+        // Joined through `UnifiedParserOutput::append`, not a raw `Vec::append` on the
+        // event vectors: this is exactly the push/finish seam that rule exists for, and
+        // the crate should not violate its own documented merge semantics here. The
+        // assembled result is unchanged either way, since `assemble` performs the same
+        // fold — but the pre-fold event stream now agrees with every other spelling.
+        let mut acc = UnifiedParserOutput {
+            events: self.push(output)?,
```
indrajit96: `parse_complete()` is documented as using `parse_into()` and `finish()`, but it calls self.push().

push() can be overridden by a vendor parser. 

For example:
```
  parse_into() emits Text("A")
  push() emits Text("B")
```

`parse_complete()` currently returns "B"

This means batch and streaming can still use different parsing implementations.

Can we di this?
```
  let mut acc = UnifiedParserOutput::default();
  self.parse_into(output, &mut acc)?;
  acc.append(&mut self.finish()?);
  Ok(assemble(&acc.events))
```
  
That keeps `parse_into()` as the single parsing path promised by the trait documentation.

Not a blocker but something that can cause sublte issue when we add more models

parsers/v2/src/tool_calling/kimi_k2.rs
```diff
@@ -70,6 +70,8 @@ fn spec(config: &KimiK2ParserConfig) -> WrappedBlockSpec {
         bare_recovery_latch: BareRecoveryLatch::Set,
         invoke_latch: InvokeLatch::Always,
         drop_invoke_crossing_block_end: true,
+        // Every wrapped family's markers are special tokens today.
+        preserve_special_tokens: true,
```
indrajit96, line 74: The original Qwen mismatch is fixed, but the same value still has two sources for Kimi K2 and MiniMax M2/M3.

This comment also applies to:
```
  - parsers/v2/src/tool_calling/minimax_m2.rs
  - parsers/v2/src/tool_calling/minimax_m3.rs
```

indrajit96 (approved):
Thanks a lot for this massive lift off!!
@keivenchang 
Approved with a couple of nits I see nothing blocking

parsers/v2/src/unified/mod.rs
keivenchang: Fixed in 883a3195. I moved `push()` and `parse_complete()` off `UnifiedParser` onto the blanket `UnifiedParserExt` trait. `parse_complete()` now calls `parse_into()` and then `finish()` directly, and the vendor integration regression proves that a conflicting inherent `push()` cannot change registry-shaped batch behavior.

parsers/v2/src/tool_calling/kimi_k2.rs
keivenchang: Fixed in 883a3195. Kimi K2, MiniMax M2, and MiniMax M3 now return `self.scanner.preserve_special_tokens()`, so each scanner spec is the single source of truth instead of duplicating `true` in its adapter.

PR #179, opened 2026-08-10
feat(renderer): add `fromjson` chat-template filter

chanh:
## Problem

Published chat templates apply the `fromjson` filter to a tool call's `arguments` — which arrive as a JSON *string* — in order to iterate the decoded object. minijinja has no such builtin and the renderer doesn't register one, so the render aborts:

```
Failed to apply prompt template: unknown filter: filter fromjson is unknown (in tool_use:63)
```

The affected template branch is typically only reached once tool **results** are in the message history. Single-turn tool requests render fine, so this stays invisible until real multi-turn agentic traffic arrives.

We hit this in production on `stepfun-ai/step-3.7-flash`: **every multi-turn tool call returned a 500**, roughly a third of that model's traffic. It's especially easy to miss because the requests fail *before* tokenization (`input_tokens: 0`), so they record no input-sequence-length sample either — the mean ISL just reads low, as if the traffic mix had changed, rather than showing an error spike.

Reproduced against two deployments of the same checkpoint, identical request bodies:

| case | Python/vLLM (HF Jinja2) | Dynamo (minijinja) |
|---|---|---|
| multi-turn with `role:"tool"` result | 200, `prompt_tokens=292` | **500** |
| first-turn `tools` only | 200, 241 | 200, 241 |
| plain text, no tools | 200, 21 | 200, 21 |

The gap is specifically the environment: transformers registers a richer filter set for chat templates than the renderer currently mirrors, so templates relying on anything outside that subset fail per-model.

## Change

Register `fromjson`, plus a `from_json` alias — both spellings appear in published templates.

- Non-string values pass through untouched, so a template that defensively applies the filter to an already-decoded value keeps rendering.
- Malformed JSON surfaces as a template error rather than a panic.

## Tests

`cargo test -p dynamo-renderer` → **139 passed, 0 failed** (134 pre-existing + 5 new). New coverage: object-string parsing, array-string parsing, non-string pass-through, malformed-input error, and an end-to-end render of the `tool_use` block shape with JSON-string `arguments` — the exact pattern that fails today.

`cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` are both clean.

## Verification

Backported to the renderer version running in our production image and deployed to a shadow endpoint. During the rollout overlap, with both versions serving the same live traffic in the same window, `fromjson` render failures over 10 minutes were **27 and 33 on the old pods, 0 on the new ones**. The previously-500ing request returns 200 with `prompt_tokens` identical to the Python/vLLM deployment (292), and the generated text matches — so the template renders the *same* prompt, not merely a different one that happens to succeed.

Signed-off per DCO.


chanh:
Bumping this for a vetter — it has been waiting on `/ok to test` since 08-10, so no CI has compiled this code yet.

To make that decision cheap, here is what the diff actually is, plus local verification at the repo's pinned toolchain.

**Vetting surface: nothing that touches CI.** Two files, both leaf modules:

```
renderer/src/template/formatters.rs
renderer/src/template/tokcfg.rs        +86 -1
```

No `.github/**`, no workflow, no `Cargo.toml`/`Cargo.lock`, no build script, no dependency change, no `deny.toml`, no CODEOWNERS. Nothing in the diff can execute at CI time outside the test binary itself.

**Local verification** on `0c83cd73`, toolchain 1.96.1 per `rust-toolchain.toml`:

| step | result |
|---|---|
| `cargo test -p dynamo-renderer` | **139 passed, 0 failed**, 1 ignored |
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --all-targets -- -D warnings` | clean |

The five new tests all pass:

```
template::tokcfg::fromjson_tests::parses_json_object_string ... ok
template::tokcfg::fromjson_tests::parses_json_array_string ... ok
template::tokcfg::fromjson_tests::passes_through_non_string ... ok
template::tokcfg::fromjson_tests::errors_on_malformed_json ... ok
template::tokcfg::fromjson_tests::renders_tool_use_block_with_json_string_arguments ... ok
```

To be clear about what this does and does not establish: a local run is not a substitute for the vetting gate, which exists so a human confirms untrusted code is safe before it runs on NVIDIA runners. It only removes the "will this even build" question so the vet is a quick read of an 86-line diff in two renderer files.

The change registers `fromjson` (plus a `from_json` alias) as a minijinja filter. Published chat templates apply it to a tool call's `arguments`, which arrive as a JSON string; minijinja has no such builtin, so the render aborts with `unknown filter: filter fromjson is unknown` and every multi-turn tool call 500s. Full production evidence is in the description.

/cc @rmccorm4 @keivenchang @ishandhanani — you have each merged here recently; whoever is closest to this, a `/ok to test 0c83cd7317d1b94fe77227d98c7c0d3645448770` would unstick it.

rmccorm4 (approved):
Updated to include recent minijinja 2.24 upgrade, LGTM - thanks for the contribution @chanh !

rmccorm4:
/ok to test [2681e1b](https://github.com/ai-dynamo/frontend-crates/pull/179/commits/2681e1b778d5689ecc3f657e45e833f6e6f267c6)

PR #183, opened 2026-08-13
fix: disable DeepSeek v4 and DeepSeek v3.2 thinking mode by default

richardhuo-nv:
## Summary

- Default auto-detected DeepSeek V4 and V3.2 native formatters to chat mode instead of thinking mode
- Add regression coverage that detected DeepSeek formatters do not enable thinking by default
- Preserve explicit per-request opt-in via `thinking=true`

### Problem to solve 
DeepSeek V4 and V3.2 were auto-selected with thinking mode enabled by default in Dynamo’s native prompt formatter. This differs from the default behavior in vLLM, SGLang, and TRT-LLM, where thinking mode is off unless explicitly enabled. The mismatch can surprise users and may reduce multi-turn KV cache hit rates by rendering additional reasoning tokens into the prompt history.

### DIFF
--- a/renderer/src/template.rs
+++ b/renderer/src/template.rs
@@ -36,13 +36,13 @@ pub fn deepseek_formatter_for(
             "Detected DeepSeek V4 model, using native Rust formatter",
         );
         return Some(PromptFormatter::OAI(Arc::new(
-            super::deepseek::v4::DeepSeekV4Formatter::new_thinking(),
+            super::deepseek::v4::DeepSeekV4Formatter::new_chat(),
         )));
     }
     if is_deepseek_v3_2_non_exp(model_type_lower, display_name_lower) {
         tracing::info!("Detected DeepSeek V3.2 model (non-Exp), using native Rust formatter");
         return Some(PromptFormatter::OAI(Arc::new(
-            super::deepseek::v32::DeepSeekV32Formatter::new_thinking(),
+            super::deepseek::v32::DeepSeekV32Formatter::new_chat(),
         )));
     }
     None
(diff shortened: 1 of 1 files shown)


renderer/src/template.rs
```diff
@@ -36,13 +36,13 @@ pub fn deepseek_formatter_for(
             "Detected DeepSeek V4 model, using native Rust formatter",
         );
         return Some(PromptFormatter::OAI(Arc::new(
-            super::deepseek::v4::DeepSeekV4Formatter::new_thinking(),
+            super::deepseek::v4::DeepSeekV4Formatter::new_chat(),
```
rmccorm4, line 39: My agent flagged in review of your [PR](https://github.com/ai-dynamo/frontend-crates/pull/183) that vllm defaults to thinking on (unlike sglang/trtllm). On a manual double check, that seems to be true:

- https://github.com/vllm-project/vllm/blob/2ac1f683f11cf0f526c896e5fa5e0e22ebc6ffa2/vllm/tokenizers/deepseek_v4.py#L34
- https://github.com/vllm-project/vllm/blob/3d204dfdaaf09d67d49c7855630ef949754e0f8f/rust/src/chat/src/renderer/deepseek_v4/encoding.rs#L133


Since the frameworks don't unanimously agree with each other, and the deepseek API itself mentions thinking on by default -- I'm inclined to leave it on by default how it currently is.

Would it make sense to go the other way and suggest a change to TRTLLM to default to thinking on matching the deepseek API: https://api-docs.deepseek.com/guides/thinking_mode for future convenience? (Keeping the deployment level setting for your needs as a workaround for now)

richardhuo-nv:
close due to this is not a good way to fix the issue.

PR #185, opened 2026-08-13
feat(parsers-v2): add Muse Glimmer unified and streaming tool-call parsers

KrishnanPrash:
## Overview

Adds Muse-Glimmer-30B parser support as a v2 `UnifiedParser`: one state machine that emits reasoning, content, and tool-call deltas in model order, plus the tool-only projection of the same scanner. The model uses ATEM tool markup and recipient-routed reasoning.

## What this adds

- `parsers/v2/src/unified/muse_glimmer.rs`: the unified parser.
- `parsers/v2/src/tool_calling/muse_glimmer.rs`: `MuseChannelScanner` (recipient-dispatched) and `MuseGlimmerToolStreamParser`, the tool-only projection of the same scanner. Registered as `muse_glimmer` in `create_tool_parser_for_family` and `create_unified_parser_for_family`.
- Muse cases in the v2 golden corpus (the 33 shared scenarios plus the reasoning and tool interleaving cases).

## Why a muse-specific scanner

`WrappedBlockScanner` dispatches on fixed marker literals. Muse decides the channel from the dynamic `to=<recipient>` header (the same `<|start|>` ... `<|message|>` opener begins reasoning, content, or a tool block), so there is no marker pair to declare. `MuseChannelScanner` owns the routing and reuses the shared primitives; one scanner drives both surfaces.

## Validation

- `cargo test --workspace` pass, clippy `-D warnings` and fmt clean, `check.sh ci` green, `check_family_coverage --family muse_glimmer` 0/0, golden corpus regenerates byte-identical.
- v2 batch equals stream at every byte split; a real trace ships as a fixture.
- The v2 tool-call output matched the reference over the 112-record e2e suite with zero marker leaks. The full unified served path (reasoning plus tools through one parser) lands with the companion Dynamo change below.

## Known limitations (documented and pinned)

- A marker split across a `<|eom|>` seam can survive one strip pass and reappear in assembled text. Requires model-invalid input (a reserved-token fragment as prose across a real control token). Pinned as a chunk-invariance exception.
- Integer arguments past `u64::MAX` format via `f64`.
- `<|start|>` inside an ATEM parameter value is read as a channel switch.

## Companion Dynamo change (separate PR)

To serve muse through v2, Dynamo consumes the unified parser (`create_unified_parser_for_family`) for muse by default, which replaces both the v1 reasoning parser and the tool jail. Dynamo's `parser_requires_special_tokens` must also list the muse aliases, or serving strips the channel framing before any parser runs (Inkling precedent, dynamo #11823).

## Versions

- `dynamo-parsers-v2` 0.2.0 to 0.2.1 (additive, on top of the #178 unified-surface refactor).


parsers/v2/src/unified/muse_glimmer.rs
```diff
@@ -0,0 +1,1004 @@
...
+//!
+//! What the unified path adds is ORDER: one machine routes reasoning, content and calls
+//! directly, so a thought between two calls stays between them instead of being hoisted
+//! ahead of the calls the way the split path does.
+
+use crate::tool_calling::muse_glimmer::{MuseChannelScanner, muse_scanner};
+use crate::tool_calling::traits::{Result, Tool};
+use crate::unified::{UnifiedParser, UnifiedParserEvent, UnifiedParserOutput};
+
+impl UnifiedParser for MuseChannelScanner {
```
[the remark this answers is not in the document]
KrishnanPrash: 9945785 - override `reset` to clear the scanner buffer, latches and tool index, so a recovered stream restarts at index 0.

parsers/v1/Cargo.toml
```diff
@@ -4,7 +4,7 @@
 [package]
 name = "dynamo-parsers"
 readme = "README.md"
-version = "8.0.0"
+version = "7.1.1"
```
indrajit96: I am confused about this change why is this downgrade needed? I think some merge with main messed it up.

parsers/v2/src/tool_calling/muse_glimmer.rs
```diff
@@ -0,0 +1,1391 @@
...
+        self.buffer.drain(..msg_pos + MESSAGE.len());
+        self.emit_text(out, &prefix);
+        self.allow_bare_header = false;
+        self.last_body_char = None;
+        match recipient.as_deref() {
+            Some(REASONING_RECIPIENT) => {
+                // Adjacent thoughts join with a newline, matching v1 and both
+                // engines' single `reasoning_text` field; a thought after a call or
+                // answer starts clean instead (the latch is cleared on emit).
+                if std::mem::take(&mut self.pending_reasoning_join) {
```
indrajit96: The newline is emitted when the to=self header arrives, before the block produces content. Empty reasoning blocks therefore add visible whitespace, despite the existing contract that empty blocks emit nothing.

  Example:
```
  to=self<|message|>a<|eom|>
  <|start|>assistant to=self<|message|><|eom|>
```
  Currently produces:
```
  reasoning("a\n")
```
  Expected:
```
  reasoning("a")
```
Suggested fix: 
Defer the newline until the first nonempty reasoning bytes arrive. Only arm the join latch when a reasoning channel actually emitted content. 
Preserve a\nb for two nonempty adjacent blocks, matching vLLM’s intended behavior.

parsers/v2/src/tool_calling/muse_glimmer.rs
```diff
@@ -0,0 +1,1391 @@
...
+                        bare_pos == Some(body_end) && start_pos != Some(body_end);
+                }
+                let body: String = self.buffer.drain(..body_end).collect();
+                if let Some(c) = body.chars().next_back() {
+                    self.last_body_char = Some(c);
+                }
+                self.buffer.drain(..term_len);
+                match self.state {
+                    State::InReasoning => {
+                        push_run(out, Kind::Reasoning, &body);
```
indrajit96: Content passes through stripped(), but reasoning bodies are emitted raw here, in the incremental path, and during finish(). 
A complete orphan marker can therefore escape in reasoning, contradicting the documented I3 invariant.

Example:
```
  to=self<|message|>a<|message|>b<|eom|>
```
  Currently produces:
```
  reasoning("a<|message|>b")
```
  Expected:
```
  reasoning("ab")
```
Suggested fix: 
Add an `emit_reasoning()` helper that applies stripped() and handles the deferred separator from comment 2. Use it for closed bodies, incremental bodies, and finish-time reasoning.

indrajit96 (commented):
LGTM!

Core parser behavior and targeted tests look solid.
Some Nits and versioning questions

parsers/v1/Cargo.toml
KrishnanPrash: ac94da2 - restore `dynamo-parsers` to main's 8.0.0; the PR no longer touches parsers/v1 at all.

parsers/v2/src/tool_calling/muse_glimmer.rs
KrishnanPrash: 106d907 - kept as-is and pinned with a test: SGLang appends the newline on the header before the body is known, and vLLM's batch parser returns "a\n" for this exact input, so deferring it would diverge from both references.

parsers/v2/src/tool_calling/muse_glimmer.rs
KrishnanPrash: 106d907 - kept as-is and pinned with a test: vLLM's channel-header regex needs a recipient, so a bare `<|message|>` never ends a body, and SGLang appends the body unstripped; stripping would make Dynamo the only implementation that does.

keivenchang (approved):
Thanks.

keivenchang:
hey Dallas, nice work on this. digging into the conformance table after approving — muse_glimmer only covers 36 of 81 unified cases. guided-decoding and prefilled-response groups get skipped since there's no `initialize_request` override yet, so the generator would panic if it tried. not asking you to fix this one, it's already documented as a known gap. can you add a tracking issue (or TODO) for the missing guided-decoding/prefilled-response coverage so it doesn't get lost?

keivenchang (changes requested):
conformance (tests) to be completed:
image

keivenchang (approved):
Just gonna pre-approve to unblock you.

KrishnanPrash:
Added the TODO for the guided-decoding/prefilled-response coverage. Thanks for the approval @keivenchang!

PR #187, opened 2026-08-13
fix(tokenizers): avoid UTF-8 boundary panic in DecodeStream

### LINKED ISSUE #186: DecodeStream panics when incremental decoding rewrites a multibyte UTF-8 suffix
## Description

While running a streaming inference benchmark through the Dynamo frontend, the benchmark repeatedly stopped making progress at the same point. The frontend log showed a panic from `dynamo_tokenizers::DecodeStream::step`:

```text
byte index 15 is not a char boundary; it is inside '𫷲'
```

A second request reproduced the same failure at byte index 37 inside `𝟙`.

## Affected versions

- `dynamo-tokenizers 1.3.2` (the version where the failure was observed)
- `dynamo-tokenizers 1.8.0` and current `main` contain the same slicing logic

## Root cause

`DecodeStream::step` computes the incremental output by slicing the newly decoded string at the previous decoded string's byte length:

```rust
let emitted = new_text[prefix_text.len()..].to_string();
```

Incremental decoding can rewrite the trailing text when another token is appended. Because Rust string indices are byte offsets, `prefix_text.len()` is not guaranteed to remain a UTF-8 character boundary in `new_text`. Slicing at that offset then panics.

This can be reproduced deterministically with a tokenizer that returns:

- `decode([1]) -> "abcde"` (5 bytes)
- `decode([1, 2]) -> "abcd😀"` (8 bytes)

The previous length, 5, lies inside the four-byte `😀` scalar, which starts at byte 4 in the new string.

`Sequence::append_token_id` already handles the same condition by moving the split point backward until it reaches a valid character boundary, but `DecodeStream::step` does not.

## Expected behavior

`DecodeStream::step` should not panic when incremental decoding rewrites a multibyte suffix. The split offset should be rounded down to a valid UTF-8 character boundary before slicing.

I have a small patch and regression tests covering the deterministic case above and the two observed boundary offsets, and will open a linked PR.

fly-go-run:
## Summary

- prevent `DecodeStream::step` from slicing a decoded string at an invalid UTF-8 byte boundary when incremental decoding rewrites the trailing text
- round the previous byte length down to a character boundary before the first generated chunk is emitted
- reject complete decodes that rewrite text already returned to the caller instead of silently emitting a corrupt delta
- add regression coverage for the two boundary offsets observed in production, same-boundary rewrites, equal/shorter rewrites, and normal append-only decoding

## Validation

- `cargo test -p dynamo-tokenizers` (75 tests passed)
- `cargo clippy -p dynamo-tokenizers --all-targets --all-features -- -D warnings`

Fixes #186

### DIFF
--- a/tokenizers/src/lib.rs
+++ b/tokenizers/src/lib.rs
@@ -401,6 +401,14 @@ pub fn create_tokenizer_from_file_with_options(
 // and https://github.com/vllm-project/vllm/blob/da2705198fa19030a25d0bea437f7be6547d47d4/vllm/transformers_utils/detokenizer_utils.py#L51
 const INITIAL_INCREMENTAL_DETOKENIZATION_OFFSET: usize = 5;
 
+fn floor_char_boundary(text: &str, requested: usize) -> usize {
+    let mut boundary = requested.min(text.len());
+    while boundary > 0 && !text.is_char_boundary(boundary) {
+        boundary -= 1;
+    }
+    boundary
+}
+
 /// DecodeStream will keep the state necessary to produce individual chunks of
 /// strings given an input stream of token_ids.
 ///
@@ -474,7 +482,13 @@ impl DecodeStream {
 
         let new_text = new_result.as_str();
         if new_text.len() > prefix_text.len() && !new_result.is_partial() {
-            let emitted = new_text[prefix_text.len()..].to_string();
+            // Decoding the same token window with one more token can rewrite the
+            // last Unicode scalar value. In that case the previous byte length
+            // may point into the middle of a multi-byte UTF-8 character in the
+            // new string. Round the split point down to a valid boundary instead
+            // of indexing the string at an invalid byte offset and panicking.
+            let prefix_text_len = floor_char_boundary(new_text, prefix_text.len());
+            let emitted = new_text[prefix_text_len..].to_string();
 
             self.prefix_offset = self.read_offset;
             self.read_offset = self.all_token_ids.len();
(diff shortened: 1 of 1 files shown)


tokenizers/src/lib.rs
```diff
@@ -401,6 +401,14 @@ pub fn create_tokenizer_from_file_with_options(
 // and https://github.com/vllm-project/vllm/blob/da2705198fa19030a25d0bea437f7be6547d47d4/vllm/transformers_utils/detokenizer_utils.py#L51
 const INITIAL_INCREMENTAL_DETOKENIZATION_OFFSET: usize = 5;
 
+fn floor_char_boundary(text: &str, requested: usize) -> usize {
```
rmccorm4: Rust 1.96.1 already provides str::floor_char_boundary; I don't think the custom helper and helper-only test are necessary here - can we use the built-in function instead?

tokenizers/src/lib.rs
```diff
@@ -474,7 +482,13 @@ impl DecodeStream {
 
         let new_text = new_result.as_str();
         if new_text.len() > prefix_text.len() && !new_result.is_partial() {
-            let emitted = new_text[prefix_text.len()..].to_string();
+            // Decoding the same token window with one more token can rewrite the
+            // last Unicode scalar value. In that case the previous byte length
+            // may point into the middle of a multi-byte UTF-8 character in the
+            // new string. Round the split point down to a valid boundary instead
+            // of indexing the string at an invalid byte offset and panicking.
+            let prefix_text_len = floor_char_boundary(new_text, prefix_text.len());
```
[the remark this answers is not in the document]
rmccorm4: Echoing this comment

Rounding down is only safe while prefix_text is un-emitted prompt context. After the first generated chunk, this condition means the tokenizer rewrote text the client has already received, so emitting from the rounded boundary silently corrupts the stream. With an empty prompt, decode([1]) = "abcde" and decode([1, 2]) = "abcd😀", this patch returns "abcde😀" instead of "abcd😀".
Please track whether output has already been emitted. Allow the rounded-boundary recovery only for initial prompt context; after emission, return a decode error unless we add a broader holdback/finalization design.

rmccorm4 (commented):
Hi @fly-go-run, thanks for the contribution! Left a couple comments

tokenizers/src/lib.rs
fly-go-run: Switched this to `str::floor_char_boundary` and removed the custom helper and its standalone test.

tokenizers/src/lib.rs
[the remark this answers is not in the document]
fly-go-run: Good catch. I added an emission-state flag so boundary recovery is only allowed before the first generated chunk is returned. After that, a rewrite returns an error instead of re-emitting text. I added regression tests for both cases.

rmccorm4:
/ok to test [7da5328](https://github.com/ai-dynamo/frontend-crates/pull/187/commits/7da532838a882feda90aed68436130654845902a)

tokenizers/src/lib.rs
```diff
@@ -474,10 +481,24 @@ impl DecodeStream {
...
         let new_text = new_result.as_str();
         if new_text.len() > prefix_text.len() && !new_result.is_partial() {
-            let emitted = new_text[prefix_text.len()..].to_string();
+            let requested_split = prefix_text.len();
+            let split = new_text.floor_char_boundary(requested_split);
+
+            // Rewinding into prompt context is safe because prompt text was not
+            // returned by this stream. Once generated text has been returned,
+            // rewinding would duplicate or corrupt text already seen by the caller.
+            if self.has_emitted && split != requested_split {
```
rmccorm4: Thanks for adding the emitted-state guard. This still only catches rewrites where the previous byte length stops being a UTF-8 boundary. A rewrite can preserve that boundary and still change text the caller already received: if `[1]` decodes to `abcde` and `[1, 2]` decodes to `abcdXY`, the stream returns `abcdeY` while the full decode is `abcdXY`.

Could we validate content continuity instead? Once `has_emitted` and `new_result` is complete, return the terminal error unless `new_text.starts_with(&prefix_text)`. That check needs to happen before `new_text.len() > prefix_text.len()` so equal-length and shorter rewrites are caught too. After emission, `prefix_text.len()` is then a safe split; `floor_char_boundary` is only needed for un-emitted prompt context. Please add a regression covering a same-boundary rewrite as well.

fly-go-run: Thanks, I updated it in 575eac7

rmccorm4 (approved):

rmccorm4:
/ok to test [575eac7](https://github.com/ai-dynamo/frontend-crates/pull/187/commits/575eac71f718979b06cadd91c9506fdb405dbe67)

rmccorm4:
Thanks for the contribution @fly-go-run !

PR #145, opened 2026-07-27
feat: add support for Kimi-K3

ynpandey-nv:
## Summary
- add Kimi-K3 prompt rendering and tokenizer support
- add Kimi-K3 reasoning and tool-calling parsers
- add coverage for Kimi-K3 structural-tag and XTML behavior

## Test plan
- [x] `git diff --check origin/main...HEAD`
- [ ] `cargo fmt --all -- --check` (Cargo is not installed in this environment)
- [ ] CI

PR #188, opened 2026-08-14
fix: Fix kimi K2 structural tag issues

indrajit96:
## Overview:

Adds native Kimi K2 structural-tag generation so guided decoding uses the same tool-call format that the Kimi K2 parser understands.

## Root cause:

Dynamo could parse Kimi K2's native, marker-delimited tool calls, but `frontend-crates` did not have a structural-tag builder for that same format. When Dynamo had to enforce a `required` or named tool call, it therefore fell back to constraining the output as a generic JSON array.

This created a format mismatch: Kimi K2 is trained to produce its native tool-call markers, while guided decoding required raw JSON and the response parser expected the native markers. Some prompts happened to complete the generic JSON and passed. With other prompts, the grammar accepted the opening `[` but rejected the model's preferred continuation. Whitespace remained valid, so generation could continue with whitespace until `max_tokens` without producing a tool call.

## Details:

- Registers a structural-tag builder for the `kimi_k2` parser.
- Builds Kimi K2's native tool-call envelope, including dynamic call IDs and request tool schemas.
- Supports `required`, named, and `auto` choices while preserving parallel-call behavior.
- Matches vLLM/xgrammar schema semantics for Kimi: an omitted or `true` per-tool `strict` value uses the declared schema, while an explicit `strict: false` permits unconstrained valid JSON unless global strict mode overrides it.
- Shares that Kimi-specific schema policy with the existing Kimi K3 builder while leaving Dynamo's generic schema policy unchanged.
- Supports K2-Instruct and K2.5/K2.6 reasoning prompts that use the `kimi_k25` `</think>` terminator. Original K2-Thinking prompts using `◁think▷` markers are explicitly outside this builder's scope.
- Adds focused coverage for tool choices, schema modes, reasoning-prefix handling, and real parser-map registration.
- Bumps `dynamo-parsers` from `7.1.1` to `8.0.0`. Adding `KimiK2` to the public `StructuralTagBuilder` enum can break downstream exhaustive matches, so this requires a major-version bump.

## Where should the reviewer start?

Start with `parsers/v1/src/tool_calling/structural_tag/kimi_k2.rs`. Then review the shared Kimi schema policy and public builder registration in `parsers/v1/src/tool_calling/structural_tag/builder.rs`, followed by the parser registration in `parsers/v1/src/tool_calling/config.rs`.

## Related Issues

**🔗 This PR is linked to an issue:**

- Relates to the Kimi forced-tool issue [ai-dynamo/dynamo#13231](https://github.com/ai-dynamo/dynamo/issues/13231); the complete behavior fix spans this and the companion Dynamo PR.

### DIFF
--- a/parsers/v1/src/tool_calling/config.rs
+++ b/parsers/v1/src/tool_calling/config.rs
@@ -853,7 +853,7 @@ impl ToolCallConfig {
         // Reference: https://huggingface.co/moonshotai/Kimi-K2-Instruct/blob/main/docs/tool_call_guidance.md
         Self {
             parser_config: ParserConfig::KimiK2(KimiK2ParserConfig::default()),
-            structural_tag_builder: None,
+            structural_tag_builder: Some(StructuralTagBuilder::KimiK2),
         }
     }
 
--- a/parsers/v1/src/tool_calling/structural_tag/builder.rs
+++ b/parsers/v1/src/tool_calling/structural_tag/builder.rs
@@ -122,7 +122,8 @@ pub enum StructuralTagBuilder {
     ///
     /// When generation starts in a prompt-injected reasoning block, this
     /// builder assumes the K2.5/K2.6 `kimi_k25` `</think>` terminator. Original
-    /// K2-Thinking, which emits its own `◁think▷` markers, is not supported.
+    /// K2-Thinking is not supported because its chat template leaves the
+    /// opening `<think>` for the model to generate.
     KimiK2,
 
     /// Kimi K3's native XTML response/tools channel format.
@@ -223,7 +224,7 @@ impl StructuralTagBuilder {
             Self::TriggeredTags(config) => config.reasoning_end.as_deref(),
             Self::DsmlToolCalls(config) => config.reasoning_end.as_deref(),
             // K2.5/K2.6 `kimi_k25` reasoning terminator. Original K2-Thinking
-            // uses different markers and must not be routed to this builder.
+            // generates the opening `<think>` itself and needs a separate path.
             Self::KimiK2 => Some("</think>"),
             Self::KimiK3 => Some("<|close|>think<|sep|>"),
         }
--- a/parsers/v1/src/tool_calling/structural_tag/builder.rs
+++ b/parsers/v1/src/tool_calling/structural_tag/builder.rs
@@ -82,7 +82,11 @@ pub(crate) fn resolve_tools_to_include<'a>(
     }
 }
 
-/// Resolve one tool's argument schema for structural tag generation.
+/// Resolve one tool's argument schema using Dynamo's generic schema policy.
+///
+/// In [`StructuralTagSchemaMode::Auto`], only `strict: true` selects the
+/// declared schema. Model-native builders may intentionally follow different
+/// upstream semantics and should use a model-specific policy helper.
 pub(crate) fn resolve_tool_schema(tool: &ToolDefinition, strict_schema: bool) -> Value {
     // xgrammar uses `true` for syntactically valid but schema-unconstrained JSON.
     let default_schema = json!(true);
@@ -95,6 +99,16 @@ pub(crate) fn resolve_tool_schema(tool: &ToolDefinition, strict_schema: bool) ->
     }
 }
 
+/// Whether Kimi's vLLM/xgrammar-compatible format should use the declared
+/// parameter schema.
+///
+/// Kimi treats only an explicit `strict: false` as an opt-out. An omitted
+/// `strict` value therefore uses the declared schema, unlike Dynamo's generic
+/// [`StructuralTagSchemaMode::Auto`] policy.
+pub(crate) fn kimi_uses_declared_tool_schema(tool: &ToolDefinition, strict_schema: bool) -> bool {
+    strict_schema || tool.strict != Some(false)
+}
+
 /// Builder for model-family-specific tool-call structural tags.
 #[derive(Debug, Clone)]
 pub enum StructuralTagBuilder {
@@ -105,6 +119,10 @@ pub enum StructuralTagBuilder {
     DsmlToolCalls(DsmlToolCallsConfig),
 
     /// Kimi K2's native special-token tool-call section format.
+    ///
+    /// When generation starts in a prompt-injected reasoning block, this
+    /// builder assumes the K2.5/K2.6 `kimi_k25` `</think>` terminator. Original
+    /// K2-Thinking, which emits its own `◁think▷` markers, is not supported.
     KimiK2,
 
     /// Kimi K3's native XTML response/tools channel format.
@@ -204,6 +222,8 @@ impl StructuralTagBuilder {
         match self {
             Self::TriggeredTags(config) => config.reasoning_end.as_deref(),
             Self::DsmlToolCalls(config) => config.reasoning_end.as_deref(),
+            // K2.5/K2.6 `kimi_k25` reasoning terminator. Original K2-Thinking
+            // uses different markers and must not be routed to this builder.
             Self::KimiK2 => Some("</think>"),
             Self::KimiK3 => Some("<|close|>think<|sep|>"),
         }
--- a/parsers/v1/src/tool_calling/structural_tag/builder.rs
+++ b/parsers/v1/src/tool_calling/structural_tag/builder.rs
@@ -12,6 +12,7 @@ use super::dsml::{self, DsmlToolCallsConfig};
 use super::format::{
     AnyTextFormat, AnyTokensFormat, Format, SequenceFormat, StructuralTag, TagFormat,
 };
+use super::kimi_k2;
 use super::kimi_k3;
 use super::triggered_tags::{self, TriggeredTagsConfig};
 
@@ -103,6 +104,9 @@ pub enum StructuralTagBuilder {
     /// DeepSeek DSML format with a `triggered_tags` wrapper and invoke list.
     DsmlToolCalls(DsmlToolCallsConfig),
 
+    /// Kimi K2's native special-token tool-call section format.
+    KimiK2,
+
     /// Kimi K3's native XTML response/tools channel format.
     KimiK3,
 }
@@ -123,6 +127,7 @@ impl StructuralTagBuilder {
         let structural_tag = match self {
             Self::TriggeredTags(config) => triggered_tags::build_triggered_tags(config, ctx)?,
             Self::DsmlToolCalls(config) => dsml::build_dsml_tool_calls(config, ctx)?,
+            Self::KimiK2 => kimi_k2::build_kimi_k2(ctx)?,
             Self::KimiK3 => kimi_k3::build_kimi_k3(ctx)?,
         };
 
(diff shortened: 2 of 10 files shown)


parsers/v1/src/tool_calling/structural_tag/kimi_k2.rs
```diff
@@ -0,0 +1,254 @@
...
+const TOOL_CALL_ARGUMENT_BEGIN: &str = "<|tool_call_argument_begin|>";
+const TOOL_CALL_END: &str = "<|tool_call_end|>";
+const TOOL_CALLS_SECTION_BEGIN: &str = "<|tool_calls_section_begin|>";
+const TOOL_CALLS_SECTION_END: &str = "<|tool_calls_section_end|>";
+
+fn tool_schema(tool: &ToolDefinition, strict_schema: bool) -> Value {
+    // Match vLLM/xgrammar: use the declared parameters unless the request
+    // explicitly opts out with strict=false. Global strict mode overrides the
+    // opt-out. Xgrammar uses `true` for unconstrained but valid JSON.
+    if strict_schema || tool.strict != Some(false) {
```
keivenchang: This looks backwards at first, but I think it needs fixing before publish. It's a verbatim copy of `kimi_k3.rs:198-205`, and it disagrees with the shared parent: `builder.rs:747 resolve_tool_schema` uses `strict_schema || tool.strict.unwrap_or(false)` while this uses `strict_schema || tool.strict != Some(false)`. They diverge on the common `strict: None` case.

xgrammar and vLLM say this copy is right: `strict == false` means unconstrained JSON; absent or true means the declared parameter schema. So the PARENT is the divergent one. Routing K2 through `resolve_tool_schema` today would make K2 wrong.

You can either fix the parent to match upstream and route K2/K3 through it, or document there why Dynamo intentionally differs. Please don't leave a third copy with no note saying which one is authoritative.

parsers/v1/src/tool_calling/structural_tag/kimi_k2.rs
```diff
@@ -0,0 +1,254 @@
...
+            triggers: vec![TOOL_CALLS_SECTION_BEGIN.to_string()],
+            tags: vec![TagFormat {
+                begin: TOOL_CALLS_SECTION_BEGIN.to_string(),
+                content: Box::new(calls_format),
+                end: TOOL_CALLS_SECTION_END.to_string(),
+            }],
+            at_least_one: outer_at_least_one,
+            stop_after_first: ctx.stop_after_first(),
+        })
+    } else {
```
keivenchang, line 105: I'm not sure K2-Thinking is supported by this builder, but we need to decide that. This branch forces `<|tool_calls_section_begin|>` at generation position 0, which is right for K2-Instruct. K2-Thinking doesn't prefill `<think>` in its generation prompt, though, and Dynamo gates the reasoning prefix only on a prompt-injected opener. The grammar can demand the section marker before the model emits its own `<think>`.

If K2-Thinking is supported, add an exact rendered-prompt test for required/named. Otherwise narrow the module doc so nobody wires it up later.

parsers/v1/src/tool_calling/structural_tag/builder.rs
```diff
@@ -198,6 +204,7 @@ impl StructuralTagBuilder {
         match self {
             Self::TriggeredTags(config) => config.reasoning_end.as_deref(),
             Self::DsmlToolCalls(config) => config.reasoning_end.as_deref(),
+            Self::KimiK2 => Some("</think>"),
```
keivenchang, line 229: Small one, and not a live bug. I first thought hardcoding `</think>` broke `tool_parser=kimi_k2` + `reasoning_parser=kimi`, because that parser uses `◁think▷`/`◁/think▷`. It doesn't: `prompt_injected_reasoning_start` doesn't recognize that opener, so `starts_in_reasoning` is false and this wrapper never gets installed. `</think>` also matches K2.5/K2.6 and current xgrammar.

The structural bit stands out. `TriggeredTagsConfig` and `DsmlToolCallsConfig` carry `reasoning_end`; this unit variant has nowhere to put it, which also explains the `&[]` ban tokens below. Fine to defer, but one line saying it assumes the `kimi_k25`-style terminator would've saved me the trip.

parsers/v1/src/tool_calling/structural_tag/builder.rs
```diff
@@ -103,6 +104,9 @@ pub enum StructuralTagBuilder {
     /// DeepSeek DSML format with a `triggered_tags` wrapper and invoke list.
     DsmlToolCalls(DsmlToolCallsConfig),
 
+    /// Kimi K2's native special-token tool-call section format.
+    KimiK2,
```
keivenchang, line 128: On the bot's semver comment: I went back and forth. `KimiK3` landed the same way in `d08ae6e9` (#145) without a crate version bump, so there's precedent. But this enum is public and exhaustive, and a new variant breaks downstream crates that match every variant. I think we should explicitly decide whether this ships as another compatible 7.x instead of letting it ride.

parsers/v1/src/tool_calling/config.rs
```diff
@@ -853,7 +853,7 @@ impl ToolCallConfig {
         // Reference: https://huggingface.co/moonshotai/Kimi-K2-Instruct/blob/main/docs/tool_call_guidance.md
         Self {
             parser_config: ParserConfig::KimiK2(KimiK2ParserConfig::default()),
-            structural_tag_builder: None,
+            structural_tag_builder: Some(StructuralTagBuilder::KimiK2),
```
keivenchang, line 856: BTW, `structural_tag/tests.rs:601` — `parser_map_structural_tag_smoke` walks the real parser map and checks that the builder is wired, but `"kimi_k2"` isn't in its name list. This registration line has no test.

parsers/v1/src/tool_calling/structural_tag/kimi_k2.rs
```diff
@@ -0,0 +1,254 @@
...
+        return Ok(None);
+    }
+
+    let mut calls: Vec<TagFormat> = tools
+        .into_iter()
+        .map(|tool| call_tag(tool, ctx.strict_schema()))
+        .collect();
+
+    let calls_format = if matches!(ctx.tool_choice, ToolChoice::Named(_)) {
+        Format::Tag(calls.pop().expect("named tool choice resolves one tool"))
```
keivenchang: nit: `expect` depends on an invariant owned by another module; it holds today because `builder.rs:734-742` returns exactly one tool or an error for `Named`, but `kimi_k3.rs` avoids it and `calls.into_iter().next().ok_or_else(...)` is the same length.

keivenchang (commented):
hey Indrajit, thanks for this, just reviewed it.

Hey also, I want to raise a concern, FYI, this doesn't bump `parsers/v1/Cargo.toml`, so merging won't publish it. ai-dynamo/dynamo#13235 stays red until someone cuts a release and moves its lockfile off `dynamo-parsers` 7.0.1. Are you doing that, or is someone else?

parsers/v1/src/tool_calling/structural_tag/kimi_k2.rs
indrajit96: Fixed. Good point. Makes sense.
I documented the intentional difference and added a shared func `kimi_uses_declared_tool_schema` that follows vLLM/xgrammar semantics. Both K2 and K3 now use this helper, so there is no third copy.
Also added tests covering strict as omitted, true, and false, including the global strict-mode override.

parsers/v1/src/tool_calling/structural_tag/kimi_k2.rs
indrajit96: Fixed. 
Good point. I checked the official model templates and decided not to claim K2-Thinking support here.
This builder supports K2-Instruct and K2.5/K2.6. 
K2-Thinking generates its own <think> opener, while K2.5/K2.6 inject <think> in the prompt. 
Because Dynamo only marks reasoning as active for a prompt-injected opener, this grammar could force the tool-call marker too early for K2-Thinking.

I narrowed the documentation accordingly. Supporting K2-Thinking would need a separate grammar path and exact rendered-prompt tests.

parsers/v1/src/tool_calling/structural_tag/builder.rs
indrajit96: Agreed. 
I kept KimiK2 as a simple builder option with no additional configuration and documented that it assumes the K2.5/K2.6 kimi_k25-style </think> terminator. Original K2-Thinking generates its own <think> opener and would require a separate path.

parsers/v1/src/tool_calling/structural_tag/builder.rs
indrajit96: I bumped dynamo-parsers from 7.1.1 to 8.0.0 and updated the workspace dependency, lockfile, and changelog.

parsers/v1/src/tool_calling/config.rs
indrajit96: Fixed.
Good catch. I added kimi_k2 to parser_map_structural_tag_smoke, so the test now verifies through the real parser map that kimi_k2 is registered with a structural-tag builder.
```
The focused structural-tag suite passes: 47 tests, 0 failures.
```

parsers/v1/src/tool_calling/structural_tag/kimi_k2.rs
indrajit96: Fixed.
Agreed. I replaced `pop().expect(...)` with `into_iter().next().ok_or_else(...)`, matching the Kimi K3 approach.

indrajit96:
> hey Indrajit, thanks for this, just reviewed it.
> 
> Hey also, I want to raise a concern, FYI, this doesn't bump `parsers/v1/Cargo.toml`, so merging won't publish it. [ai-dynamo/dynamo#13235](https://github.com/ai-dynamo/dynamo/pull/13235) stays red until someone cuts a release and moves its lockfile off `dynamo-parsers` 7.0.1. Are you doing that, or is someone else?

@keivenchang 
Thanks for catching this I did not know. I will handle the release and downstream update.

I bumped dynamo-parsers from 7.1.1 to 8.0.0, including the workspace dependency, lockfile, and changelog. 
I chose a major bump because adding a variant to the public exhaustive enum can break downstream matches.

Once this PR merges and 8.0.0 is published automatically right? 
I’ll update Dynamo’s dependency and lockfile in the companion PR so #13235 is no longer pinned to 7.0.1.

keivenchang (approved):
went through the new head — all six addressed, and folding the schema rule into `kimi_uses_declared_tool_schema` so K2 and K3 share one parent is exactly right. thanks for the quick turnaround Indrajit.

FYI two non-blockers for whenever: the 8.0.0 CHANGELOG entry sits under Features with no breaking marker, and `#[non_exhaustive]` on `StructuralTagBuilder` would make the next model family a minor bump instead of another major.

parsers/v1/src/tool_calling/structural_tag/builder.rs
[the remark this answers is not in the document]
indrajit96: Fixed

PR #189, opened 2026-08-17
feat(parsers-v2): add structural tag support

jellysnack:
## Summary

This PR adds XGrammar structural-tag support to parsers v2.

It includes:

- structural-tag builders for `qwen3_coder`, `deepseek_v4`, and `glm47`
- support for OpenAI tool-choice and parallel tool-call semantics
- configurable tool schema, reasoning, and special-token constraints
- tool calling combined with structured output
- a shared interface for adding structural tags to additional parser families

## Motivation

Structural tags are supported by parsers v1 but are not yet available with parsers v2. This PR adds that support.

## Differences from v1

The v2 implementation uses v1 as a behavioral baseline while separating request policy from model-specific formatting and adding several capabilities.

| Area | v1 | v2 |
|---|---|---|
| Orchestration | Tool-call formats and `tool_choice=none` bans are built through separate methods | A single builder resolves the request and applies tool-choice, reasoning, and exclusion policy |
| Model builders | Individual builders interpret parts of the request policy themselves | Tool selection and cardinality are normalized before reaching the model-specific grammar |
| Named choice | Restricts generation to the selected tool, while cardinality still follows `parallel_tool_calls` | Produces exactly one call to the selected tool, matching the OpenAI named `tool_choice` contract |
| Reasoning | Prompt-opened reasoning is wrapped only for required and named choices | Shared handling applies to every enabled tool-calling mode. Alternatively, the inference engine can wait for the reasoning close marker and activate only the post-reasoning grammar |
| Special markers | `tool_choice=none` primarily uses token-level bans | String exclusions prevent tool-call strings regardless of tokenization and can also keep tool calls out of reasoning and reasoning markers out of free text |
| Tools + structured output | Not supported | `auto` can choose between model-native tool calls and a JSON Schema-constrained final response |
| Argument ordering | Declared JSON Schema order | Optional `any_order`, disabled by default because it weakens required-property and duplicate-key validation |
| Extensibility | Model formats are represented by separate builder variants and configuration types | Common flat formats share `TemplatedToolCallFormat`, complex formats implement `ToolCallGrammar` directly |

Request semantics are represented by `StructuralTagContext` and resolved into a model-independent policy before reaching the model-specific grammar.

`StructuralTagOptions` contains serving-time behavior settings configured by the integration and reused across requests, including:

- special-token exclusions
- structural-tag-owned or inference-engine-owned reasoning boundaries
- tool argument property ordering.

The XGrammar wire representation remains private, so it can evolve without becoming part of the public parsers v2 API.

For `auto` requests with both tools and a structured-output schema, the builder produces an alternative between a calls-only model-native tool-call format and a JSON Schema-constrained final response. This allows the model to call a tool first and produce the structured final response after receiving the tool result.

## Supported model formats

- Qwen3 Coder tool-call XML
- DeepSeek V4 DSML tool-call blocks
- GLM 4.7-style XML tool calls

The parser-family registry now declares both the v2 parser factory and its optional structural-tag capability, keeping model-family routing in one place.

## Validation

The structural-tag builders were validated end to end with downstream Dynamo changes that integrate parsers v2 structural tags into request preprocessing. The Dynamo integration will be submitted as a separate PR after this frontend-crates change is merged. 

Parser compatibility improvements discovered during end-to-end validation will be submitted separately.


parsers/v2/src/structural_tag/builders/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    /// Builds a structural tag with request-independent behavior overrides.
+    pub fn build_with_options(
+        &self,
+        context: &StructuralTagContext<'_>,
+        options: &StructuralTagOptions,
+    ) -> anyhow::Result<Option<Value>> {
+        let policy = match resolve_tool_calling(context)? {
+            ResolvedToolCalling::Disabled => {
```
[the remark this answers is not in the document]
jellysnack: This is intentional. The builder only owns structural tag constraints involving tool calls. `structured_output_schema` is provided solely to build the structured tag when both tools and structured output are present. With an empty tools list, no tool-call structural tag is needed, so the builder returns `None` and the caller retains its standalone json schema constraint

rmccorm4:
Thanks for the contribution @jellysnack ! Tagging @keivenchang @indrajit96 to help review

parsers/v2/src/tool_calling/qwen3_coder.rs
```diff
@@ -35,7 +37,13 @@ const PARAMETER_START: &str = "<parameter=";
 fn spec() -> WrappedBlockSpec {
     WrappedBlockSpec {
         family: "qwen3_coder",
-        block_starts: vec![BLOCK_START.to_string()],
+        // Qwen uses "\n\n" before the first call and "\n" between calls.
+        // Consume them as part of the opener, but also accept the bare form.
+        block_starts: vec![
+            FIRST_BLOCK_START.to_string(),
+            SEPARATED_BLOCK_START.to_string(),
```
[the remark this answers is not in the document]
jellysnack: The parser fix requires a larger diff, so I removed the parser changes from this PR and will submit them separately

jellysnack:
> **parsers/v2/src/structural_tag/builders/mod.rs:82** — `RequiredToolChoiceFormat::TextAllowed` still builds `triggered_tags` with a required policy, which serializes `at_least_one: true` and disallows the leading free text this option is meant to permit. Fix: build the text-allowed required case as leading `any_text` followed by a required triggered tag, or otherwise pass a non-required triggered phase that actually allows leading text.
> 
> 🤖 AI Fix

The `TextAllowed` option was misleading so I removed `RequiredToolChoiceFormat` entirely. Required requests now consistently use `TriggeredTags` and start with a tool call including the model-specific prefix where needed

indrajit96:
Thanks @jellysnack for this taking a look right now

rmccorm4:
/ok to test [35d583c](https://github.com/ai-dynamo/frontend-crates/pull/189/commits/35d583c73e7c5ea2a783d2c5b5b990238ed7dbb7)

parsers/v2/src/structural_tag/builders/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+    tags: Vec<TagFormat>,
+    reasoning_begin: Option<&str>,
+    reasoning_end: Option<&str>,
+    exclude_special_tokens: bool,
+    policy: &ResolvedToolCallingPolicy<'_>,
+) -> Format {
+    Format::TriggeredTags(TriggeredTagsFormat {
+        triggers: vec![trigger.to_string()],
+        tags,
+        excludes: if exclude_special_tokens {
```
indrajit96, line 183: Exclude GLM’s non-trigger control markers from free text

`triggered_calls_format()` currently populates `TriggeredTagsFormat.excludes` with only the reasoning markers:

  ```rust
  ["<think>", "</think>"]
  ```

  For GLM, `</tool_call>`, `<arg_key>`, `</arg_key>`, `<arg_value>`, and `</arg_value>` are also reserved structural tokens. Because they are not excluded here, the model can emit them inside an ordinary free-text response. XGrammar may then
  strip them or interpret them as malformed/bare tool-call syntax rather than literal text.

  The opening `<tool_call>` marker must remain allowed because it is the trigger that transitions from free text into a valid tool call. All other GLM control markers should be excluded.

  For example, this should not be accepted as GLM free text:

  ```text
  The tool returned </tool_call> unexpectedly.
  ```

  Today, `</tool_call>` is permitted because the generated `triggered_tags` format only excludes `<think>` and `</think>`.

  I verified this with the following regression test, added after `auto_builds_glm_triggered_dispatch` in `parsers/v2/src/structural_tag/builders/glm47.rs` (after line 80):

  ```rust
  #[test]
  fn auto_excludes_non_trigger_glm_markers_from_free_text() {
      let tools = tools();
      let actual = build(
          &GLM47,
          StructuralTagToolChoice::Auto,
          &tools,
          None,
          StructuralTagSchemaMode::Auto,
          false,
      );

      let excludes = actual["format"]["excludes"]
          .as_array()
          .expect("triggered_tags should contain an excludes array");

      for marker in [
          THINK_BEGIN,
          THINK_END,
          "</tool_call>",
          "<arg_key>",
          "</arg_key>",
          "<arg_value>",
          "</arg_value>",
      ] {
          assert!(
              excludes
                  .iter()
                  .any(|value| value.as_str() == Some(marker)),
              "GLM free text must exclude reserved marker {marker:?}; got {excludes:?}"
          );
      }

      assert!(
          excludes
              .iter()
              .all(|value| value.as_str() != Some("<tool_call>")),
          "the opening marker must remain available as the tool-call trigger"
      );
  }
  ```

  This test currently fails with:

  ```text
  GLM free text must exclude reserved marker "</tool_call>";
  got [String("<think>"), String("</think>")]
  ```

  Suggested fix: allow `triggered_calls_format()` to receive grammar-specific free-text exclusions in addition to `reasoning_begin` and `reasoning_end`. For `GLM47_FORMAT`, pass:

  ```rust
  &[
      TOOL_CALL_END,
      "<arg_key>",
      "</arg_key>",
      "<arg_value>",
      "</arg_value>",
  ]
  ```

  Do not include `TOOL_CALL_BEGIN`, since `<tool_call>` must remain available as the trigger.

  This matches XGrammar’s GLM implementation, whose `TEXT_EXCLUDES` contains the thinking markers plus every non-trigger GLM control marker:
  https://github.com/mlc-ai/xgrammar/blob/a8790f57/python/xgrammar/builtin_structural_tag.py#L1886-L1895

parsers/v2/src/structural_tag/builders/qwen3_coder.rs
```diff
@@ -0,0 +1,23 @@
...
+
+pub(crate) static QWEN3_CODER: StructuralTagBuilder =
+    StructuralTagBuilder::new(&QWEN3_CODER_FORMAT);
+
+static QWEN3_CODER_FORMAT: TemplatedToolCallFormat = TemplatedToolCallFormat {
+    tool_call_begin_prefix: "<tool_call>\n<function=",
+    tool_call_begin_suffix: ">\n",
+    tool_call_end: "\n</function>\n</tool_call>",
+    tool_call_trigger: "<tool_call>\n<function=",
+    first_tool_call_prefix: "\n\n",
```
indrajit96: Apply Qwen’s `\n\n` suffix only after a reasoning boundary

  `first_tool_call_prefix` is currently configured as `"\n\n"`. The shared template emits this prefix for every `required` or named tool call, regardless of whether generation starts inside reasoning.

  For Qwen, however, these newlines are `THINK_SUFFIX`: they belong specifically after `</think>`. They are not an unconditional prefix of a tool call.

  This causes two inconsistent behaviors:

  1. With `starts_in_reasoning = false`, `required` and named calls incorrectly begin with two newlines:

  ```text
  \n\n<tool_call>
  <function=get_weather>
  ```

  They should begin directly with:

  ```text
  <tool_call>
  <function=get_weather>
  ```

  2. With `starts_in_reasoning = true` and `tool_choice = auto`, the two newlines are not required at all because the auto path does not use `first_tool_call_prefix`. This allows:

  ```text
  </think><tool_call>
  ```

  instead of the model’s expected:

  ```text
  </think>

  <tool_call>
  ```

  I verified the first case with this regression test, added after `required_starts_with_a_triggered_tool_call` in `parsers/v2/src/structural_tag/builders/templated_tool_calls.rs`:

  ```rust
  #[test]
  fn qwen_required_without_reasoning_starts_at_tool_call_trigger() {
      let tools = tools();
      let actual = build(
          &QWEN3_CODER,
          StructuralTagToolChoice::Required,
          &tools,
          None,
          StructuralTagSchemaMode::Auto,
          false,
      );

      assert_eq!(
          actual["format"]["type"],
          "triggered_tags",
          "without an open reasoning block, Qwen should not emit leading newlines"
      );
      assert_eq!(
          actual["format"]["triggers"],
          json!(["<tool_call>\n<function="])
      );
  }
  ```

  This test currently fails with:

  ```text
  assertion `left == right` failed:
  without an open reasoning block, Qwen should not emit leading newlines

  left: String("sequence")
  right: "triggered_tags"
  ```

  The returned `sequence` contains a leading `ConstStringFormat("\n\n")`, demonstrating that the prefix is emitted even though `starts_in_reasoning` is false.

  Suggested fix:

  - Change `first_tool_call_prefix` on this line to `""`.
  - Add a model-specific `reasoning_suffix` to `ToolCallGrammar`/`TemplatedToolCallFormat`.
  - Configure Qwen’s `reasoning_suffix` as `"\n\n"`.
  - Emit that suffix in `wrap_reasoning_if_needed()` immediately after the reasoning tag and before the tool/text suffix.

  Conceptually:

  ```rust
  first_tool_call_prefix: "",
  reasoning_suffix: "\n\n",
  ```

  and in `wrap_reasoning_if_needed()` around `builders/mod.rs:232`:

  ```rust
  let mut elements = vec![
      reasoning_tag(grammar, reasoning_end, exclude_special_tokens),
  ];

  let reasoning_suffix = grammar.reasoning_suffix();
  if !reasoning_suffix.is_empty() {
      elements.push(Format::ConstString(ConstStringFormat {
          value: reasoning_suffix.to_string(),
      }));
  }

  elements.push(suffix);
  ```

  This makes the newlines conditional on `starts_in_reasoning` and applies them consistently to `auto`, `required`, and named choices.

  XGrammar models this distinction explicitly: without reasoning it returns `suffix_tag` directly; with reasoning it emits `</think>`, then `THINK_SUFFIX`, then `suffix_tag`:
  https://github.com/mlc-ai/xgrammar/blob/a8790f57/python/xgrammar/builtin_structural_tag.py#L1291-L1300

indrajit96 (commented):
Hi @jellysnack thanks a lot for adding this!

@keivenchang Adding you for desing review.
PR looks good to me and mostly carries the V1 structural-tag architecture into V2 with stronger encapsulation.

Belwo are few deviations I found from V1 to V2 and my notes

| Design choice | V1 | V2 in this PR | Review note |
| --- | --- | --- | --- |
| Parser and builder association | `ToolCallConfig` contains both the parser configuration and optional structural-tag builder. | `FamilySpec` associates the parser factory with an optional static builder. | Appropriate for V2's factory-based parser creation. |
| Public API and extensibility | V1 exposes constructible builder configurations and XGrammar format types. | V2 exposes semantic request types while keeping builders and XGrammar wire types private. | Better encapsulation, but external consumers cannot construct or register V2 builders. |
| Policy resolution | Tool selection, schema behavior, and cardinality are handled within builder paths. | `ResolvedToolCallingPolicy` centralizes selected tools, mode, cardinality, and schema mode. | Good separation and easier cross-family consistency. |
| `tool_choice = none` | Uses a separate `build_tool_call_ban()` path with token-level exclusions. | The main `build()` path returns an `AnyText` format with text-pattern exclusions. | Meaningful semantic change that should be confirmed against backend behavior. |
| Structured output | No equivalent generic alternation between native tool calls and a structured final response. | `auto` can produce an `Or` between native tool calls and a JSON-schema response. | New V2 behavior; confirm that Dynamo requires this public feature. |
| Reasoning boundary | Mainly wraps prompt-opened reasoning for required and named calls. | Supports wrapping auto, required, named, and structured-output paths, plus `ReasoningBoundary::External`. | Intentional expansion beyond V1; ownership of the boundary should be documented. |
| Deployment integration | Dynamo obtains the builder from the selected V1 `ToolCallConfig`. | Dynamo must call `structural_tag_builder_for_family(family)` after selecting V2. | A Dynamo integration change is required; this PR only exposes the crates API. |
| Reference validation | Tests primarily cover Dynamo's existing builder behavior. | Tests assert locally generated XGrammar JSON. | Consider parity tests against pinned XGrammar reference builders to avoid encoding incorrect assumptions. |

@jellysnack can you open a corresponding PR on Dynamo showing the integragration?

Internally I can pick both the PRs create a custom build and run CI along with some custom 2e2 tests.

jellysnack:
> @jellysnack can you open a corresponding PR on Dynamo showing the integragration?

Sure. And good catches on the review comments, I’ll address those and open a draft Dynamo integration PR shortly

jellysnack:
@indrajit96 Hi!

I’ve addressed both issues and also added a small GLM parser fix after E2E testing showed that the model could emit whitespace between XML elements when using structural tag

Here’s also the draft PR with the corresponding Dynamo integration: https://github.com/ai-dynamo/dynamo/pull/13903

parsers/v2/src/structural_tag/builders/mod.rs
rmccorm4, line 1: Qwen and DeepSeek still have the same bare-call hole that was fixed for GLM. XGrammar accepts <function=...> or <｜DSML｜invoke...> as normal text, but the parsers later recover it as a tool call. That lets calls bypass the schema and parallel_tool_calls=false. Does it make sense to exclude those bare-call markers from auto-mode free text?

rmccorm4 (approved):
LGTM aside from 1 comment

parsers/v2/src/structural_tag/builders/mod.rs
jellysnack: done

jellysnack:
Hi @rmccorm4 @keivenchang, just a friendly ping on this PR

indrajit96:
/ok to test 
e2994f1

indrajit96 (approved):
LGTM!
All comments addressed!

indrajit96:
/ok to test [e2994f1](https://github.com/ai-dynamo/frontend-crates/pull/189/commits/e2994f17badb9a479be786bde29e6cf58e9f0bf1)

parsers/v2/src/structural_tag/builders/mod.rs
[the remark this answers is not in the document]
rmccorm4: Confirmed on current head: no code change is needed. The builder only composes structured output with a tool-call structural tag when tools are present. In the downstream integration ([ai-dynamo/dynamo#13903](https://github.com/ai-dynamo/dynamo/pull/13903/files#diff-a696efce9788816ffc82292ea767d8f840a1277a7e77d3f41836910edc9e2b3b)), `auto` plus structured output with no tools leaves the existing response-format guided-JSON constraint as the sole assistant-output constraint. This is by design.

parsers/v2/src/structural_tag/builders/mod.rs
rmccorm4: Fixed in [`0e7c28f`](https://github.com/ai-dynamo/frontend-crates/commit/0e7c28fdb686e65abcf4102e37b85459748705da): GLM now supplies `FREE_TEXT_EXCLUDES` with `</tool_call>` and all argument markers, while leaving `<tool_call>` available as the trigger. The new `auto_excludes_non_trigger_markers_from_free_text` regression test asserts the complete exclusion set.

parsers/v2/src/structural_tag/builders/qwen3_coder.rs
rmccorm4: Fixed in [`0e7c28f`](https://github.com/ai-dynamo/frontend-crates/commit/0e7c28fdb686e65abcf4102e37b85459748705da): Qwen now uses an empty `first_tool_call_prefix` and `reasoning_suffix: "\n\n"`. The shared wrapper emits that suffix only after a structural-tag-owned reasoning close, covering auto, required, and named tool choices without adding leading newlines when generation starts outside reasoning.

parsers/v2/src/tool_calling/v1core/xml/glm47_parser.rs
```diff
@@ -487,7 +487,7 @@ fn parse_tool_call_block(
     let arg_value_start_escaped = regex::escape(&config.arg_value_start);
     let arg_value_end_escaped = regex::escape(&config.arg_value_end);
 
-    // Pattern: <arg_key>key</arg_key><arg_value>value(</arg_value> | end-of-block)
+    // Pattern: <arg_key>key</arg_key> whitespace* <arg_value>value(</arg_value> | end-of-block)
```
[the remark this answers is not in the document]
rmccorm4: nit

PR #193, opened 2026-08-18
feat(parsers): stream guided tool-call fragments, and the contract cost of doing so

keivenchang:
#### Overview:

A downstream user reported **~10s to first tool-call frame** with `tool_choice="required"` and streaming on; their vanilla-vLLM deployment starts streaming in ~0.2s. Final output is correct — valid `tool_calls`, `finish_reason="tool_calls"` — so this is a **time-to-first-frame defect, not a correctness defect**.

**Draft on purpose.** This opens a design decision rather than closing it — see *What is red*.

#### The cause is not qwen3-specific:

Every v2 family emits at call-completion granularity. Native buffers to `invoke_end` then pushes one `ToolCallDelta`; guided emits when the JSON value completes, and its own test asserts no `ToolCall` event exists before completion. No fragment-level machinery exists anywhere in `parsers/v2`.

#### What works:

Policy **A**, "commit point" — nothing is emitted until the call's `"name"` value has **closed**, at which point the name can no longer change. Argument bytes then stream as fragments, bounded to the arguments value's own span, cut on character boundaries.

Two acceptance tests, both asserting **ordering, not wall-clock**:

| test | before | after |
|---|---|---|
| `required_guided_emits_the_name_before_the_call_completes` | **RED** (`got []`) | green |
| `required_guided_streams_arguments_in_multiple_fragments` | — | green |

The second is the one that matters. It feeds the payload **one character at a time** and asserts exactly **one** name delta, **≥2** argument frames, and that fragments reassemble. The first alone would pass on an implementation that emitted the name early and still dumped every argument byte in one delta — which is most of the reported defect.

#### What is red, and why that is the deliverable:

**12 tests fail.** They encode behaviour early commitment cannot preserve: a non-object argument payload must **void** the call, a parameterless call must not void its siblings, chunk-boundary independence. All of it assumes the parser can still change its mind about a call it has not emitted. **A fragment on the wire cannot be unsaid.**

Gating already removed 13 of the original 25:

- `InvalidGuidedPayloadPolicy::Reject` never streams — that policy's typed error is only keepable while nothing has been emitted.
- Named choice never streams — its payload is *bare arguments* with no wrapper to locate a span in. Emitting a name there produced calls with `{}` arguments, worse than the latency it was fixing.

The remaining 12 need a **decision, not a patch**:

1. Move the commit point past validation — preserves all 12, loses most of the latency win.
2. Accept a weaker contract on the streaming path — keeps the win, rewrites 12 tests.

#### Not in scope:

Native markup, the other five families, EOF/truncation semantics, and the golden corpus are untouched and follow the decision above.

Separately: `tool_choice="required"` may never reach this code at all — it routes to the v1 jail today. Tracked in DIS-2708; without it the reporter's configuration is unchanged by this PR.

Closes [DIS-2707](https://linear.app/nvidia/issue/DIS-2707)

### DIFF
--- a/parsers/v2/src/unified/guided_cursor.rs
+++ b/parsers/v2/src/unified/guided_cursor.rs
@@ -1,0 +1,760 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Incremental lexer over one guided `required` tool-call payload.
+//!
+//! # Why this is one type
+//!
+//! Streaming a call before its payload has closed needs four facts that only a
+//! JSON lexer can answer: which call element the scan is inside, that element's
+//! decoded `name`, where its argument OBJECT begins, and which bytes of that
+//! object are safe to release. An earlier revision answered them with three
+//! independent `str::find` helpers, and each one was wrong in its own way —
+//! `"name"` was matched anywhere in the payload (so `{"x":"name","arguments":…}`
+//! emitted a call named `arguments`), `_` decoded to the literal characters
+//! `u005f`, the `parameters` alias was not recognised at all, and every helper
+//! rescanned the whole accumulated buffer on every push, making a
+//! character-at-a-time stream quadratic.
+//!
+//! Those are not four bugs, they are one: lexical state with no owner. This type
+//! is that owner. It holds the string/escape state, the brace depth, the current
+//! element index, the decoded name, and the released-byte offset in one place, and
+//! it advances strictly FORWARD — each call consumes only the bytes appended since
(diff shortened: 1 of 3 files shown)


keivenchang:
Folded into #190.

Both changes reshape `parsers/v2/src/unified/mod.rs` (+171/−59 there, +274/−41 here), and both add a field to `GuidedState`. Reviewing them as separate PRs means reading that file twice and rebasing whichever lands second.

The commit moved across unchanged as `d9f094ea`. CI on this PR's final head was green (6 pass, 1 skip), and the combined tree on #190 is 1495 workspace tests passing, 80/80 unified golden, clippy and fmt clean.

DIS-2707 stays open and is now tracked on #190 alongside DIS-2701.

PR #194, opened 2026-08-19
feat(parsers): stream guided tool calls in v1 and v2

keivenchang:
## Overview

Guided tool calls now stream as they arrive, in BOTH parser generations. Under `tool_choice=required` or a named tool we already pinned the output to a known JSON shape, but both crates sat on the whole call and emitted one delta at the end. The client didn't learn the function name until the entire call was already generated.

## Details

v1 and v2 take different paths, so the work wasn't the same.

**v1 (`dynamo-parsers`)** goes into the jail, and the jail basically says, "I am never going to emit anything till the very end." So v1 needed a completely new guided-decoding lexer inside the jail.

The lexer walks the list item by item. For each item it buffers just the header — the tool name and the `{` that opens the arguments — then streams the argument bytes as they arrive:

```
buffered:   [{"name":"get_weather","parameters":{
streamed:                                        "city"  :"Par  is"}
```

That header wait is unavoidable. Until the name closes, `"get_weather"` could still turn out to be some other key's value, and you can't take back a delta you already sent. In the live run one tool call came out as 169 deltas.

This is opt-in through `JailedStreamBuilder::guided_streaming`, because turning it on changes the emission shape for every consumer of the jail. The published `apply_tool_calling_jail` signature stays the same, and all existing Immediate-mode tests pass unmodified.

Before and after PR #194, inside the jail. Every box is the same except the two in green. Watch the arrows into CLIENT.

**BEFORE PR #194** — jail `Immediate` mode. One arrow out, at the very end:

```mermaid
flowchart TD
    A[chunk<br/>arrives] --> B[append<br/>to buffer]
    B --> C{payload<br/>closed?}
    C -- no, loop<br/>silently --> A
    C -- yes --> E[parse<br/>whole buffer]
    E --> F[emit<br/>tool call]
    F --> OUT([CLIENT])
```

**AFTER PR #194** — `guided_streaming` enabled. Still buffers, but no longer silent: the cursor sends on EVERY loop:

```mermaid
flowchart TD
    A[chunk<br/>arrives] --> B[append<br/>to buffer]
    B --> N1[guided cursor:<br/>emit name,<br/>then arg bytes]
    N1 --> C{payload<br/>closed?}
    N1 -.->|every chunk| OUT([CLIENT])
    C -- no --> A
    C -- yes --> E[parse<br/>whole buffer]
    E --> N2[subtract<br/>what was<br/>already streamed]
    N2 --> F[emit<br/>remainder]
    F --> OUT
    classDef added fill:#c8f7c5,stroke:#22863a,stroke-width:3px,color:#111
    class N1,N2 added
```

The jail still accumulates exactly like before, and it still has to: the completion check needs the whole buffer to see the payload close, the cursor re-reads it each pass, and if the cursor never commits (backend ignored the grammar) that buffer is what the native fallback parses. What changed is that accumulating is no longer SILENT.

**v2 (`dynamo-parsers-v2`)** already had the lexer. `required` streamed after #190, but a named choice got refused, and lifting that refusal is the whole v2 half of this PR.

The lexer looked for that header — a `name`, then the `{`. A named payload has NONE of it, because we already told the model which tool to call, so it just writes the arguments:

```
required:  [{"name":"get_weather","parameters":{"city":"Paris"}}]
named:     {"city":"Paris"}
```

So it never found a header, never started, and the call sat buffered till the end. Worse, arguments carrying their own `name` field could get that value mistaken for the tool name.

So, named mode skips the hunt: the name comes from the request, the first `{` is the argument opener, and bytes stream until it closes. Both modes share one rule after that. Both need the arguments to be provably an object before anything goes out, so a bare string, number, null or array streams nothing and surfaces as text. Streaming stays opt-in — the default `Reject` policy still buffers.

Completion subtracts whatever already went out, in both crates, so arguments never arrive twice.

A backend can ignore the grammar and emit its native markup anyway — MiniMax M2 does exactly that under a named `tool_choice`. So only whitespace may come before a bare argument object; anything else shuts the cursor off and hands the payload to the jail's native fallback.

Measured end to end against a served model, long payload, comparing against a baseline of 0 tool-call chunks and one terminal burst.

What the columns mean:

- **tool_choice** — `required` means the model spits out an array of `{name, parameters}`. A **named** tool means it just spits out that one tool's arguments, because we already told it which tool to call.
- **path** — **v1** is the jail, which is what you get by default with `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2` unset. **v2** is the unified parser, which you get with the flag on. Guided streaming does not read that flag at all; v2 is in the table just to show both generations stream.
- **deltas** — how many streaming chunks actually carried a piece of the tool call. Before this change that number was 1: the name and the whole argument blob, dumped at the very end.
- **first delta** — when the first tool-call chunk reached the client, as a percentage of the stream. Lower is better. This is the moment the client finds out which function is being called.
- **delivery span** — so, this is the time between the first and last tool-call delta (guys, pay attention to this one). It is the column that tells streaming apart from buffering. A buffered response hands you everything at once, so its span is basically 0% no matter how many chunks you slice it into. A span near 100% means bytes were flowing the whole time the model was generating.

| tool_choice | path | deltas | first delta | delivery span |
|---|---|---|---|---|
| required | v1 | 169 | 9.2% in | 90.2% |
| named | v1 | 181 | 1.0% in | 98.7% |
| required | v2 | 229 | 19.9% in | 79.4% |
| named | v2 | 260 | 1.4% in | 98.4% |

## Where

- `parsers/v1/src/tool_calling/jail/guided_stream.rs` (new), `jail/mod.rs`, `tests/jail_guided_stream.rs` (new)
- `parsers/v2/src/unified/guided_cursor.rs`, `unified/mod.rs`, `unified/qwen3.rs`, `lib.rs`

### DIFF
--- a/conformance/tests/guided_streaming_contract.rs
+++ b/conformance/tests/guided_streaming_contract.rs
@@ -184,7 +184,9 @@ async fn v1_and_v2_share_the_guided_streaming_contract() {
 /// message: the array shape delivered `},{"name":` as text and the single-call shape
 /// delivered a bare `}`. The cursor releases argument bytes only up to the argument
 /// object's own closing brace, so everything after it is envelope. Once streaming has
-/// committed, that tail is never assistant text.
+/// committed, that tail is never assistant text. v2 had the identical bug at a
+/// different, pre-existing site (frontend-crates#195) - both generations are driven
+/// here, the way the sibling contract test above does.
 #[tokio::test]
 async fn a_truncated_guided_payload_never_leaks_its_envelope_as_text() {
     let cases = [
@@ -203,18 +205,26 @@ async fn a_truncated_guided_payload_never_leaks_its_envelope_as_text() {
     ];
 
     for case in cases {
-        let (carrying, names, arguments, content) = v1(&case).await;
-        assert_eq!(names, vec!["search".to_string()], "{:?}", case.payload);
-        assert_eq!(arguments, case.arguments, "{:?}", case.payload);
-        assert!(
-            content.is_empty(),
-            "truncated payload {:?} leaked envelope bytes as text: {content:?}",
-            case.payload
-        );
-        assert!(
-            carrying > 2,
-            "truncated payload {:?} buffered instead of streaming",
-            case.payload
-        );
+        let v1 = v1(&case).await;
+        let v2 = v2(&case);
+        for (name, output) in [("v1", &v1), ("v2", &v2)] {
+            assert_eq!(
+                output.1,
+                vec!["search".to_string()],
+                "{name} {:?}",
+                case.payload
+            );
+            assert_eq!(output.2, case.arguments, "{name} {:?}", case.payload);
+            assert!(
+                output.3.is_empty(),
+                "{name} truncated payload {:?} leaked envelope bytes as text: {output:?}",
+                case.payload
+            );
+            assert!(
+                output.0 > 2,
+                "{name} truncated payload {:?} buffered instead of streaming",
+                case.payload
+            );
+        }
     }
 }
--- a/conformance/tests/guided_streaming_contract.rs
+++ b/conformance/tests/guided_streaming_contract.rs
@@ -1,0 +1,220 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use dynamo_parsers::tool_calling::jail::{Annotated, JailedStream};
+use dynamo_parsers_v2::{
+    InvalidGuidedPayloadPolicy, Tool, UnifiedParserEvent, UnifiedParserExt, UnifiedParserInit,
+    UnifiedParserStartingState, UnifiedToolOutputMode, create_unified_parser_for_family,
+};
+use dynamo_protocols::types::{
+    ChatChoiceStream, ChatCompletionMessageContent, ChatCompletionStreamResponseDelta,
+    CreateChatCompletionStreamResponse, Role,
+};
+use futures::StreamExt;
+use futures::stream;
+
+struct Case {
+    named: bool,
+    payload: &'static str,
+    arguments: &'static str,
+}
+
+fn v1_chunk(content: &str) -> Annotated<CreateChatCompletionStreamResponse> {
+    #[allow(deprecated)]
+    let choice = ChatChoiceStream {
+        index: 0,
+        delta: ChatCompletionStreamResponseDelta {
+            role: Some(Role::Assistant),
+            content: Some(ChatCompletionMessageContent::Text(content.to_string())),
+            tool_calls: None,
+            function_call: None,
+            refusal: None,
+            reasoning_content: None,
+        },
+        finish_reason: None,
+        logprobs: None,
+    };
+    Annotated {
+        data: Some(CreateChatCompletionStreamResponse {
+            id: "guided-conformance".to_string(),
+            choices: vec![choice],
+            created: 0,
+            model: "test".to_string(),
+            system_fingerprint: None,
+            object: "chat.completion.chunk".to_string(),
+            usage: None,
+            service_tier: None,
+        }),
+        id: None,
+        event: None,
+        comment: None,
+        error: None,
+    }
+}
+
+fn chunks(payload: &str) -> Vec<&str> {
+    payload
+        .char_indices()
+        .map(|(at, ch)| &payload[at..at + ch.len_utf8()])
+        .collect()
+}
+
+async fn v1(case: &Case) -> (usize, Vec<String>, String, String) {
+    let builder = JailedStream::builder().guided_streaming(true);
+    let jail = if case.named {
+        builder.tool_choice_named("search".to_string()).build()
+    } else {
+        builder.tool_choice_required().build()
+    };
+    let input = chunks(case.payload).into_iter().map(v1_chunk);
+    let output: Vec<_> = jail
+        .apply_with_finish_reason(stream::iter(input))
+        .collect()
+        .await;
+
+    let mut carrying = 0;
+    let mut names = Vec::new();
+    let mut arguments = String::new();
+    let mut content = String::new();
+    for response in output {
+        let Some(data) = response.data else {
+            continue;
+        };
+        for choice in data.choices {
+            if let Some(ChatCompletionMessageContent::Text(text)) = choice.delta.content {
+                content.push_str(&text);
+            }
+            if let Some(calls) = choice.delta.tool_calls {
+                carrying += 1;
+                for call in calls {
+                    if let Some(function) = call.function {
+                        if let Some(name) = function.name {
+                            names.push(name);
+                        }
+                        if let Some(fragment) = function.arguments {
+                            arguments.push_str(&fragment);
+                        }
+                    }
+                }
+            }
+        }
+    }
+    (carrying, names, arguments, content)
+}
+
+fn v2(case: &Case) -> (usize, Vec<String>, String, String) {
+    let tools = [Tool {
+        name: "search".to_string(),
+        description: None,
+        parameters: serde_json::json!({"type":"object"}),
+        strict: None,
+    }];
+    let mut parser = create_unified_parser_for_family("qwen3", &tools).unwrap();
+    parser
+        .initialize_request(UnifiedParserInit {
+            prompt_token_ids: Vec::new(),
+            starting_state: UnifiedParserStartingState::None,
+            tool_output_mode: UnifiedToolOutputMode::GuidedJson {
+                named_tool: case.named.then(|| "search".to_string()),
+            },
+            invalid_guided_payload: InvalidGuidedPayloadPolicy::StreamBestEffort,
+        })
+        .unwrap();
+
+    let mut events = Vec::new();
+    for chunk in chunks(case.payload) {
+        events.extend(parser.push(chunk).unwrap());
+    }
+    events.extend(parser.finish().unwrap().events);
+
+    let mut carrying = 0;
+    let mut names = Vec::new();
+    let mut arguments = String::new();
+    let mut content = String::new();
+    for event in events {
+        match event {
+            UnifiedParserEvent::ToolCall(call) => {
+                carrying += 1;
+                if let Some(name) = call.name {
+                    names.push(name);
+                }
+                arguments.push_str(&call.arguments);
+            }
+            UnifiedParserEvent::Text(text) => content.push_str(&text),
+            UnifiedParserEvent::Reasoning(_) => {}
+        }
+    }
+    (carrying, names, arguments, content)
+}
+
+#[tokio::test]
+async fn v1_and_v2_share_the_guided_streaming_contract() {
+    let cases = [
+        Case {
+            named: false,
+            payload: r#"[{"name":"search","parameters":{"query":"Rust","limit":10}}]"#,
+            arguments: r#"{"query":"Rust","limit":10}"#,
+        },
+        Case {
+            named: true,
+            payload: "  {\"query\":\"Rust\",\"limit\":10}",
+            arguments: r#"{"query":"Rust","limit":10}"#,
+        },
+    ];
+
+    for case in cases {
+        let v1 = v1(&case).await;
+        let v2 = v2(&case);
+        for (name, output) in [("v1", &v1), ("v2", &v2)] {
+            assert_eq!(output.1, vec!["search".to_string()], "{name}");
+            assert_eq!(output.2, case.arguments, "{name}");
+            assert!(
+                output.3.is_empty(),
+                "{name} leaked JSON as text: {output:?}"
+            );
+            assert!(
+                output.0 > 2,
+                "{name} buffered instead of streaming: {output:?}"
+            );
+        }
+    }
+}
+
+/// A truncated guided payload used to leak its call envelope into the assistant
+/// message: the array shape delivered `},{"name":` as text and the single-call shape
+/// delivered a bare `}`. The cursor releases argument bytes only up to the argument
+/// object's own closing brace, so everything after it is envelope. Once streaming has
+/// committed, that tail is never assistant text.
+#[tokio::test]
+async fn a_truncated_guided_payload_never_leaks_its_envelope_as_text() {
+    let cases = [
+        // Cut mid-envelope, just after a complete first call.
+        Case {
+            named: false,
+            payload: r#"[{"name":"search","parameters":{"query":"Rust"}},{"name":"#,
+            arguments: r#"{"query":"Rust"}"#,
+        },
+        // Cut before the array's own `]`.
+        Case {
+            named: false,
+            payload: r#"[{"name":"search","parameters":{"query":"Rust"}}"#,
+            arguments: r#"{"query":"Rust"}"#,
+        },
+    ];
+
+    for case in cases {
+        let (carrying, names, arguments, content) = v1(&case).await;
+        assert_eq!(names, vec!["search".to_string()], "{:?}", case.payload);
+        assert_eq!(arguments, case.arguments, "{:?}", case.payload);
+        assert!(
+            content.is_empty(),
+            "truncated payload {:?} leaked envelope bytes as text: {content:?}",
+            case.payload
+        );
+        assert!(
+            carrying > 2,
+            "truncated payload {:?} buffered instead of streaming",
+            case.payload
+        );
+    }
+}
--- a/parsers/v1/src/tool_calling/jail/guided_stream.rs
+++ b/parsers/v1/src/tool_calling/jail/guided_stream.rs
@@ -1,0 +1,962 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Incremental lexer over a guided (`tool_choice`) tool-call payload.
+//!
+//! # Why a lexer and not a few `str::find` calls
+//!
+//! Streaming a guided call before its payload has closed needs four facts that
+//! only a JSON lexer can answer: which array element the scan is inside, that
+//! element's decoded `name`, where its argument OBJECT begins, and which bytes of
+//! that object are safe to put on the wire. Substring search answers none of them
+//! correctly — `"name"` occurs as a VALUE in `{"x":"name","parameters":{}}`,
+//! `_` is six characters that must decode to one, the `parameters` spelling
+//! is as legal as `arguments`, and rescanning the whole accumulated buffer on
+//! every chunk makes a token-at-a-time stream quadratic.
+//!
+//! [`GuidedStreamCursor`] owns that state in one place and advances strictly
+//! FORWARD: every call consumes only the bytes appended since the last one, the
+//! same "scan only what is new" discipline as `JsonCompletionProgress` in this
+//! module's parent.
+//!
+//! # What it deliberately does not do
+//!
+//! It does not validate. The jail's buffered parse remains the only judge of
+//! whether a payload is a legal call set. The cursor answers one narrower
+//! question: *has enough arrived that a name and an argument OBJECT can no longer
+//! turn into something else?* Requiring the argument value to open with `{` is
+//! what makes an early emission safe — `null`, a string, a number and an array
+//! never reach the commit point, so a shape the buffered path would void is never
+//! put on the wire.
+//!
+//! # The one thing it cannot promise
+//!
+//! An emitted byte cannot be retracted. An element that names BOTH `arguments`
+//! and `parameters` is ambiguous, and the cursor refuses to commit it whenever
+//! the second spelling appears before the commit point (which is every ordering
+//! except one: the alias that opens the object arriving first AND the name having
+//! already closed). In that one ordering the call is already on the wire; the
+//! cursor then blocks the element — no further argument bytes, no second call —
+//! rather than pretending it can unsay them.
+
+use std::collections::BTreeMap;
+
+use super::ToolChoiceFormat;
+
+/// The two spellings a guided call may use for its argument object. A call
+/// carrying both is ambiguous — see the module header.
+const ARGUMENT_ALIASES: [&str; 2] = ["arguments", "parameters"];
+
+/// One fragment the cursor is willing to stand behind.
+///
+/// `name` is carried exactly once per call, on the delta that commits it; every
+/// later delta for that call carries argument bytes and `name: None`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct GuidedDelta {
+    pub tool_index: usize,
+    pub name: Option<String>,
+    pub arguments: String,
+}
+
+/// One call whose opening delta has already reached the client.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct StreamedCall {
+    /// Argument bytes released so far, in their original wire representation.
+    pub arguments: String,
+}
+
+/// What the payload is shaped like, derived from the jail's [`ToolChoiceFormat`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum Mode {
+    /// `tool_choice=required`: `[{"name":…,"arguments":{…}}, …]`.
+    Array,
+    /// `tool_choice=<named>`: the payload IS the bare argument object; the name
+    /// is known up front and never appears in the bytes.
+    Single { tool_name: String },
+}
+
+/// Where the scan sits inside a call object (array mode only).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum Slot {
+    /// Between `{` or `,` and the next key.
+    Key,
+    /// A key literal closed; waiting for its `:`.
+    Colon,
+    /// Past the `:`; the next token is that key's value.
+    Value,
+}
+
+/// Per-element scan state. Reset at every call boundary.
+#[derive(Debug, Default, Clone)]
+struct Element {
+    name: Option<String>,
+    /// Offset of the `{` that opened the argument object.
+    args_start: Option<usize>,
+    /// Offset just past the argument object's closing `}`, once it arrived.
+    args_end: Option<usize>,
+    /// The scan is inside the argument object right now.
+    in_args: bool,
+    /// An argument alias key has been seen for this element.
+    alias_seen: bool,
+    /// This element may never commit, and may never release another byte:
+    /// either an alias bound a non-object value, or both spellings appeared.
+    blocked: bool,
+    committed: bool,
+    /// Argument bytes already released, relative to `args_start`.
+    released: usize,
+}
+
+/// Forward-only lexer over one guided payload.
+#[derive(Debug, Clone)]
+pub(crate) struct GuidedStreamCursor {
+    mode: Mode,
+    /// Bytes already lexed. The scan never revisits them.
+    scanned: usize,
+    depth: i32,
+    in_string: bool,
+    escape: bool,
+    /// Opening quote of a literal the cursor needs (a call key, or the `name`
+    /// value). Decoding happens once, when the closing quote arrives, so JSON
+    /// escape semantics stay owned by `serde_json`.
+    literal_start: Option<usize>,
+    /// Depth at which call keys sit: 2 under an array root, 1 under an object
+    /// root. Array mode only.
+    key_depth: i32,
+    root_seen: bool,
+    /// The payload does not open as a call shape; nothing may ever commit.
+    disabled: bool,
+    slot: Slot,
+    pending_key: Option<String>,
+    element: Element,
+    index: usize,
+    /// Calls already put on the wire, keyed in the cursor's source-index space.
+    streamed: BTreeMap<usize, StreamedCall>,
+}
+
+impl GuidedStreamCursor {
+    pub(crate) fn new(format: &ToolChoiceFormat) -> Self {
+        let mode = match format {
+            ToolChoiceFormat::ArrayOfTools => Mode::Array,
+            ToolChoiceFormat::SingleObject { tool_name } => Mode::Single {
+                tool_name: tool_name.clone(),
+            },
+        };
+        Self::with_mode(mode)
+    }
+
+    fn with_mode(mode: Mode) -> Self {
+        Self {
+            mode,
+            scanned: 0,
+            depth: 0,
+            in_string: false,
+            escape: false,
+            literal_start: None,
+            key_depth: 1,
+            root_seen: false,
+            disabled: false,
+            slot: Slot::Key,
+            pending_key: None,
+            element: Element::default(),
+            index: 0,
+            streamed: BTreeMap::new(),
+        }
+    }
+
+    pub(crate) fn reset(&mut self) {
+        let mode = self.mode.clone();
+        *self = Self::with_mode(mode);
+    }
+
+    pub(crate) fn streamed(&self) -> &BTreeMap<usize, StreamedCall> {
+        &self.streamed
+    }
+
+    /// Lex the bytes appended since the last advance and emit whatever became
+    /// safe to send.
+    ///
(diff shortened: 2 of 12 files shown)


parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -682,6 +804,7 @@ impl ChoiceJailState {
             )
             .await;
         unjailed_choice.logprobs = jail_logprobs;
+        self.subtract_streamed_calls(&mut unjailed_choice);
```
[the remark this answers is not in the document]
keivenchang: fixed in 31494db9 — two guided calls no longer collapse into position 0. Before the fix, a client received both argument objects as one call: `{0: "{\"query\":\"Rust\"}{\"query\":\"Zig\"}"}`; the regression test pins that exact failure.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -2369,14 +2548,15 @@ pub fn apply_tool_calling_jail<S>(
     tool_choice: Option<dynamo_protocols::types::ChatCompletionToolChoiceOption>,
     tool_definitions: Option<Vec<crate::tool_calling::ToolDefinition>>,
     uses_tool_call_structural_tag: bool,
+    guided_streaming: bool,
```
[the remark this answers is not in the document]
keivenchang: fixed in 31494db9 — existing callers keep the published five-argument API unchanged, while guided streaming uses the new `apply_tool_calling_jail_with_guided_streaming` API.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -682,9 +804,19 @@ impl ChoiceJailState {
...
+        // Count what this payload REBUILT, not what survives subtraction. A fully
+        // streamed call leaves no chunk behind, so counting survivors would leave the
+        // offset at zero and the next payload would reuse this payload's tool indices.
+        let rebuilt_tool_calls = unjailed_choice
+            .delta
+            .tool_calls
+            .as_ref()
+            .map_or(0, Vec::len);
+        let streamed_tool_calls = self.guided_streamed.len();
+        self.subtract_streamed_calls(&mut unjailed_choice);
```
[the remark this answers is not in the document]
keivenchang: fixed in 3210096c — a nameless array item no longer causes the next valid tool call to merge with an earlier one. Before the fix, the client received `{0: "{\"x\":1}", 1: "{\"x\":1}{\"y\":2}"}`; the regression test covers that skipped-position case.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -525,6 +537,13 @@ impl ChoiceJailState {
         self.is_jailed = false;
         self.accumulated_logprobs = None;
         self.completion_progress.reset();
+        // The cursor's byte offsets and released-byte counts describe THIS payload.
+        // Carrying them into a second jailed value would suppress its arguments as
+        // "already streamed", so they reset at the payload boundary like the lexer.
+        if let Some(cursor) = self.guided_cursor.as_mut() {
+            cursor.reset();
+        }
+        self.guided_streamed.clear();
```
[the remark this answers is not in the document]
keivenchang: I don't think the reported duplication occurs: both completion paths remove arguments already sent before clearing that record, and the duplicate-output tests fail if this step is removed. I did find the nearby mismatch in `finalize`, and fixed it in 9401259c by routing both completion paths through the same reconciliation step.

parsers/v1/src/tool_calling/jail/guided_stream.rs
```diff
@@ -0,0 +1,926 @@
...
+    /// Put the current call on the wire once its shape can no longer change.
+    ///
+    /// Both facts are required: the name is known, AND the argument value has
+    /// been seen to open with `{`. Committing on the name alone is what would let
+    /// `"arguments": null` reach the client as a call.
+    fn maybe_commit(&mut self, out: &mut Vec<GuidedDelta>) {
+        if self.element.committed || self.element.blocked || self.element.args_start.is_none() {
+            return;
+        }
+        let name = match &self.mode {
```
KrishnanPrash, line 432: **AI Generated Comment**

**Comment 1 - guided streaming bypasses the named-tool filter**

```rust
let name = match &self.mode {
    Mode::Single { tool_name } => tool_name.clone(),

// REPRO  tool_choice = named("search")
// payload: {"name":"other_tool","arguments":{"a":1}}
//
//   guided_streaming(false) -> no tool call (filter drops it)
//   guided_streaming(true)  -> name = "search"
//                              args = {"name":"other_tool","arguments":{"a":1}}
```

The cursor commits the requested name before the named-tool filter at `mod.rs:2046` can run, so a wrong-tool payload that `main` drops now reaches the client as an executable call under the trusted name. Read `named_tool_name` in `maybe_commit`, or move the filter before the commit point.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -658,6 +677,140 @@ impl ChoiceJailState {
...
+            let full = chunk
+                .function
+                .as_ref()
+                .and_then(|f| f.arguments.as_deref())
+                .unwrap_or("");
+            // `released` counts bytes this cursor put on the wire for this call, so it
+            // must index into the same argument string the completion path rebuilt. If
+            // it does not, the two disagree about what the client has seen: say so
+            // rather than silently dropping a tail.
+            let remainder = match full.get(released..) {
```
KrishnanPrash: **AI Generated Comment**

**Comment 2 - named arguments are corrupted by an offset mismatch**

```rust
// BEFORE
let remainder = match full.get(released..) {

// AFTER
let Some(tail) = full.strip_prefix(streamed.as_str())

// REPRO  tool_choice = named("search")
// payload:     "  {\"query\":\"Rust\",\"limit\":10}"
// client args: {"query":"Rust","limit":10}0}
```

`released` counts bytes from the argument object's `{`, but for a named choice `full` is the whole accumulated buffer (`mod.rs:2135`), so any byte before that brace shifts the tail and breaks the JSON. Store the released bytes and subtract with `strip_prefix`, which is what v2 does at `parsers/v2/src/unified/mod.rs:1907`.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -658,6 +677,140 @@ impl ChoiceJailState {
...
+    /// Strip from a completed choice everything the cursor already put on the wire.
+    ///
+    /// The completion path rebuilds every call in full. For a call that was streamed,
+    /// the client already has its id, name and the first `released` argument bytes, so
+    /// re-sending them would duplicate the arguments. Emit only the tail.
+    fn subtract_streamed_calls(&self, unjailed: &mut ChatChoiceStream) {
+        if self.guided_streamed.is_empty() {
+            return;
+        }
+        let Some(tool_calls) = unjailed.delta.tool_calls.as_mut() else {
```
KrishnanPrash: **AI Generated Comment**

**Comment 3 - a truncated payload also goes out as assistant text**

```rust
let Some(tool_calls) = unjailed.delta.tool_calls.as_mut() else {
    return;
};

// REPRO  named, payload {"q":"Par
//   args = {"q":"Par
//   text = {"q":"Par
//
// control, guided_streaming(false)
//   args = ""
//   text = {"q":"Par
```

A truncated payload rebuilds no tool calls, so this early return skips the ledger and the fallback at `mod.rs:2089` sends the same bytes again as visible content. Apply the released span to the content channel too, as v2 does at `parsers/v2/src/unified/mod.rs:2062`; `required` and `named` both show this.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -658,6 +677,140 @@ impl ChoiceJailState {
...
+            return;
+        };
+        tool_calls.retain_mut(|chunk| {
+            let local = (chunk.index as usize).saturating_sub(self.emitted_tool_calls_count);
```
KrishnanPrash: **AI Generated Comment**

**Comment 4 - one call is delivered twice on two indices**

```rust
let local = (chunk.index as usize).saturating_sub(self.emitted_tool_calls_count);

// REPRO  tool_choice = required
// payload: [{"parameters":{"q":1}},{"name":"b","parameters":{"x":1}}]
//
//   idx 1  id=A  name=b  {"x":1}   (streamed)
//   idx 0  id=B  name=b  {"x":1}   (rebuilt whole)
```

The cursor indexes by array element position and keeps counting elements it refuses to commit, but `create_tool_call_choice` renumbers the survivors densely at `mod.rs:2064`, so this lookup misses and a fully streamed call is sent again with a second id. Give both sides one index space, or key the ledger on a value the rebuild can also compute.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -658,6 +677,140 @@ impl ChoiceJailState {
...
+                .as_ref()
+                .and_then(|f| f.arguments.as_deref())
+                .unwrap_or("");
+            // `released` counts bytes this cursor put on the wire for this call, so it
+            // must index into the same argument string the completion path rebuilt. If
+            // it does not, the two disagree about what the client has seen: say so
+            // rather than silently dropping a tail.
+            let remainder = match full.get(released..) {
+                Some(tail) => tail.to_string(),
+                None => {
```
KrishnanPrash: **AI Generated Comment**

**Comment 5 - the disagreement fallback guarantees the duplication it warns about**

```rust
// BEFORE
None => {
    tracing::warn!(...);
    full.to_string()
}

// AFTER
None => {
    tracing::warn!(...);
    return false;
}
```

This branch runs only after the cursor already put bytes on the wire, so sending `full` on top of them always duplicates the arguments. Keep the warning and emit nothing.

KrishnanPrash (commented):
**AI Generated Comment**

This review was produced by an AI agent (Claude Code) and reviewed against the head commit `9401259`. Every finding below was reproduced locally with a failing test in a scratch worktree; the control runs used `guided_streaming(false)` to confirm the behaviour is new in this PR.

The feature is valuable and the lexer is well built. The problem is the seam between the cursor and the buffered parser: the two disagree about byte offsets, about index spaces, and about the content channel. v2 has none of these because it reconciles with one absolute offset into the one buffer its cursor lexed.

Four blocking findings (comments 1 to 4), three strong recommendations (5 to 7), three nits (8 to 10). All are in the v1 path. Full suites pass at head, so none of these are covered by an existing test.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -212,6 +214,14 @@ struct ChoiceJailState {
...
     /// worthwhile. Parser acceptance is never cached here.
     completion_progress: JailCompletionProgress,
+    /// Guided-payload cursor, present only in `Immediate` mode. Under a guided
+    /// grammar the payload's shape is already known, so calls can be released as
+    /// they arrive instead of at the closing brace.
+    guided_cursor: Option<GuidedStreamCursor>,
+    /// What the cursor has already put on the wire, per tool index: the number of
+    /// ARGUMENT bytes released. The completion path must emit only the remainder,
+    /// or the client receives the arguments twice.
+    guided_streamed: BTreeMap<usize, usize>,
```
KrishnanPrash: **AI Generated Comment**

**Comment 6 - v1 and v2 reconcile with two different designs**

```rust
// v1  BTreeMap<sparse idx, relative bytes>
//     matched against a string that a different parser rebuilt
//
// v2  cursor.released_end(), absolute,
//     into the same buffer the cursor lexed
```

v2 keeps one offset in one buffer and has none of the four blocking defects above, while v1's ledger crosses two coordinate systems and only holds on the `RawValue` rebuild path. Consider giving v1 the cursor with v2's `released_end()` and `committed()` API and deleting the ledger; note also that `grep -r 'guided_streaming\|StreamBestEffort' conformance/` returns nothing, so the feature has no v1/v2 parity coverage.

parsers/v1/tests/jail_guided_stream.rs
```diff
@@ -0,0 +1,358 @@
...
+            end += 1;
+        }
+        out.push(chunk(&payload[at..end]));
+        at = end;
+    }
+    out
+}
+
+/// (number of responses carrying tool-call deltas, name deltas, concatenated arguments)
+async fn drive(
```
KrishnanPrash, line 68: **AI Generated Comment**

**Comment 7 - the test harness cannot see the text channel**

```rust
async fn drive(
    jail: JailedStream,
    chunks: Vec<Annotated<CreateChatCompletionStreamResponse>>,
) -> (usize, Vec<String>, String) {

// reads delta.tool_calls only; never reads delta.content
```

`drive` never reads `delta.content`, so no test in this file can observe comment 3, and `truncated_after_commit_does_not_duplicate_streamed_arguments` stays green while the client receives the payload twice. Return the content channel as well, and add fixtures with leading whitespace and with an envelope payload; every payload here is compact JSON, which is the one shape where the offset mismatch cannot fire.

parsers/v1/src/tool_calling/jail/guided_stream.rs
```diff
@@ -0,0 +1,926 @@
...
+//!
+//! An emitted byte cannot be retracted. An element that names BOTH `arguments`
+//! and `parameters` is ambiguous, and the cursor refuses to commit it whenever
+//! the second spelling appears before the commit point (which is every ordering
+//! except one: the alias that opens the object arriving first AND the name having
+//! already closed). In that one ordering the call is already on the wire; the
+//! cursor then blocks the element — no further argument bytes, no second call —
+//! rather than pretending it can unsay them.
+
+// The jail wires this cursor in separately; until then only the tests below
```
KrishnanPrash: **AI Generated Comment**

**Comment 8 (nit) - stale allow(dead_code) and a comment that is now false**

```rust
// BEFORE
// The jail wires this cursor in separately; until then only the tests
// below exercise it, and `dead_code` would fire on every item.
#![allow(dead_code)]

// AFTER
// (delete all three lines)
```

This PR is the one that wires the cursor in, so the comment now states the opposite of the truth. I removed the attribute and ran `cargo check -p dynamo-parsers` with and without `--all-targets`: zero warnings either way.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -2213,6 +2409,17 @@ impl JailedStreamBuilder {
     }
 
     /// Enable immediate jail mode for tool_choice=required
+    /// Release guided tool calls incrementally instead of buffering to the closing brace.
```
KrishnanPrash, line 2439: **AI Generated Comment**

**Comment 9 (nit) - a doc comment moved onto the wrong function**

```rust
// BEFORE
/// Enable immediate jail mode for tool_choice=required
/// Release guided tool calls incrementally ...
pub fn guided_streaming(...)

pub fn tool_choice_required(...)

// AFTER
/// Release guided tool calls incrementally ...
pub fn guided_streaming(...)

/// Enable immediate jail mode for tool_choice=required
pub fn tool_choice_required(...)
```

The first line of `guided_streaming`'s doc block belongs to `tool_choice_required`, which now has no doc comment at all.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -2371,12 +2579,39 @@ pub fn apply_tool_calling_jail<S>(
...
+///
+/// Separate entry point so the existing signature keeps working for published
+/// consumers: incremental release changes the emission SHAPE (one terminal tool-call
+/// delta becomes a name delta followed by argument fragments), so a caller opts in.
+pub fn apply_tool_calling_jail_with_guided_streaming<S>(
+    tool_call_parser: Option<String>,
+    tool_choice: Option<dynamo_protocols::types::ChatCompletionToolChoiceOption>,
+    tool_definitions: Option<Vec<crate::tool_calling::ToolDefinition>>,
+    uses_tool_call_structural_tag: bool,
+    guided_streaming: bool,
```
KrishnanPrash, line 2654: **AI Generated Comment**

**Comment 10 (nit) - two adjacent unnamed bools in a published signature**

```rust
uses_tool_call_structural_tag: bool,
guided_streaming: bool,
```

A caller that swaps these two arguments still compiles and silently gets the wrong behaviour. The crate already exposes `JailedStreamBuilder`; taking a builder, or a small two-field config struct, makes the call site self-describing.

parsers/v1/src/tool_calling/jail/guided_stream.rs
keivenchang: v1 treats the generated object as arguments for the tool the request already selected. For example, if the request names `search`, `{"name":"other_tool","arguments":{"a":1}}` is passed to `search` as input; waiting for the full object to guess whether it names another tool would remove streaming. v2 makes the same tradeoff.

parsers/v1/src/tool_calling/jail/mod.rs
keivenchang: fixed in 63f277c7 — leading spaces can no longer corrupt the tool arguments.

parsers/v1/src/tool_calling/jail/mod.rs
keivenchang: fixed in 63f277c7 — an incomplete tool call no longer appears again as visible assistant text.

parsers/v1/src/tool_calling/jail/mod.rs
keivenchang: fixed in 63f277c7 — one tool call can no longer be sent twice under two different positions.

parsers/v1/src/tool_calling/jail/mod.rs
keivenchang: fixed in 63f277c7 — if the final parse disagrees with arguments already sent, v1 keeps the sent result and does not send a duplicate.

parsers/v1/src/tool_calling/jail/mod.rs
keivenchang: fixed in 63f277c7 — v1 now keeps one record of arguments already sent, and the shared test verifies that v1 and v2 stream the same result.

parsers/v1/tests/jail_guided_stream.rs
keivenchang: fixed in 63f277c7 — the tests now verify visible assistant text as well as tool calls, with coverage for leading spaces, incomplete calls, skipped calls, disagreement, and every input split.

parsers/v1/src/tool_calling/jail/guided_stream.rs
keivenchang: fixed in 63f277c7 — removed the stale allowance and comment.

parsers/v1/src/tool_calling/jail/mod.rs
keivenchang: fixed in 63f277c7 — moved each description to the function it actually describes.

parsers/v1/src/tool_calling/jail/mod.rs
keivenchang: fixed in 63f277c7 — callers now choose guided streaming through a named API, so two adjacent true/false settings cannot be swapped.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -1864,11 +2074,10 @@ impl JailedStream {
...
                 }
 
-                // Assign final indices: renumber survivors 0..n (no gaps from
-                // the filter) then add the cumulative offset for consistency
-                // with the MarkerBased branch across multi-emission streams.
-                for (new_idx, chunk) in tool_call_chunks.iter_mut().enumerate() {
-                    chunk.index = (tool_call_offset + new_idx) as u32;
+                // Add the cumulative offset without renumbering: required-choice
+                // indices are source array positions and may be sparse.
+                for chunk in &mut tool_call_chunks {
```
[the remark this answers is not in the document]
keivenchang: fixed in 214337e9 — after a named request filters out the wrong tool, the matching call remains at position 0; the regression runs with streaming both off and on.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -658,6 +672,149 @@ impl ChoiceJailState {
...
+                });
+                true
+            });
+            if tool_calls.is_empty() {
+                unjailed.delta.tool_calls = None;
+            }
+        }
+
+        if !had_rebuilt_calls {
+            let released_end = cursor.released_end().min(payload.len());
```
KrishnanPrash: **AI Generated Comment**

**Comment 11 - the content fix subtracts the argument object, not the call envelope**

```rust
let released_end = cursor.released_end().min(payload.len());
let remainder = &payload[released_end..];

// REPRO  tool_choice = required, stream truncated
// payload: [{"name":"a","parameters":{"x":1}},{"name":
//   guided_streaming(true)  -> call a, args {"x":1}
//                              text = },{"name":        <- leaked
//
// payload: [{"name":"a","parameters":{"x":1}}
//   guided_streaming(true)  -> call a, args {"x":1}
//                              text = }                 <- leaked
```

`released_end` stops at the closing brace of the argument OBJECT, so a truncation landing after the arguments close but before the payload ends leaves call-envelope syntax in the remainder and sends it as visible assistant text. This is not a regression: `guided_streaming(false)` leaks the entire raw payload on the same inputs, so the change strictly improves this case, it just does not close it. Since `JailMode::Immediate` means the whole payload is guided JSON by construction, a remainder is never assistant prose, so the smaller fix is to emit no content at all once the cursor has committed and the rebuild produced no calls; a backend that ignored the grammar leaves the cursor `disabled` with nothing committed, so the native-markup fallback is untouched. Cheapest way to pin it: `conformance/tests/guided_streaming_contract.rs` already asserts `output.3.is_empty()` with the message "leaked JSON as text", so adding a truncated third case makes your own harness fail on the v1 side today.

parsers/v1/src/tool_calling/jail/guided_stream.rs
```diff
@@ -0,0 +1,957 @@
...
+        let from = start + self.element.released;
+        if bound <= from {
+            return;
+        }
+        let fragment = payload[from..bound].to_string();
+        self.element.released = bound - start;
+        self.released_end = self.released_end.max(bound);
+        self.streamed
+            .get_mut(&self.index)
+            .expect("a committed element has a streamed-call record")
```
KrishnanPrash: **AI Generated Comment**

**Comment 12 - a panic in the per-chunk streaming path guards a four-method invariant**

```rust
self.streamed
    .get_mut(&self.index)
    .expect("a committed element has a streamed-call record")
    .arguments
    .push_str(&fragment);
```

I traced this and believe it is unreachable today: `flush` returns early unless `element.committed`, `maybe_commit` is the only writer of that flag and it inserts at the same `self.index`, and `finish_element` and `reset` both clear the flag whenever they move the index. The concern is that the invariant is maintained across four methods in a lexer built to be extended, and the cost of a future commit or advance site getting it wrong is a panic inside a per-chunk streaming path. `if let Some(record)` is the same number of lines and degrades to a dropped fragment rather than a killed request.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -658,6 +672,149 @@ impl ChoiceJailState {
...
+                    return true;
+                };
+                let full = chunk
+                    .function
+                    .as_ref()
+                    .and_then(|f| f.arguments.as_deref())
+                    .unwrap_or("");
+                let remainder = full
+                    .strip_prefix(&streamed.arguments)
+                    .or_else(|| full.trim_start().strip_prefix(streamed.arguments.as_str()));
```
KrishnanPrash: **AI Generated Comment**

**Comment 13 (nit) - the second-chance prefix match may now be dead**

```rust
let remainder = full
    .strip_prefix(&streamed.arguments)
    .or_else(|| full.trim_start().strip_prefix(streamed.arguments.as_str()));
```

The `trim_start` retry was needed when the rebuilt arguments could carry leading whitespace, but both rebuild paths now return `RawValue::get()`, which is the object's own source slice and cannot have any. If that is right the retry is dead; if it is not right, it is a silent second chance that can match a different call whose arguments share a prefix, which is the failure the strict `strip_prefix` was chosen to prevent. Worth resolving either way rather than keeping it as a hedge.

parsers/v1/src/tool_calling/jail/mod.rs
```diff
@@ -874,6 +1039,11 @@ impl ChoiceJailState {
                 .await;
             // Attach the full accumulated logprobs to the final choice
             final_choice.logprobs = self.take_accumulated_logprobs();
+            // Same rule as normal completion: a truncated payload still rebuilds every
+            // call in full here, so anything the cursor already streamed must be
+            // subtracted or its arguments are delivered twice.
+            let accumulated = self.accumulated_content.clone();
```
KrishnanPrash: **AI Generated Comment**

**Comment 14 (nit) - a full-buffer clone that `mem::take` would avoid**

```rust
let accumulated = self.accumulated_content.clone();
self.reconcile_streamed_calls(&mut final_choice, &accumulated);
```

`end_jail` clears `accumulated_content` a few lines below, so `std::mem::take` satisfies the same borrow without copying the whole buffer at stream end.

KrishnanPrash (commented):
**AI Generated Comment**

Re-review of `214337e` by an AI agent (Claude Code), following up on the earlier review of `9401259`. Every point below was checked against this head; the truncation finding was reproduced locally with matching `guided_streaming(false)` control runs.

All ten earlier comments are addressed, and the four blocking ones were fixed structurally rather than patched: the `guided_streamed` ledger is gone in favour of the cursor owning `streamed()` and `released_end()`, subtraction is now a prefix test, both rebuild paths return raw `RawValue` bytes, and `parse_indexed_calls` gives the cursor and the buffered parser one index space. That removes three of the four defects by construction. The new `guided_streaming_contract` test is a real cross-generation contract, not a token one: it asserts byte-identical arguments, no text leak, and a chunk count a buffered implementation cannot reach.

Verified at this head: `dynamo-parsers` and `dynamo-parsers-v2` pass across 16 test binaries with 0 failures, and `guided_streaming_contract` passes on its own.

Nothing below is blocking. One SHOULD (comment 11) is a real but non-regressing leak, one SHOULD (comment 12) is robustness, and two are nits.

parsers/v1/src/tool_calling/jail/mod.rs
keivenchang: fixed in 8b9dcc67 — reproduced both cases: the truncated array leaked `},{"name":` and the single-call cut leaked `}` into assistant text. Once v1 has streamed a call, those remaining bytes are tool-call envelope, so it now emits no text; the regression test pins both cuts and keeps the normal text fallback when a backend ignores the constraint.

parsers/v1/src/tool_calling/jail/guided_stream.rs
keivenchang: fixed in 8b9dcc67 — confirmed the current code cannot reach this panic, then removed it because a future change could still kill a request mid-stream. Tests now stop on a broken invariant, while production logs the error and drops one fragment instead of the whole request.

parsers/v1/src/tool_calling/jail/mod.rs
keivenchang: fixed in 8b9dcc67 — confirmed `RawValue::get()` strips surrounding whitespace, so the retry was dead and could match the wrong call. Removed it; the leading-whitespace and every-split regressions cover the strict match.

parsers/v1/src/tool_calling/jail/mod.rs
keivenchang: fixed in 8b9dcc67 — removed the clone entirely, along with the unused payload input and its only related offset field. The code no longer copies the full buffer when the stream ends.

conformance/tests/guided_streaming_contract.rs
```diff
@@ -0,0 +1,220 @@
...
+        // Cut before the array's own `]`.
+        Case {
+            named: false,
+            payload: r#"[{"name":"search","parameters":{"query":"Rust"}}"#,
+            arguments: r#"{"query":"Rust"}"#,
+        },
+    ];
+
+    for case in cases {
+        let (carrying, names, arguments, content) = v1(&case).await;
```
KrishnanPrash: **AI Generated Comment**

**Comment 15 - the new truncation test is in the parity file but only runs v1**

```rust
// this test
let (carrying, names, arguments, content) = v1(&case).await;

// its sibling, same file, line 168
for (name, output) in [("v1", &v1), ("v2", &v2)] {

// REPRO  the same truncated cases through BOTH generations
// [{"name":"search","parameters":{"query":"Rust"}},{"name":
//     v1 text ""      v2 text },{"name":
// [{"name":"search","parameters":{"query":"Rust"}}
//     v1 text ""      v2 text }
```

This new test asserts the right property but only drives v1, while its sibling in the same parity file loops over both generations, so it cannot observe that v2 leaks exactly the bytes v1 stopped leaking on exactly these inputs. To be clear about scope: the v2 leak itself is PRE-EXISTING and not yours. `released_end` already appears four times in `parsers/v2/src/unified/mod.rs` at base `bd87983` from #190, the leaking site is `mod.rs:2062` (`raw_payload[released_end..]`), and both uses this PR adds are in `settle_streamed_named`, which is clean. So the v2 fix is a follow-up issue, not a change to make here. The in-scope part is one line: loop the truncated cases over both generations the way the sibling test does. It will fail on the v2 side today, which is precisely why it belongs in this file.

KrishnanPrash (approved):
**AI Generated Comment**

Third-round review of `8b9dcc6` by an AI agent (Claude Code), following the reviews of `9401259` and `214337e`. **Approving.** Nothing open is blocking, and the one comment below is a test-coverage note with a one-line fix.

All four comments from the previous round were taken, and the fix for the clone cascaded further than I asked: dropping it let you remove the `payload` parameter from both reconcile functions and retire the `released_end` field and accessor, which had no other reader. I checked for orphaned readers and there are none.

Verified at this head, not assumed:

- `cargo clippy -p dynamo-parsers --all-targets -- -D warnings` is clean, so the `debug_assert!(false, ..)` in the `expect` replacement does not trip CI. I had this queued as a finding and dropped it after checking.
- I tested the mirror of the guard this commit adds. It suppresses `delta.content` once the cursor has committed, so the risk is eating content that should survive. A complete required payload followed by prose, a complete named payload followed by prose, and two payloads in one stream with the second truncated all behave identically with the flag on and off, prose intact. Native `<tool_call>` markup is byte-identical on and off, so the cursor still stays disabled and the marker fallback is untouched. No mirror bug.
- `dynamo-parsers`, `dynamo-parsers-v2` and `dynamo-conformance-fixtures-v2`: 28 test-result lines, 0 failures.

Over three rounds this went from four blocking defects to none, and the fixes were structural rather than local: the ledger moved into the cursor, subtraction became a prefix test, both rebuild paths return raw `RawValue` bytes, and one index space replaced two. Thanks for the pushback on the named-choice comment in the first round, which was mine to get wrong.

conformance/tests/guided_streaming_contract.rs
rmccorm4, line 1: > This is opt-in through JailedStreamBuilder::guided_streaming, because turning it on changes the emission shape for every consumer of the jail.

1. What's the user facing way to opt-in to this when using dynamo?
2. When would we recommend they opt-in or out?

conformance/tests/guided_streaming_contract.rs
keivenchang: You're right. I reproduced both v2 leaks: the truncated array emitted `},{"name":` as assistant text, and the single-call cut emitted `}`. I initially filed #195 because the bug predates this PR, but fixed it here in bed5a646 after deciding v1 and v2 should stay aligned; the conformance test now runs both versions, and #195 is closed.

conformance/tests/guided_streaming_contract.rs
keivenchang: For v1, currently, Dynamo turns guided streaming on by default for `required` or named-tool requests that use JSON-guided output; there is no extra setting. Use `required` when any tool must be called, a named choice when one specific tool must be called, and `auto` when the model may return normal text.

For v2, turn on `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2`. That switch enables v2 for supported requests, so treat it as opting into the full v2 parser path, not only guided streaming.

PR #196, opened 2026-08-21
feat: Kimi K3 structural tag default

indrajit96:
## Summary

This is the grammar-building half of a coordinated change addressing two
tool-calling correctness problems:

1. **Kimi K3 grammar correctness.** K3 `tool_choice="auto"` requests could begin
   a native XTML tool call but still emit undeclared tools, empty required
   arguments, incorrectly typed values, duplicate required arguments, or
   malformed XTML that the parser could not consume.

2. **Schema enforcement for default-on structural guidance.** The generic
   structural-tag policy treated an omitted OpenAI `strict` field as relaxed.
   The new policy uses the declared schema when `strict` is omitted or `true`;
   only explicit `strict:false` opts out.

This PR implements those parser and grammar changes. The companion
[ai-dynamo/dynamo#13654](https://github.com/ai-dynamo/dynamo/pull/13654)
owns deployment defaults, activation policy, frontend integration, and
compatibility fallback.

## Kimi K3 grammar hardening

Kimi K3 emits tool calls through its native XTML `tools`, `call`, and `argument`
elements. The new `auto` builder follows that model-native format while
preserving the semantics of automatic tool choice:

- ordinary assistant text remains unconstrained;
- guidance activates lazily only after the model emits
  `<|open|>tools<|sep|>`;
- calls are restricted to tool names declared in the request;
- automatic calls require positive call indices;
- `parallel_tool_calls:false` limits generation to one call;
- required string arguments must be non-empty;
- integer, number, boolean, array, object, and null arguments use typed content;
- `$ref`, `$defs`, `anyOf`, `oneOf`, `allOf`, `enum`, and `const` participate in
  XTML type inference;
- required arguments use the schema's canonical `required` order, avoiding
  factorial grammar expansion;
- loose text between arguments is replaced by complete declared optional
  argument tags, preventing duplicate required keys from overwriting a
  constrained value;
- schemas with only optional properties may emit an empty argument body;
- ambiguous schemas fall back to parser-compatible XTML instead of generating
  malformed calls; and
- named tool choice prevents a second generic tool-call representation from
  being emitted in the response channel.

`TriggeredTagsFormat` also gains optional `excludes`, allowing the K3 grammar to
reject invalid reasoning and call-marker transitions before the tools trigger.
Existing builders serialize unchanged when no exclusions are configured.

```mermaid
flowchart LR
    A[K3 begins generation] --> B{Opens tools channel?}
    B -- No --> C[Normal assistant answer]
    B -- Yes --> D[Triggered K3 grammar]
    D --> E[Select a declared tool]
    E --> F[Generate call index]
    F --> G{Schema representable?}
    G -- Yes --> H[Required typed arguments followed by declared optional arguments]
    G -- No --> I[Parser-compatible XTML fallback]
    H --> J[Valid native K3 call]
    I --> J
    J --> K[K3 parser produces OpenAI tool_calls]
```

`tool_choice="auto"` still does not force a tool call. The grammar constrains a
call only after the model chooses to begin the native tools channel.

## Shared schema behavior

`StructuralTagSchemaMode::Auto` now follows these rules for all structural-tag
builders:

| Tool setting | Argument guidance |
|---|---|
| `strict` omitted | Use the declared parameter schema |
| `strict:true` | Use the declared parameter schema |
| `strict:false` | Keep the native envelope but relax argument-schema enforcement |
| Global schema mode `strict` | Use the declared schema, overriding `strict:false` |

This PR does not change parser-to-model mappings or deployment defaults.

## Correctness validation

Kimi K3 Dynamo and SGLang KVV correctness before and after structural guidance

With Dynamo and SGLang, K3 KVV correctness improved:

- non-thinking: `25/49 = 51.0%` to `48/49 = 98.0%`;
- thinking: `69/98 = 70.4%` to `94/98 = 95.9%`.

The controlled native SGLang ablation provides the strongest causal evidence:

- structural guidance enabled: `47/48 = 97.9%` non-thinking and
  `90/92 = 97.8%` thinking;
- structural guidance disabled: `22/48 = 45.8%` and `61/92 = 66.3%`;
- re-enabling guidance restored the original scores.

An activation-only Dynamo patch improved non-thinking to `37/49` but did not
repair thinking correctness, exposing the old empty and weakly typed argument
grammar. The complete builder was needed for the final result.

A later near-full Dynamo and vLLM run using this frontend-crates head reached
`47/48 = 97.9%` non-thinking and `83/97 = 85.6%` thinking after excluding
context-length cases.

## Performance validation

Kimi K3 p50 TTFT with and without the structural-guidance changes

The tested Kimi workloads did not show a TTFT regression:

- concurrency-1 probe: `0.400 s` without the change and `0.356 s` with it;
- concurrency-24 streaming: `1.92 s` without the change and `1.79 s` with it.

These are two measured workload shapes and should not be interpreted as a
general performance improvement.

## Unit validation

Focused tests cover:

- optional `auto` activation and normal-answer passthrough;
- thinking-to-tools transitions;
- declared and named tool selection;
- required and parallel calls;
- canonical required-argument ordering;
- duplicate-required-key prevention;
- optional-only schemas and empty argument bodies;
- typed string, numeric, boolean, array, object, and null arguments;
- local references and schema-composition keywords;
- ambiguous-schema fallback; and
- serialization compatibility for existing structural-tag builders.


parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -56,6 +61,219 @@ fn one_of(elements: Vec<Format>) -> Format {
     }
 }
 
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
```
rmccorm4, line 64: The large majority of changes in this file seem generic to me, not specific to Kimi K3 - can we refactor it to be in a common module?

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -228,6 +586,9 @@ pub(crate) fn build_kimi_k3(
     if tools.is_empty() {
         return Ok(None);
     }
+    if matches!(ctx.tool_choice, crate::tool_calling::ToolChoice::Auto) {
+        return Ok(Some(build_auto_structural_tag(tools, ctx)));
+    }
```
rmccorm4, line 655: Can you explain the difference in enforcing structural tags for `tool_choice=auto` vs. `tool_choice=required`? Wouldn't we want a similar guidance/enforcement for `tool_choice=required` and named tools?

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Reason I added this here was to ONLY have this behavior for kimi_k3.
Do we want to move this to other models as well?
I will run some perf numbers on Nemotron Lightning today and then address this.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: This was added after inspecting KVV.
Only the `auto` path needs different handling because a tool call is optional:
Named choice is already constrained through the existing mandatory path: it selects only the named tool and requires a call. 
Required choice also uses the existing mandatory grammar when structural tags are enabled; its default K3 path currently relies on the prompt-level XTML tool-choice instruction.
Our E2E tests have verified this behavior for named and required tool calls.
I will run the custom suite again and post result here.
Currently wont change code.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,6 +412,146 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
...
+                    .flatten()
+                    .map(|xtml_type| nonempty_argument_format(key, xtml_type))
+            })
+            .collect();
+        if alternatives.is_empty() {
+            return Format::AnyText(AnyTextFormat { excludes: vec![] });
+        }
+        elements.push(one_of(alternatives));
+    } else {
+        for key in required {
```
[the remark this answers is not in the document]
indrajit96: Good catch. Fixed this so the JSON Schema required array no longer determines XTML argument order.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,6 +412,146 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
...
+                matches!(schema, Value::Bool(_) | Value::Object(_))
+                    .then(|| single_xtml_type(schema, &root))
+                    .flatten()
+                    .map(|xtml_type| nonempty_argument_format(key, xtml_type))
+            })
+            .collect();
+        if alternatives.is_empty() {
+            return Format::AnyText(AnyTextFormat { excludes: vec![] });
+        }
+        elements.push(one_of(alternatives));
```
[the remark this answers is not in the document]
indrajit96: Comment is correct.
Fixed this.
But this slightly reduces KVV score Thinking decreased from 94/98 to 87/98.
But the fix is correct and needed

Issue : Current PR does not respect this schema such as
```
  {
    "type": "object",
    "properties": {
      "path": {"type": "string"}
    },
    "required": []
  }
```
Because `required` is empty, path is optional and the model may legally emit no arguments, equivalent to `{}`. However, our grammar in the PR was more aggressive than the declared schema and forced the model to generate at least one argument.

The fix makes the argument sequence optional when the schema has no required properties. If the model chooses to emit an argument, we emit that.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,6 +412,146 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
...
+        begin: format!("{OPEN}call tool=\"{}\" index=\"", escape_attr(&tool.name)),
+        content: Box::new(Format::Sequence(SequenceFormat {
+            elements: vec![
+                Format::Regex(RegexFormat {
+                    pattern: "[1-9][0-9]*".to_string(),
+                }),
+                Format::ConstString(ConstStringFormat {
+                    value: format!("\"{SEP}"),
+                }),
+                auto_arguments_block(tool),
```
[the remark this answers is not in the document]
indrajit96: Fixed!
Issue : The new auto-mode grammar ignored the tool’s explicit strict: false setting. It continued using the declared parameter schema and forced required fields, inferred XTML types, and non-empty values.

For example:
```
  {
    "name": "read_file",
    "strict": false,
    "parameters": {
      "type": "object",
      "properties": {
        "path": {"type": "string"}
      },
      "required": ["path"]
    }
  }
```
Before the fix, auto mode still forced a non-empty path, despite the caller explicitly opting out of strict schema enforcement.

The fix now passes the request’s strictness policy into the auto argument builder:

  - Omitted strict or strict: true uses the declared schema.
  - Explicit strict: false permits zero or more schema-unconstrained arguments.
  - Global strict-schema mode overrides strict: false and restores schema enforcement.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,6 +413,207 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
...
+        content: Box::new(Format::Regex(RegexFormat {
+            pattern: format!("{STRING_ATOM}+"),
+        })),
+        end: ARGUMENT_CLOSE.to_string(),
+    })
+}
+
+fn loose_auto_arguments() -> Format {
+    Format::AnyText(AnyTextFormat {
+        excludes: vec![CALL_OPEN.to_string(), CALL_CLOSE.to_string()],
```
rmccorm4: [P1] These `AnyText` gaps exclude only call markers, so they can emit a duplicate required argument. XGrammar 0.2.1 accepted `city=Paris` followed by an empty `city`; the K3 parser’s last-write-wins behavior returned `{"city":""}`. Exclude required argument openers or emit a canonical non-duplicating sequence.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,6 +413,207 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
...
+    })
+}
+
+fn unordered_required_arguments(arguments: &[Format]) -> Format {
```
rmccorm4: [P1] This materializes every required-argument order: one five-field tool serialized to 161,337 bytes and 120 alternatives. At six fields, line 526 falls back to `any_text` and drops required-field enforcement. Guided generation only needs one deterministic valid order; keep a canonical order instead.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -56,6 +62,219 @@ fn one_of(elements: Vec<Format>) -> Format {
...
+    } else if NUMBER_KEYWORDS.iter().any(|key| schema.contains_key(*key)) {
+        vec![JsonType::Number]
+    } else {
+        JSON_TYPES.to_vec()
+    }
+}
+
+fn single_xtml_type(schema: &Value, root: &Value) -> Option<&'static str> {
+    let types = schema_types(schema, root, &HashSet::new());
+    (types.len() == 1).then(|| types[0].xtml_name())
```
rmccorm4, line 274: [P2] Check cardinality after mapping and deduplicating XTML types. `anyOf(integer, number)` has two JSON types but both map to XTML `number`; exact head therefore serializes the required argument block as unrestricted `any_text`.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,6 +413,207 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
...
+
+fn auto_arguments_block(tool: &ToolDefinition, strict_schema: bool) -> Format {
+    if !super::builder::kimi_uses_declared_tool_schema(tool, strict_schema) {
+        return arguments_block(None);
+    }
+
+    let Some(parameters) = tool.parameters.as_ref().and_then(Value::as_object) else {
+        return Format::AnyText(AnyTextFormat { excludes: vec![] });
+    };
+    let root = Value::Object(parameters.clone());
```
rmccorm4, line 507: [P3] This clones the complete parameter schema once per tool solely to provide a read-only root for inference. Borrow the original `Value`, or let the walker accept the existing map.

rmccorm4 (commented):
Local exact-head review findings.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
rmccorm4: I mostly mean to put definitions, functions, helpers, etc. that are generic into something like `parsers/v1/src/tool_calling/structural_tag/common.rs` if they're not unique to kimi in how they're defined. It's OK if they're only used/called by 1 or 2 models for now - they just mostly didn't seem kimi specific in implementation.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,6 +413,207 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
     star(one_of(tags))
 }
 
+fn nonempty_argument_format(key: &str, xtml_type: &str) -> Format {
```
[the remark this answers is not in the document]
indrajit96: Fixed in the current head. K3 Auto now represents required non-string content with `Format::JsonSchema`; raw XTML regex content is used only for strings. `auto_required_non_string_arguments_use_their_json_schemas` covers integer, boolean, array, and object arguments, and the `$ref` test verifies that root definitions remain available.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Fixed in the current head. The grammar emits each required key exactly once in the schema’s canonical `required` order, then permits only complete declared optional-argument tags. Optional alternatives exclude required keys, so a later duplicate required argument can no longer overwrite the constrained value.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Fixed in the current head. Required arguments now use one deterministic sequence in the schema’s canonical `required` order. The permutation expansion and six-field loose fallback are gone, so grammar size is linear in the number of required arguments and wider schemas retain required-field enforcement.

parsers/v1/src/tool_calling/structural_tag/builder.rs
```diff
@@ -82,33 +82,27 @@ pub(crate) fn resolve_tools_to_include<'a>(
...
-/// Resolve one tool's argument schema using Dynamo's generic schema policy.
+/// Whether a tool's declared parameter schema should constrain its arguments.
 ///
-/// In [`StructuralTagSchemaMode::Auto`], only `strict: true` selects the
-/// declared schema. Model-native builders may intentionally follow different
-/// upstream semantics and should use a model-specific policy helper.
+/// [`StructuralTagSchemaMode::Auto`] treats an omitted `strict` field as
+/// schema-enforced. Only an explicit `strict: false` opts out. Global strict
+/// mode overrides that per-tool opt-out.
+pub(crate) fn uses_declared_tool_schema(tool: &ToolDefinition, strict_schema: bool) -> bool {
```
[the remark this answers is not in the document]
indrajit96: This behavior is intentional in the final coordinated design with ai-dynamo/dynamo#13654. In `Auto` schema mode, omitted `strict` and `strict:true` use the declared schema; only explicit `strict:false` opts out, while global `strict` mode overrides that opt-out. The PR description and focused generic tests now document and pin this cross-builder policy, so I have not reverted it to a Kimi-only predicate.

parsers/v1/src/tool_calling/structural_tag/tests.rs
```diff
@@ -648,3 +650,29 @@ fn parser_map_structural_tag_smoke() {
         assert_eq!(tag["type"], "structural_tag", "'{name}'");
     }
 }
+
+#[test]
```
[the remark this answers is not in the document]
indrajit96: Fixed. Removed `structural_tag_builder_parser_set_is_stable`; the existing parser-map smoke coverage and focused K3 builder tests remain.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,10 +416,202 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
...
+        begin: format!(
+            "{OPEN}argument key=\"{}\" type=\"{xtml_type}\"{SEP}",
+            escape_attr(key)
+        ),
+        content: Box::new(content),
+        end: ARGUMENT_CLOSE.to_string(),
+    }))
+}
+
+fn loose_auto_arguments(required_keys: &[&str]) -> Format {
```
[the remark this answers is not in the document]
indrajit96: Fixed in `72a9a41`. The unrestricted `AnyText` tail was removed. After required arguments, the grammar now permits only repetitions of complete declared optional-argument tags, so it cannot generate a raw body segment that the K3 parser would reject.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -402,7 +810,310 @@ mod tests {
...
+            argument["content"]["json_schema"]["$ref"],
+            "#/$defs/identifier"
+        );
+        assert_eq!(
+            argument["content"]["json_schema"]["$defs"]["identifier"]["type"],
+            "integer"
+        );
+    }
+
+    #[test]
```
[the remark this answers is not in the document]
indrajit96: Fixed in `72a9a41`. Removed `auto_wide_required_schema_uses_canonical_order` and kept `auto_required_arguments_use_canonical_schema_order` as the focused coverage.

parsers/v1/src/tool_calling/structural_tag/format.rs
```diff
@@ -88,6 +88,8 @@ pub struct TriggeredTagsFormat {
     pub tags: Vec<TagFormat>,
     pub at_least_one: bool,
     pub stop_after_first: bool,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub excludes: Vec<String>,
```
keivenchang, line 92: [P1] Adding this required public field breaks downstream struct literals. `TriggeredTagsFormat` is exposed through the public `structural_tag::format` module, so a consumer constructing the existing four-field struct will fail to compile after upgrading from 8.2.1, while this PR does not change the crate's major version. Preserve the existing construction API, or make this an intentional major-version change.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,10 +416,210 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
...
 
+fn auto_argument_format(
+    key: &str,
+    schema: &Value,
+    root: &Value,
+    root_defs: &Map<String, Value>,
+    require_nonempty_string: bool,
+) -> Option<Format> {
+    let xtml_type = single_xtml_type(schema, root)?;
+    let content = if xtml_type == "string" {
```
keivenchang, line 427: [P1] This branch turns every single-type string schema into an unconstrained raw regex, so `enum`, `const`, `pattern`, and `maxLength` are ignored. It also uses `+` for every required string, rejecting valid schemas such as `{"type":"string","minLength":0}` when the model emits an empty value. Required controls presence of the argument; it does not imply a non-empty string. Build the raw XTML content from the string schema constraints and use `+` only when the schema requires at least one character.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -56,6 +61,219 @@ fn one_of(elements: Vec<Format>) -> Format {
...
+        vec![JsonType::String]
+    } else if NUMBER_KEYWORDS.iter().any(|key| schema.contains_key(*key)) {
+        vec![JsonType::Number]
+    } else {
+        JSON_TYPES.to_vec()
+    }
+}
+
+fn single_xtml_type(schema: &Value, root: &Value) -> Option<&'static str> {
+    let types = schema_types(schema, root, &HashSet::new());
```
keivenchang, line 273: [P2] `schema_types` preserves `Integer` and `Number` as separate values, but both map to the same XTML type (`number`) in `xtml_name`. Therefore a schema such as `anyOf: [{"type":"integer"}, {"type":"number"}]` reaches `single_xtml_type` with two JSON types and falls back to the permissive argument grammar even though both alternatives are representable by the same XTML tag type. Deduplicate after the XTML mapping, or retain a schema-aware numeric format instead of dropping required-field enforcement.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,10 +416,210 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
...
+
+fn auto_arguments_block(tool: &ToolDefinition, strict_schema: bool) -> Format {
+    if !super::builder::uses_declared_tool_schema(tool, strict_schema) {
+        return arguments_block(None);
+    }
+
+    let Some(parameters) = tool.parameters.as_ref().and_then(Value::as_object) else {
+        return arguments_block(None);
+    };
+    let root = Value::Object(parameters.clone());
```
keivenchang, line 507: [P3] This clones the complete parameter schema once per tool only to provide a root for local-reference inference. Request-time grammar construction can duplicate large `$defs` trees for every tool, increasing allocation and serialization cost. Pass the existing parameter value as the root, or make the inference walker accept the borrowed object without cloning it.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,10 +416,210 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
...
+    root: &Value,
+    root_defs: &Map<String, Value>,
+) -> Option<Format> {
+    let arguments = properties
+        .iter()
+        .filter(|(key, _)| !required_keys.contains(&key.as_str()))
+        .filter_map(|(key, schema)| {
+            matches!(schema, Value::Bool(_) | Value::Object(_))
+                .then(|| auto_argument_format(key, schema, root, root_defs, false))
+                .flatten()
```
rmccorm4, line 475: [P2] Preserve optional arguments with union types. `auto_argument_format` returns `None` when the schema cannot be reduced to one type, and this `filter_map` silently removes the property from the grammar. For an optional property with `type: ["string", "null"]`, the generated Auto grammar rejects both a valid string argument and a valid null argument, while the base grammar accepts both (reproduced with XGrammar 0.2.1 and 0.2.3 on this head). Represent the alternatives or retain a parser-compatible fallback for that declared key instead of making the argument impossible to emit.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,10 +416,210 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
...
+        })
+    } else {
+        // Non-string K3 argument bodies are JSON. Preserve the complete
+        // property schema instead of using the string-only regex, and carry
+        // root definitions so local references remain resolvable.
+        let mut embedded = schema.as_object()?.clone();
+        for (definition_key, definitions) in root_defs {
+            embedded
+                .entry(definition_key.clone())
+                .or_insert_with(|| definitions.clone());
```
rmccorm4, line 445: [P2] Preserve the original root when embedding local references. Type inference resolves `$ref` against the full tool parameters, but this code then makes the property schema a new root and copies only `$defs`/`definitions`. For example, with `properties: {"other": {"type": "integer"}, "value": {"$ref": "#/properties/other"}}` and `required: ["value"]`, building succeeds but XGrammar compilation fails with `Cannot find field properties in #/properties/other`. The base grammar compiles and accepts the call; this head fails on both XGrammar 0.2.1 and 0.2.3. Resolve or rebase supported local references before extracting the property schema.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Thanks for clarifying. I re-reviewed these helpers after completing the correctness fixes. Although parts look generic, their current semantics are coupled to K3’s XTML representation: K3 marker syntax, raw string bodies, and the mapping of JSON `integer`/`number` to XTML `type="number"`. No other builder currently consumes the same abstraction, so moving them into `common.rs` would create a shared API without a demonstrated second use and broaden this PR. I’ve kept them local for now and would prefer to extract the genuinely shared pieces when another format needs the same behavior.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Clarifying my earlier response: named and required choices already use the existing mandatory structural-tag path. Named choice includes only the selected tool and requires its call; required choice includes the allowed tools and requires at least one call. Auto needs a different format because no tool call is also valid: it allows ordinary response text and switches into the constrained K3 tool grammar only when the tools channel begins. The focused tests cover the optional Auto suffix and mandatory named/required paths, so no additional code change is needed here.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Fixed in `992559b`. Cardinality is now checked after mapping and deduplicating XTML types, so `anyOf(integer, number)` produces one XTML `number` argument instead of falling back to the permissive grammar. `auto_numeric_union_uses_one_number_tag_with_the_declared_schema` covers this case.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Fixed in `992559b`. `auto_arguments_block` now borrows `tool.parameters` directly as the local-reference root instead of cloning the complete parameter schema once per tool. Only the derived schema fragment needed by XGrammar is materialized.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,10 +417,239 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
...
 
+fn auto_argument_format(
+    key: &str,
+    schema: &Value,
+    root: &Value,
+    root_defs: &Map<String, Value>,
+    require_nonempty_string: bool,
+) -> Option<Format> {
+    let xtml_type = single_xtml_type(schema, root)?;
+    let content = if xtml_type == "string" {
```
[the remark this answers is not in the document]
indrajit96: Fixed across `992559b` and `2cbdb54`. Auto string content is now derived from the declared schema for `enum`, `const`, `pattern`, and length constraints. `minLength` without `maxLength` uses an open-ended `{min,}` repetition, while required strings remain empty when the schema permits it.

For schemas combining `pattern` with length bounds, the pattern takes precedence, matching XGrammar’s own policy and avoiding an unsupported lookahead-based grammar. Focused regression tests cover both remaining edge cases.

parsers/v1/src/tool_calling/structural_tag/format.rs
```diff
@@ -88,6 +88,8 @@ pub struct TriggeredTagsFormat {
     pub tags: Vec<TagFormat>,
     pub at_least_one: bool,
     pub stop_after_first: bool,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
```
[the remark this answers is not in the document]
indrajit96: Fixed in `992559b`. The new required field was removed from the public `TriggeredTagsFormat` struct and all constructors. K3 Auto now expresses the optional tools suffix using existing format nodes, preserving downstream four-field struct literals.

parsers/v1/src/tool_calling/structural_tag/format.rs
indrajit96: Fixed in `992559b`. `TriggeredTagsFormat` retains its existing four-field public construction API. K3 Auto now composes the exclusion behavior from existing `Sequence`, `AnyText`, and `Optional<Tag>` nodes, avoiding a source-breaking required field.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Fixed across `992559b` and `2cbdb54`. Auto string arguments now preserve `enum`, `const`, `pattern`, `maxLength`, bounded `minLength`/`maxLength`, and `minLength` without a maximum. Required controls argument presence only, so schemas permitting an empty string no longer receive an unconditional `+` regex.

When `pattern` and length bounds coexist, the pattern takes precedence, matching XGrammar’s JSON-schema policy because XGrammar 0.2.3 does not support the lookahead needed for a general regex intersection. Focused tests cover enum values, patterns, bounded strings, allowed empty strings, `minLength`-only schemas, and pattern-plus-length schemas.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Fixed in `992559b`. JSON types are mapped to XTML types before deduplication and cardinality checking. Integer and number therefore collapse to one XTML `number` type while retaining the declared numeric schema.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Fixed in `992559b`. The original `tool.parameters` value is now borrowed as the reference root; the complete schema is no longer cloned solely for type inference.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Fixed in `992559b`. `auto_argument_format` now emits alternatives for every representable XTML type instead of returning `None` for multi-type schemas. An optional `type: ["string", "null"]` property therefore remains available with both string and null tag alternatives. `auto_optional_union_keeps_each_representable_xtml_type` covers the reproducer.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Fixed in `992559b`. The builder retains the original parameter schema as the reference root and recursively resolves supported local references before embedding the property schema. The exact `#/properties/other` reproducer now produces a resolved integer schema without a dangling `$ref`. `auto_required_ref_into_properties_is_resolved_from_the_original_root` covers it.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -56,17 +61,322 @@ fn one_of(elements: Vec<Format>) -> Format {
...
+        // XGrammar 0.2.3 regexes do not support lookahead, so a general
+        // intersection of `pattern` and length bounds cannot be expressed.
+        // Match XGrammar's JSON-schema policy: `pattern` takes precedence when
+        // both are present instead of silently discarding the pattern.
+        let anchored_start = pattern.starts_with('^');
+        let anchored_end = pattern.ends_with('$') && !pattern.ends_with("\\$");
+        let pattern = pattern
+            .strip_prefix('^')
+            .unwrap_or(pattern)
+            .strip_suffix('$')
```
rmccorm4: [P2] Preserve escaped literal dollars when removing regex anchors. `anchored_end` correctly recognizes that `^price\$` has no end anchor, but `strip_suffix('$')` still removes its literal dollar, leaving a trailing backslash that escapes the generated closing parenthesis. The resulting structural tag fails XGrammar compilation with `The parenthesis is not closed` on both 0.2.1 and 0.2.3; the base grammar accepts `price$`. Strip the suffix only when it is an actual anchor, accounting for backslash parity.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -163,6 +439,87 @@ fn permissive_argument_tag() -> TagFormat {
...
+fn resolve_local_refs(schema: &Value, root: &Value, seen_refs: &HashSet<String>) -> Option<Value> {
+    match schema {
+        Value::Array(values) => values
+            .iter()
+            .map(|value| resolve_local_refs(value, root, seen_refs))
+            .collect::<Option<Vec<_>>>()
+            .map(Value::Array),
+        Value::Object(object) => {
+            if let Some(reference) = object.get("$ref").and_then(Value::as_str) {
+                if seen_refs.contains(reference) {
```
rmccorm4: [P2] Preserve valid recursive schemas instead of dropping the argument. A repeated local reference is not necessarily invalid: an optional object argument with `properties: {"child": {"$ref": "#/$defs/node"}}` and `$defs.node = {"type": "object", "properties": {"next": {"$ref": "#/$defs/node"}}}` is a valid recursive schema. Returning `None` here propagates through `schema_for_xtml_type` and filters the entire optional argument out; the generated grammar rejects its valid `{"child":{}}` value, which the base accepts (XGrammar 0.2.1 and 0.2.3). Retain recursive references and their definitions rather than requiring complete inlining.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -163,6 +439,87 @@ fn permissive_argument_tag() -> TagFormat {
...
+            JsonType::Integer => "integer",
+            JsonType::Boolean => "boolean",
+            JsonType::Array => "array",
+            JsonType::Object => "object",
+            JsonType::Null => "null",
+        })
+        .collect::<Vec<_>>();
+    match json_types.as_slice() {
+        [] => None,
+        [json_type] => Some(serde_json::json!({"type": json_type})),
```
rmccorm4: [P2] Retain value constraints when narrowing union alternatives. This branch replaces the declared schema with a type-only schema. For a required property with `{"type":["integer","null"],"enum":[1,null]}`, the numeric alternative therefore accepts `42` despite `strict:true` (reproduced on XGrammar 0.2.1 and 0.2.3). The same path drops bounds and nested object/array constraints. Narrow the resolved schema to the selected XTML type while preserving its applicable constraints.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -194,10 +543,197 @@ fn arguments_block(parameters: Option<&Value>) -> Format {
     star(one_of(tags))
 }
 
+fn auto_argument_format(key: &str, schema: &Value, root: &Value) -> Option<Format> {
+    let alternatives = xtml_types(schema, root)
+        .into_iter()
+        .filter_map(|xtml_type| {
+            let content = if xtml_type == "string" {
+                string_content_format(schema)
```
rmccorm4: [P2] Resolve string schemas before building their content. Type inference follows references/composition, but the string branch passes the original schema directly to `string_content_format`, which reads only top-level enum/const/pattern/length fields. A property referencing a string enum `["safe"]` consequently accepts `unsafe`; `anyOf: [{"type":"string","enum":["safe"]},{"type":"null"}]` does too (XGrammar 0.2.1 and 0.2.3). Resolve references and derive the applicable string alternative before emitting raw XTML content so these constraints survive.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -163,6 +439,87 @@ fn permissive_argument_tag() -> TagFormat {
...
+                }
+                return Some(serde_json::json!({
+                    "allOf": [resolved, Value::Object(siblings)]
+                }));
+            }
+
+            object
+                .iter()
+                .map(|(key, value)| {
+                    resolve_local_refs(value, root, seen_refs).map(|value| (key.clone(), value))
```
rmccorm4: [P2] Do not resolve `$ref` inside literal schema data. Recursing through every object value also visits `const`, `enum`, defaults and examples, where `$ref` is ordinary data. For a required property with `{"type":"object","const":{"$ref":"literal"}}`, the resolver tries to resolve `literal`, returns `None`, and the Auto builder falls back to permissive arguments. Head accepts a string argument for this object constant while base rejects it (XGrammar 0.2.1 and 0.2.3). Traverse only schema-bearing keywords and preserve literal data unchanged.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Good catch. The builder now removes a trailing `$` only when it is an unescaped end anchor. `^price\$` retains its literal dollar. I added a regression test for this case.

Fixed in commit [829d5d0](https://github.com/ai-dynamo/frontend-crates/commit/829d5d02880360218b4ffcc929d44cdda39257da).

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Agreed. Repeated local references now remain in the embedded schema with their definitions, so a valid recursive optional argument is retained. I added tests for this case and for recursion through `#/properties/...`.

Fixed in commit [829d5d0](https://github.com/ai-dynamo/frontend-crates/commit/829d5d02880360218b4ffcc929d44cdda39257da).

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Fixed. Narrowing now preserves the applicable enum, bounds, and nested schema constraints instead of replacing the alternative with a type-only schema. The integer/null reproducer has a regression test.

Fixed in commit [829d5d0](https://github.com/ai-dynamo/frontend-crates/commit/829d5d02880360218b4ffcc929d44cdda39257da).

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Fixed. String content is built from the resolved, applicable schema alternative. The tests cover both a reference to the `safe` enum and a string/null union, and retain the `safe` constraint.

Fixed in commit [829d5d0](https://github.com/ai-dynamo/frontend-crates/commit/829d5d02880360218b4ffcc929d44cdda39257da).

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Agreed. Reference traversal now follows schema-bearing keywords only; `$ref` inside `const`, `enum`, defaults, or examples remains literal data. I added your object-constant reproducer as a regression test.

Fixed in commit [829d5d0](https://github.com/ai-dynamo/frontend-crates/commit/829d5d02880360218b4ffcc929d44cdda39257da).

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -163,57 +513,528 @@ fn permissive_argument_tag() -> TagFormat {
...
+            .is_some_and(|values| {
+                values
+                    .iter()
+                    .any(|value| has_unresolved_local_refs(value, root))
+            })
+    })
+}
+
+fn preserve_recursive_root_refs(mut schema: Value, root: &Value) -> Value {
+    if !has_unresolved_local_refs(&schema, &schema) {
```
rmccorm4: [P2] Preserve the original root for recursive `#` references. A reference can resolve in the extracted schema while pointing to the wrong node: `#` always resolves locally, so this early return skips the necessary rebasing. With parameters `{"type":"object","properties":{"value":{"type":"object","properties":{"parent":{"$ref":"#"}},"additionalProperties":false}},"additionalProperties":false}`, the argument value `{"parent":{"value":{"parent":{"value":{}}}}}` is valid but rejected, while `{"parent":{"value":{"parent":{"parent":{}}}}}` is invalid but accepted. Both results reproduce on XGrammar 0.2.1 and 0.2.3, with the expected validity independently checked using Draft202012Validator. Preserve the original reference scope even when a pointer happens to resolve inside the extracted fragment.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -56,96 +62,440 @@ fn one_of(elements: Vec<Format>) -> Format {
...
+    let Some(schema) = schema.as_object() else {
+        return Format::AnyText(AnyTextFormat {
+            excludes: vec![CLOSE.to_string()],
+        });
+    };
+    // XTML string arguments contain raw text rather than a JSON string, so they
+    // cannot use JsonSchemaFormat. Flatten compatible allOf branches before
+    // translating their string constraints to raw-text formats.
+    let merged_schema = merge_compatible_all_of(schema);
+    let schema = merged_schema.as_ref().unwrap_or(schema);
```
rmccorm4: [P2] Preserve intersections when composing string constraints. If `allOf` contains different enum arrays, `merge_compatible_all_of` returns `None` and this fallback leaves the constraints nested where the remaining string builder ignores them. For `allOf: [{"type":"string","enum":["safe","other"]},{"enum":["safe","third"]}]`, the grammar accepts `other` even though only `safe` is valid. The same composition problem occurs with `{"type":"string","minLength":1,"anyOf":[{"enum":["safe"]},{"enum":["other"]}]}`: the length-regex return at line 465 bypasses the union restrictions and accepts `unsafe`. Both cases reproduce on XGrammar 0.2.1 and 0.2.3 and fail independent JSON Schema validation. Compose the applicable constraints instead of selecting one branch or silently ignoring an unmerged intersection.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
```diff
@@ -56,96 +62,440 @@ fn one_of(elements: Vec<Format>) -> Format {
...
+                    let len = value.chars().count() as u64;
+                    min_len.is_none_or(|min| len >= min)
+                        && max_len.is_none_or(|max| len <= max)
+                        && const_value.is_none_or(|constant| *value == constant)
+                        && compiled_pattern.as_ref().is_none_or(|pattern| {
+                            pattern.as_ref().is_ok_and(|re| re.is_match(value))
+                        })
+                })
+                .collect::<Vec<_>>();
+            return one_of(
```
rmccorm4: [P2] Handle an empty enum intersection without emitting `Or([])`. The filters above can remove every value, but `one_of` serializes an empty vector as an invalid empty `or` format. For an optional property `{"type":"string","enum":["a"],"minLength":2}`, omitting the property gives a valid empty argument object; nevertheless, the whole structural tag fails compilation with `Or format must have at least one element` on XGrammar 0.2.1 and 0.2.3. Propagate that the property has no valid values so optional alternatives can be omitted, rather than making an otherwise satisfiable tool schema fail compilation.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: You’re right. Checking whether `#` resolves in the extracted argument was insufficient because it could resolve to the wrong root. Remaining local schema references now point to a private copy of the original parameter root. I added the nested-value regression and verified the valid and invalid outputs with XGrammar 0.2.3. Fixed in `d483f3977afd282688ad64b70cd8055166e3c31e`.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Agreed on both reproductions. The fix handles the finite string-enum cases: it intersects string enums in `allOf` and applies outer constraints to each union branch. The XGrammar 0.2.3 check accepts only `safe` for the `allOf` case; for the `anyOf` case it accepts `safe` and `other` and rejects `unsafe`. Fixed in `d483f3977afd282688ad64b70cd8055166e3c31e`.

parsers/v1/src/tool_calling/structural_tag/kimi_k3.rs
indrajit96: Good catch. When filtering leaves no valid string values, the builder now omits that optional argument instead of emitting `Or([])`. I added the `enum:["a"]` plus `minLength:2` regression and verified that the resulting structural tag compiles with XGrammar 0.2.3. Fixed in `d483f3977afd282688ad64b70cd8055166e3c31e`.

PR #204, opened 2026-09-01
feat(mm-preprocessor): [PR 1/N] crate skeleton 

sherlockwu:
`dynamo-mm-preprocessor` is a Rust replacement for the image preprocessing functions behind HF `AutoProcessor`, for LLM serving routers or engines. 

This is the first PR to layout mm-preprocessor in dynamo front end. This PR only layouts module and public interfaces, with no actual implementation.

**Start with [`mm-preprocessor/README.md`](https://github.com/sherlockwu/frontend-crates/blob/kan/mm-pr1-skeleton/mm-preprocessor/README.md)** — it covers:
1. the **boundary**: what HF ships (per-family processor math, placeholder expansion, position math) plus what a router and an engine must compute *identically* — config-file spec resolution, pixel-free token accounting, source fetching, content-hash identity. 
2. **where the crate sits** (§1): two diagrams — an engine's full pixel path, and a router's pixel-free slice for load accounting and cache-aware routing;
3. **the API, by consumer** (§2): when/API/in→out tables for a router (§2.1: `spec_from_model_dir`, `fetch_bytes`, `content_hash_u64`, `dimensions`, `num_media_tokens`) and an engine (§2.2: `process_item` → `layout`/`apply_layout` → `positions`), plus a worked example — one 100×76 image through the whole trait (§2.3);
4. the **Python-parity map** (§3): which HF/SGLang API each item mirrors, and at what exactness (most bitwise; the deliberate divergences called out);
5. the three-layer bit-exactness **testing** strategy (§4) and the **roadmap** (§5, incl. the video/audio layout reservation).

## Layout

```
mm-preprocessor/src/
  processor.rs     the model-family seam: trait + non_exhaustive carriers
  registry.rs      family selection: typed/JSON spec, or spec_from_model_dir straight
                   from the HF config files (AutoProcessor.from_pretrained parity)
  models/          one module per family; qwen_vl.rs first (GLM-4V and Kimi next)
  image/           decode + header-only dimension probe + bit-exact resize kernels
                   (PIL & ATen-u8) + transforms
  fetch.rs         (feature) media source resolution — shared by router and engine
  token_layout.rs  validating placeholder expansion (never retokenizes)
  execution.rs     how the kernels execute: inline until init_pool arms the rayon pool
```


mm-preprocessor/src/driver.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/// [`process`] with explicit [`ProcessOptions`].
+pub fn process_with(
+    family: &dyn MmFamilyProcessor,
+    input: MmInput,
+    tokenize: impl Fn(&str) -> Result<Vec<i32>, String>,
+    opts: &ProcessOptions,
+) -> Result<Output, String> {
+    let _ = (family, input, tokenize, opts);
+    // PR3, staged as:
```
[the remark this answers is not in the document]
sherlockwu: Moot as of 23e7100: driver.rs left the crate entirely — request orchestration moved to the serving engine's driver, mirroring the Python layering (see DESIGN.md "The boundary").

mm-preprocessor/src/fetch.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/// [`fetch_bytes_budgeted`] with explicit [`FetchOptions`].
+pub fn fetch_bytes_budgeted_with(
+    src: &str,
+    budget: &ByteBudget,
+    opts: &FetchOptions,
+) -> Result<Vec<u8>, String> {
+    let _ = (src, budget, opts);
+    // PR2: source-precedence dispatch (http/file/path/data:/bare base64),
```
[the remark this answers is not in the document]
sherlockwu: Moot as of 23e7100: fetch.rs left the crate — source resolution is engine territory (Python home: the engine's get_image_bytes).

mm-preprocessor/src/fetch.rs
```diff
@@ -0,0 +1,71 @@
...
+
+/// Cap on any single resolved payload — HTTP, file, or base64 — so no source
+/// form can exhaust memory.
+pub const MAX_FETCH_BYTES: u64 = 64 << 20;
+
+/// A byte allowance shared by every source of one request, charged as they
+/// stream, so concurrent fetches stop at their combined size rather than each
+/// stopping at [`MAX_FETCH_BYTES`].
+#[derive(Debug)]
+pub struct ByteBudget(#[allow(dead_code)] std::sync::atomic::AtomicU64); // read from PR2
```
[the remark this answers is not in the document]
sherlockwu: Moot as of 23e7100: fetch.rs left the crate.

mm-preprocessor/src/models/qwen_vl.rs
```diff
@@ -0,0 +1,143 @@
...
+    fn from(r: Resampler) -> Self {
+        match r {
+            Resampler::AtenU8 => resize::Resample::AtenU8,
+            Resampler::Pil => resize::Resample::Pil(resize::Filter::Bicubic),
+        }
+    }
+}
+
+pub struct QwenVlProcessor {
+    // PR3 adds: the spec, plus a per-channel u8 → normalized-f32 lookup table
```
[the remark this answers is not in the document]
sherlockwu: Done in 841282e — condensed to the lasting constraint (the two normalize forms differ on 128/256 u8 inputs; the spec selects which HF form to mirror), staging narration dropped.

mm-preprocessor/src/models/qwen_vl.rs
```diff
@@ -0,0 +1,143 @@
...
+    // that rounds as the mirrored HF processor rounds (the slow path rescales
+    // then normalizes; the fast path folds the rescale into mean/std first —
+    // the two differ on 128 of the 256 u8 inputs, so picking the wrong form
+    // silently costs bit-exactness).
+    #[allow(dead_code)] // read from PR3
```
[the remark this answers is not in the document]
sherlockwu: Done in 841282e — note dropped, the bare #[allow(dead_code)] stays until the field gains readers.

mm-preprocessor/src/models/qwen_vl.rs
```diff
@@ -0,0 +1,143 @@
...
+        let spec: QwenVlSpec =
+            serde_json::from_str(json).map_err(|e| format!("qwen_vl spec: {e}"))?;
+        Self::new(spec)
+    }
+}
+
+impl MmFamilyProcessor for QwenVlProcessor {
+    fn process_item(&self, media: &DecodedMedia) -> Result<ProcessedItem, String> {
+        let _ = media;
+        // PR3: smart_resize → resize (skipped when dims already match) →
```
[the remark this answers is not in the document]
sherlockwu: Done in 841282e — narration dropped; the todo! tag alone marks where the body lands.

mm-preprocessor/src/models/qwen_vl.rs
```diff
@@ -0,0 +1,143 @@
...
+        // fused normalize+patchify (HF flatten order) → feature tensor
+        // `[gh*gw, 3*tps*ps*ps]` + aux `image_grid_thw` + `Grid([1, gh, gw])`.
+        todo!("PR3: HF Qwen2VLImageProcessor equivalent")
+    }
+
+    fn layout(&self, input_ids: &[i32], items: &[Geometry]) -> Result<TokenLayout, String> {
+        let _ = (input_ids, items);
+        // PR3: `t*h*w / merge_size²` tokens per image, via
```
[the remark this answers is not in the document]
sherlockwu: Done in 841282e — dropped.

mm-preprocessor/src/execution.rs
```diff
@@ -0,0 +1,63 @@
...
+//!   inline on the calling thread. A server supplies concurrency across
+//!   requests and owns its own core budget (it may pin threads), so a library
+//!   that silently spawns its own pools would fight it.
+//!
+//! Results are identical either way — the fan-outs are order-preserving maps
+//! and writes into disjoint slices, never reductions.
+
+/// Size the crate's CPU pool before its first use. Idempotent — the first
+/// caller wins, and once the pool exists the size is fixed; zero is ignored.
+/// Never called: `min(available_parallelism, 8)`.
```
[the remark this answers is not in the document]
sherlockwu: Reworded in 841282e rather than deleted: the line documents the default pool size when init_pool is never invoked — a lasting contract, not skeleton state — but the "Never called:" phrasing read as status. Now: "When no caller ever sizes it, the pool defaults to min(available_parallelism, 8)."

mm-preprocessor/src/models/qwen_vl.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    pub fn from_spec_json(json: &str) -> Result<Self, String> {
+        let spec: QwenVlSpec =
+            serde_json::from_str(json).map_err(|e| format!("qwen_vl spec: {e}"))?;
+        Self::new(spec)
+    }
+}
+
+impl MmFamilyProcessor for QwenVlProcessor {
```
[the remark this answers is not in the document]
sherlockwu: Deliberate at skeleton stage, keeping todo!: the crate is publish = false (mirrored in release-plz.toml), so no consumer can link this build — the panic is unreachable outside the workspace. Returning Err instead would masquerade as a legitimate request rejection that an engine would surface to its client as a 400; a loud panic is the honest marker for an unimplemented body. Bodies land in PR 3, which also flips the crate to publishable.

mm-preprocessor/src/registry.rs
```diff
@@ -0,0 +1,40 @@
...
+//! consumer selects one by its typed [`PipelineSpec`] or by serializing a
+//! spec (`{"family": ..., resolved processor params}`).
+
+/// The resolved parameters of one family pipeline — the typed form of the
+/// consumer-side spec, one variant per family arm. A consumer builds it
+/// directly, or reaches it through [`pipeline_from_spec`], where the `family`
+/// key selects the variant.
+#[derive(Clone, Debug, serde::Deserialize)]
+#[serde(tag = "family", rename_all = "snake_case")]
+#[non_exhaustive]
```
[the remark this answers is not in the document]
sherlockwu: This is not what #[non_exhaustive] does on an enum: it forbids exhaustive *matching* outside the defining crate, not construction of existing variants (RFC 2008). Downstream code constructing ProcessorSpec::QwenVl(spec) compiles fine — the SGLang integration does exactly that against this definition. A variant-level #[non_exhaustive] would block construction, but none is used here.

mm-preprocessor/src/models/qwen_vl.rs
```diff
@@ -0,0 +1,143 @@
...
+/// One media item's placement for M-RoPE: inclusive token range + patch grid.
+pub struct MropeItem {
+    pub start: u32,
+    pub end: u32,
+    pub grid: [u32; 3],
+}
+
+/// Resolved processor params, deserialized from the consumer-side spec JSON
+/// (unknown fields like `family` are ignored here).
+#[derive(Clone, Debug, serde::Deserialize)]
```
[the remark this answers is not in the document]
sherlockwu: Deliberate, no change: QwenVlProcessor::from_spec_json parses QwenVlSpec directly from the same spec JSON the tagged ProcessorSpec consumes — including the "family" key — so deny_unknown_fields would break that entry point (and serde documents deny_unknown_fields as unsupported with internally tagged enums for the typed path). The reject-unknown-knobs requirement is enforced where unknown knobs can exist: in the consumer's resolution step (SGLang's gate maps live HF processor state onto this spec and refuses anything unrecognized). By the time JSON reaches this crate it is already-resolved parameters, not raw HF config. The struct's doc comment states the tolerance explicitly.

mm-preprocessor/src/processor.rs
```diff
@@ -0,0 +1,150 @@
...
+//! flow.** A family never sees the request loop, the thread pool, or the
+//! failure protocol — it implements [`MmFamilyProcessor`], turning decoded
+//! media into named tensors and describing its prompt geometry as a
+//! [`TokenLayout`] value. The engine applies the layout mechanically
+//! ([`crate::token_layout::apply_layout`]), so expansion, per-item offsets,
+//! and position inputs all derive from one declarative structure.
+//!
+//! The carriers below are `#[non_exhaustive]` so a new family can grow them
+//! without a breaking change. Known future needs, validated against the
+//! GLM-4V and Kimi K2.5/K3 Python processors:
```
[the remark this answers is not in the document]
sherlockwu: Trimmed in 841282e: the module doc keeps the non_exhaustive growth rationale in one sentence and points at DESIGN.md §6 for the family-specific plan.

mm-preprocessor/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+//!
+//! Errors are `Result<T, String>` throughout: every `Err` is a
+//! request-rejection message a serving engine surfaces to its client, not a
+//! recoverable taxonomy.
+//!
+//! The crate reads no environment variables and, without the `parallel`
+//! feature, owns no threads; pool sizing (`par::init_pool`, feature
+//! `parallel`) is explicit configuration.
+//!
+//! This is the skeleton stage: signatures are final, bodies land with the
```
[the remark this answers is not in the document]
sherlockwu: Done in 841282e — dropped from lib.rs; README/DESIGN.md carry the status.

KrishnanPrash:
/ok to test [f494b3c](https://github.com/ai-dynamo/frontend-crates/pull/204/commits/f494b3c57767d598278579c24ae4f6b7c8450fbc)

mm-preprocessor/src/processor.rs
```diff
@@ -0,0 +1,143 @@
...
+pub trait MmFamilyProcessor: Send + Sync {
+    /// Modalities beyond images this family accepts. Default: images only.
+    fn capabilities(&self) -> Capabilities {
+        Capabilities::default()
+    }
+
+    /// Tokens one image will occupy in the expanded prompt, from its source
+    /// dimensions alone — no pixel work (HF's `_get_num_multimodal_tokens`).
+    /// Lets routers account prompt length before fetching past the header.
+    fn num_media_tokens(&self, width: usize, height: usize) -> Result<usize, String>;
```
yechank-nvidia: Despite the name, this signature is specifically num_image_tokens(width, height). Video may require frame count, timestamps, or sampling metadata, while audio requires duration/sample-rate metadata.

Could this accept modality-specific metadata, or move image token estimation to an image-specific optional interface?

mm-preprocessor/src/processor.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub segments: Vec<Segment>,
+}
+
+/// Modalities a family accepts; a serving engine rejects anything a family
+/// does not declare.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct Capabilities {
+    pub video: bool,
+    pub audio: bool,
+}
```
yechank-nvidia:
```suggestion
  pub struct Capabilities {
      image: bool,
      video: bool,
      audio: bool,
  }

  impl Capabilities {
      pub fn supports(&self, modality: Modality) -> bool {
          match modality {
              Modality::Image => self.image,
              Modality::Video => self.video,
              Modality::Audio => self.audio,
          }
      }
  }

```

Could image support be explicit here? The current shape assumes every processor supports images, so it cannot represent an audio-only processor. 

At minimum, adding image would make the modalities symmetric; alternatively,
Capabilities could expose supports(Modality) while keeping its internal representation private.

mm-preprocessor/src/processor.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// One processed media item: the primary feature tensor, named auxiliary
+/// tensors, and the geometry the family's own `layout`/`positions` hooks need.
+pub struct ProcessedItem {
+    /// The model's feature tensor for this item (qwen: `pixel_values`).
+    pub feature: Tensor,
+    pub aux: NamedTensors,
+    pub geometry: Geometry,
+}
```
yechank-nvidia:
```suggestion
 pub struct ProcessedItem {
      pub modality: Modality,
      pub feature_token_count: usize,

      pub feature: Tensor,
      pub aux: Vec<NamedTensor>,

      pub geometry: Option<Geometry>,
  }
  
  // and then
  
fn layout(
      &self,
      input_ids: &[u32],
      items: &[ProcessedItem],
  ) -> Result<TokenLayout>;
```

DecodedMedia carries the modality, but ProcessedItem drops it and requires every processed item to have Geometry. That may force mixed image/video requests to recover modality from tensor names or grids, and it gives audio models a geometry they may not need.

Could we minimally preserve modality and the computed feature_token_count here, and make geometry optional? Image/video processors using M-RoPE could return a grid, while audio processors using ordinary 1-D positions could return None.

mm-preprocessor/src/processor.rs
yechank-nvidia:
```suggestion
#[non_exhaustive]
  pub enum MediaMetadata {
      Image {
          width: u32,
          height: u32,
      },
      Video {
          width: u32,
          height: u32,
          num_frames: u32,
      },
      Audio {
          num_samples: u64,
          sample_rate: u32,
      },
  }

  fn num_media_tokens(
      &self,
      media: &MediaMetadata,
  ) -> Result<usize>;
```

Despite the name, this signature is specifically num_image_tokens(width, height). Video may require sampled frame count or timing metadata, while audio requires sample count/rate.

Could this accept one lightweight, typed metadata enum instead of decoded media—for example image dimensions, sampled video dimensions/frame count, or audio sample count/rate? That keeps the router path pixel-free without growing one optional trait method per modality.

mm-preprocessor/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+//! driver, as on the Python path; the README maps the boundary.
+//!
+//! Bit-exactness is the contract: the resize kernels ([`image::resize`]) and
+//! each family's normalize/patchify reproduce the mirrored HF processor's
+//! arithmetic exactly, so an engine can swap this crate in without output
+//! drift.
+//!
+//! Errors are `Result<T, String>`: every `Err` is a human-readable
+//! preprocessing failure for the consumer to surface, not a recoverable
+//! taxonomy.
```
yechank-nvidia:
```suggestion
#[derive(Debug)]
  #[non_exhaustive]
  pub enum MmError {
      Unsupported(String),
      InvalidInput(String),
      LimitExceeded(String),
      Internal(String),
  }

  pub type Result<T> = std::result::Result<T, MmError>;
```

 The consumer cannot own failure policy after the crate erases whether an error is an unsupported configuration, invalid media, resource limit, or internal failure.

Could this use a small #[non_exhaustive] error enum—for example Unsupported, InvalidInput, LimitExceeded, and Internal—while preserving a display message/source? 

SGLang can still return error.to_string() to its client, while Dynamo and other frontends can select fallback or rejection policy without matching strings.

mm-preprocessor/src/fetch.rs
```diff
@@ -0,0 +1,69 @@
...
+impl Default for FetchOptions {
+    fn default() -> Self {
+        Self {
+            timeout: std::time::Duration::from_secs(3),
+        }
+    }
+}
+
+/// Resolve one string-typed media source into raw encoded bytes.
+pub fn fetch_bytes(src: &str) -> Result<Vec<u8>, String> {
```
yechank-nvidia: Is fetch_bytes intended to be called with untrusted URLs from API requests? If so, what prevents access to local files/private IPs and enforces redirect, size, and timeout limits? 

Dynamo already has its own protected fetcher, so I think the processor should continue to accept resolved bytes and this helper should remain explicitly optional.

mm-preprocessor/src/execution.rs
```diff
@@ -0,0 +1,65 @@
...
+
+/// Arm the crate's CPU pool: from the first call on, the helpers below fan
+/// out on it instead of running inline. `threads == 0` picks the default size
+/// `min(available_parallelism, 8)`. Idempotent — the first caller wins, and
+/// once armed the size is fixed.
+#[cfg(feature = "parallel")]
+pub fn init_pool(threads: usize) {
+    let _ = threads;
+    todo!("arm the OnceLock-backed rayon pool")
+}
```
yechank-nvidia, line 32: Just to confirm the intended execution ownership: is mm-preprocessor expected to own this private, OnceLock-backed Rayon pool, or should each host application—such as Dynamo or the SGLang Rust API server—own its worker pool and invoke the processor synchronously?
 
The inline default seems compatible with host-owned execution, while opting into the crate pool could introduce nested parallelism. If crate ownership is intentional, how should a second init_pool call with a different thread count be handled?

mm-preprocessor/README.md
```diff
@@ -0,0 +1,273 @@
...
+SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# dynamo-mm-preprocessor
+
+Model-family multimodal preprocessing for LLM inference routers/engines — a
+Rust replacement for the image pipelines behind HF `AutoProcessor`: per-family
+decode → resize → normalize → patchify, prompt placeholder expansion, and
+position math (M-RoPE), all **bit-exact** against the mirrored HF processor.
```
yechank-nvidia, line 11: I checked the reference boundary: HF's Qwen processor emits the expanded tokens, pixel inputs, grids, and modality
token types, while its model-side get_rope_index computes M-RoPE. 

SGLang deliberately moves that computation into its multimodal processor, so keeping this pure helper here seems reasonable.

Could the README clarify that positions() is optional model-input preparation beyond strict AutoProcessor parity, and that engines with native position handling may bypass it?

mm-preprocessor/src/execution.rs
sherlockwu: Host applications normally own scheduling and invoke the processor synchronously; they should leave the crate pool unarmed to avoid nested parallelism. The crate-owned pool remains an explicit opt-in for callers with little host-side concurrency, such as Python bindings. Repeating the resolved thread count is idempotent; requesting a different count returns `MmError::InvalidInput`.

mm-preprocessor/src/lib.rs
sherlockwu: Agreed. I replaced `Result<T, String>` with a crate `Result<T>` using non-exhaustive `MmError` categories: `Unsupported`, `InvalidInput`, `LimitExceeded`, and `Internal`. Display messages and wrapped sources are preserved.

mm-preprocessor/src/fetch.rs
sherlockwu: Good point. This helper remains feature-gated and off by default, and is now explicitly documented as a trusted-source compatibility helper—not an API-request security boundary. Frontends handling untrusted URLs should use their protected fetcher and pass resolved bytes into this crate.

mm-preprocessor/src/processor.rs
sherlockwu: agreed!

mm-preprocessor/src/processor.rs
sherlockwu: Good point! applied the changes.

mm-preprocessor/src/processor.rs
sherlockwu: good point!

mm-preprocessor/README.md
sherlockwu: Agreed. The README now states that `positions()` is optional model-input preparation beyond strict `AutoProcessor` parity. It supports integrations such as SGLang that move M-RoPE calculation into preprocessing, while engines with native position handling may bypass it.

mm-preprocessor/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+
+pub type Result<T> = std::result::Result<T, MmError>;
+
+/// Media identity hash: blake3 truncated to its first 8 bytes, big-endian.
+/// One definition, so router cache-affinity keys and engine prefix-cache/dedup
+/// keys agree.
+pub fn content_hash_u64(data: &[u8]) -> u64 {
+    let _ = data;
+    todo!("blake3, first 8 bytes BE")
+}
```
krishung5: Could we clarify the intended MM routing identity contract here? The frontend and worker must use the same identity when constructing request-side routing tokens and worker cache events. Dynamo currently uses canonical decoded-media hashes for frontend-decoded requests and URL identity for passthrough, while this proposes BLAKE3(raw encoded bytes)[0..8]. Is this intended to replace those contracts? If so, how will it be versioned and coordinated across frontend/worker paths, and is exact encoded-file identity intentional rather than decoded-media identity?

mm-preprocessor/README.md
```diff
@@ -0,0 +1,287 @@
...
+| boot, per model | `registry::spec_from_model_dir` | config dir (hub download is the router's concern) → `ProcessorSpec`; `Err` = model unsupported |
+| boot, per model | `registry::build_processor` | `ProcessorSpec` → `Box<dyn MmFamilyProcessor>` |
+| per media part | consumer's protected fetcher | untrusted media source → raw bytes |
+| per trusted media part | `fetch::fetch_bytes` *(feature `fetch`)* | trusted media source → raw bytes; optional compatibility path |
+| per image part | `content_hash_u64` | bytes → `u64` cache-affinity key, identical to the engine's |
+| per image part | `image::decode::dimensions` | bytes → `MediaMetadata::Image` dimensions — header probe, no pixel decode |
+| per media part | `MmFamilyProcessor::num_media_tokens` | typed lightweight metadata → the item's expanded token count |
+
+Token counts give the expanded prompt length for routing; hashes drive
+prefix-cache-affinity routing and might travel downstream as `mm_hashes`.
```
krishung5: Is my understanding correct that the crate provides the media hash and expanded token count, while the consumer is responsible for replacing each media placeholder with the corresponding hash-derived padding tokens and constructing the final routing_token_ids passed to the router?

mm-preprocessor/src/processor.rs
```diff
@@ -0,0 +1,207 @@
...
+    Explicit(Vec<i32>),
+}
+
+/// One span of the expanded prompt.
+pub enum Segment {
+    /// Copy `src` (a range into the original ids) verbatim.
+    Text(std::ops::Range<usize>),
+    /// Media item `item`'s token span.
+    Media { item: usize, pattern: TokenPattern },
+}
```
krishung5, line 132: This looks sufficient for a single-token image placeholder. For example, with Qwen, [text, <image>, text], Text(0..1), Media(item=0), Text(2..3) implicitly means that Media consumes input_ids[1..2] and produces N feature tokens.

IIUC, for future support like video, the expansion might instead replace multiple source tokens and produce something like [<vision_start>, timestamp, feature × N, <vision_end>]. We might need a source range plus feature ranges/token roles in this structure for Media.

mm-preprocessor/README.md
sherlockwu: yes, correct

mm-preprocessor/src/lib.rs
sherlockwu: Good point. 

Added a `content_hash_canonical` for Dynamo’s hash in addition to the BLAKE3 encoded-byte hash used by SGLang’s Rust tokenizer manager when no mm_hash is supplied. 

For router/engine compatibility, whenever Dynamo supplies mm_hashes—from decoded media or passthrough URL identity—the engine uses those values without recomputing them. URL hashing can remain Dynamo-owned.

mm-preprocessor/src/processor.rs
sherlockwu: Good point! Extended Media struct.

krishung5:
/ok to test [c68ac39](https://github.com/ai-dynamo/frontend-crates/pull/204/commits/c68ac394c293b15815fb95b25bf6b971b9c15ccc)

krishung5:
Thanks @sherlockwu! Overall LGTM, one remaining question would be, I see that here we are splitting the usage into two consumers. Did we ever consider doing all the preprocessing on the frontend instead, and let the frontend transfer the preprocessed item to the engine via NIXL? Just curious on your thoughts here.

sherlockwu:
> Thanks @sherlockwu! Overall LGTM, one remaining question would be, I see that here we are splitting the usage into two consumers. Did we ever consider doing all the preprocessing on the frontend instead, and let the frontend transfer the preprocessed item to the engine via NIXL? Just curious on your thoughts here.

Thanks @krishung5 ! Yes, that's a valuable approach. The "split" only reflects today's deployment. Frontend could run the full pixel path and (somehow efficiently) ship tensors to engines. 

Two caveats:
1. Engines can use GPU-assisted mm preprocessing (NVJPEG, ...) that a CPU-only frontend can't match.
2. I am very interested in some measurements around how quickly preprocessed tensors can be transferred (because they are much larger).

krishung5:
/ok to test [cc9214b](https://github.com/ai-dynamo/frontend-crates/pull/204/commits/cc9214bca47c3368d66a6b1fffce528ae24684e4)

krishung5 (approved):

krishung5:
> > Thanks @sherlockwu! Overall LGTM, one remaining question would be, I see that here we are splitting the usage into two consumers. Did we ever consider doing all the preprocessing on the frontend instead, and let the frontend transfer the preprocessed item to the engine via NIXL? Just curious on your thoughts here.
> 
> Thanks @krishung5 ! Yes, that's a valuable approach. The "split" only reflects today's deployment. Frontend could run the full pixel path and (somehow efficiently) ship tensors to engines.
> 
> Two caveats:
> 
> 1. Engines can use GPU-assisted mm preprocessing (NVJPEG, ...) that a CPU-only frontend can't match.
> 2. I am very interested in some measurements around how quickly preprocessed tensors can be transferred (because they are much larger).

Makes sense to me! I think using NIXL + descriptor ideally should have quick transfer but would have to do some measurements for sure. Approved this PR, thanks again for the work on this!

mm-preprocessor/src/models/qwen_vl.rs
```diff
@@ -0,0 +1,140 @@
...
+pub struct QwenVlProcessor {
+    // Normalization must match the mirrored HF processor's arithmetic form:
+    // the slow path rescales then normalizes, the fast path folds the rescale
+    // into mean/std, and the two differ on half the u8 inputs.
+    #[allow(dead_code)]
+    spec: QwenVlSpec,
+}
+
+impl QwenVlProcessor {
+    pub fn new(spec: QwenVlSpec) -> Result<Self> {
```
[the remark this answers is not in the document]
sherlockwu: Fixed in fab9abf. QwenVlProcessor::new now rejects zero/inverted pixel bounds and non-finite or non-positive image_std values. Added focused unit tests covering each invalid case.

mm-preprocessor/src/models/qwen_vl.rs
```diff
@@ -0,0 +1,200 @@
...
+            return Err(MmError::invalid_input(
+                "qwen_vl spec: sizes must be positive",
+            ));
+        }
+        if spec.min_pixels == 0 || spec.min_pixels > spec.max_pixels {
+            return Err(MmError::invalid_input(
+                "qwen_vl spec: min_pixels must be positive and no greater than max_pixels",
+            ));
+        }
+        if spec
```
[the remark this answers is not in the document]
sherlockwu: Fixed in 099d5ce. The constructor now rejects non-finite image_mean values, with focused NaN and infinity coverage.

mm-preprocessor/src/models/qwen_vl.rs
```diff
@@ -0,0 +1,200 @@
...
+            assert!(matches!(
+                QwenVlProcessor::new(spec),
+                Err(MmError::InvalidInput { .. })
+            ));
+        }
+    }
+
+    #[test]
+    fn rejects_invalid_image_std() {
+        for invalid_std in [0.0, -1.0, f32::NAN, f32::INFINITY, f32::NEG_INFINITY] {
```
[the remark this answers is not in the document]
sherlockwu: Fixed in 099d5ce. Removed the redundant negative-infinity image_std case while retaining coverage for zero, negative, NaN, and infinite values.

krishung5:
/ok to test [099d5ce](https://github.com/ai-dynamo/frontend-crates/pull/204/commits/099d5cec8ed57234603da51c6acb3cdc39d24a02)

PR #211, opened 2026-09-04
feat(protocols): add Moonshot Kimi chat request fields

luckyq:
## Summary

Kimi Code CLI (Moonshot K3) sends request shapes `dynamo-protocols` could not represent, so Dynamo's frontend either rejected them or failed deserialization:

| Field | Status before | Change |
|---|---|---|
| `prompt_cache_key`, `safety_identifier` (OpenAI-standard) | not on `CreateChatCompletionRequest` | added as `Option<String>` |
| `{"role":"system","tools":[…]}` with no `content` (K3 dynamic tool loading) | deserialization error — `content` required | `ChatCompletionRequestSystemMessage` now defined locally: `content` defaults when omitted, `tools: Option<Vec<ChatCompletionTool>>` added. Builder name, `From` impls, and Rust field type are preserved |
| `tools[].type: "builtin_function"` (`$web_search`) | only `function` accepted | `ChatCompletionToolType::BuiltinFunction` |
| assistant `partial: true` (prefill / continue final message) | silently dropped | `partial: Option<bool>` on `ChatCompletionRequestAssistantMessage` |

All additions are optional and skip-serialized when absent, so existing wire output is unchanged. Consumers building `ChatCompletionRequestSystemMessage` / `ChatCompletionRequestAssistantMessage` by struct literal need `tools: None` / `partial: None`.

## Validation

- `cargo test -p dynamo-protocols` → all pass, incl. new `moonshot_kimi_request_fields_round_trip`
- `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt --check` clean
- `cargo test --workspace`: only `toolcalling_batch_parity` fails locally, because `flock` is unavailable on macOS (`conformance/tests/common/mod.rs:60`) — unrelated to this change


luckyq:
Closing in favor of #205, which covers the same protocol gap plus K3 renderer Partial Mode. The extras here (prompt_cache_key / safety_identifier on the request, ChatCompletionToolType::BuiltinFunction) can follow separately.

PR #213, opened 2026-09-04
feat: add v2 DeepSeek V4 and Kimi K3 unified parsers

keivenchang:
#### Overview:

This PR adds native v2 UnifiedParser support for DeepSeek V4 and Kimi K3 and updates the Unified conformance corpus for both families.

[Conformance report image](https://github.com/user-attachments/assets/344a6000-8945-4a1e-b843-fffaa0ebcce1)

image

#### Details:

- Adds native DeepSeek V4 and Kimi K3 unified parsing with projection-only legacy adapters, chunk-invariance coverage, malformed-input recovery, guided modes, and linear-scan tests.
- Expands current captures, historical overlays, fixture manifests, and the canonical v2 report.
- Renames Kimi K3 Unified cases to `k3-1` through `k3-8` and Gemma-specific cases to `g4-1` and `g4-2`; generic tool-calling cases remain unchanged.
- Fixes parallel Unified golden generation so concurrent CI test binaries read isolated completed trees.

#### Verification:

- Local parser library tests: `cargo test --locked -p dynamo-parsers-v2 --lib` — 557 passed.
- Local Unified conformance tests: `cargo test --locked -p dynamo-conformance-fixtures-v2 --test unified_schema_roundtrip --test unified_render --test unified_parity -- --nocapture` — 9, 11, and 12 passed.
- Frontend-crates CI: [Rust](https://github.com/ai-dynamo/frontend-crates/actions/runs/34263190390/job/102185987331), [conformance table](https://github.com/ai-dynamo/frontend-crates/actions/runs/34263190390/job/102186051362), [changes](https://github.com/ai-dynamo/frontend-crates/actions/runs/34263190390/job/102185987283), and [DCO](https://probot.github.io/apps/dco/) passed.
- Dynamo-vLLM E2E with the PR #213 image: [job results](http://ibhosale.nvidia.com/ci/jobs/20260908-205832-c839e64d/).
- Comparison run with the earlier Dynamo image: [job results](http://ibhosale.nvidia.com/ci/jobs/20260908-220152-059848e0/).
- Dynamo PR that pins and builds this parser revision: [PR #14489](https://github.com/ai-dynamo/dynamo/pull/14489).
- Published multi-architecture runtime image: [`1.5.0-ci-a488db4f435b5b681363e22eacddd8ffede243a2-vllm-runtime`](https://dynamoci.azurecr.io/v2/ai-dynamo/dynamo/manifests/1.5.0-ci-a488db4f435b5b681363e22eacddd8ffede243a2-vllm-runtime).
- E2E trigger page: [ibhosale.nvidia.com/ci/trigger](http://ibhosale.nvidia.com/ci/trigger).

| Model | Earlier image | PR #213 image |
| --- | ---: | ---: |
| Gemma 4 custom | 46 passed, 2 failed | 46 passed, 2 failed |
| Gemma 4 BFCL | 42 passed, 8 failed | 42 passed, 8 failed |
| DeepSeek V4 custom | 46 passed, 2 failed | 46 passed, 2 failed |
| Kimi K2.6 deployment | CrashLoopBackOff | CrashLoopBackOff |

These runs compare two images, not the v1 and v2 parser modes. The PR #213 image ran `deepseek_v4_flash_v2` with `DYN_ENABLE_EXPERIMENTAL_PARSERS_V2=1`; a valid parser comparison still requires `deepseek_v4_flash` and `deepseek_v4_flash_v2` on the same image. DeepSeek V4 BFCL was incomplete on the earlier image because generation failed for `multi_turn_miss_param_2`, so those BFCL results are not directly comparable.

`parsers/v2/src/unified/deepseek_v4.rs`, `parsers/v2/src/unified/kimi_k3.rs`, `parsers/v2/src/unified/mod.rs`, `conformance/utils/src/unified_taxonomy.py`, `conformance/utils/src/generate_conformance_table.py`, and `conformance/tests/common/mod.rs`

/coderabbit profile chill

Related to: [DIS-2851](https://linear.app/nvidia/issue/DIS-2851)

Related to: [DIS-2852](https://linear.app/nvidia/issue/DIS-2852)


parsers/v2/src/unified/kimi_k3.rs
```diff
@@ -0,0 +1,3238 @@
...
+    }
+
+    fn guided_grammar(&self) -> GuidedGrammar {
+        GuidedGrammar {
+            control_markers: ALL_MARKERS
+                .iter()
+                .flat_map(|marker| marker.variants())
+                .map(str::to_string)
+                .collect(),
+            invoke_start: CALL_OPEN.canonical.to_string(),
```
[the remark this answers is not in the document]
keivenchang: Fixed in 985b7486. Guided K3 now routes invoke-open detection through the family-owned boundary, so both canonical and spaced variants retain their matched length and strip the wrapper as one boundary. The regression covers every valid UTF-8 split, and the full dynamo-parsers-v2 library suite passes with 557 tests.

parsers/v2/src/tool_calling/dsml.rs
```diff
@@ -1,282 +1,429 @@
...
+use crate::unified::{UnifiedParser, UnifiedParserExt};
+
+pub(crate) const BLOCK_START: &str = "<｜DSML｜tool_calls>";
+pub(crate) const BLOCK_END: &str = "</｜DSML｜tool_calls>";
+pub(crate) const INVOKE_START_PREFIX: &str = "<｜DSML｜invoke name=\"";
+pub(crate) const INVOKE_END: &str = "</｜DSML｜invoke>";
+pub(crate) const PARAMETER_PREFIX: &str = "<｜DSML｜parameter name=";
+pub(crate) const PARAMETER_END: &str = "</｜DSML｜parameter>";
+
+/// Construct the only DeepSeek V4 DSML state machine.
```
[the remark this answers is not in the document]
keivenchang: Fixed in 0b5a802e. Removed the redundant constructor comment; the module-level documentation already owns the DSML state-machine explanation.

parsers/v2/src/unified/kimi_k3.rs
```diff
@@ -0,0 +1,3254 @@
...
+                .find(|at| text.ends_with(&marker[..*at]))
+        })
+        .max()
+        .unwrap_or_default();
+    text.len() - holdback
+}
+
+fn parse_tag_header(text: &str, open: Marker) -> Option<(Vec<(String, String)>, usize)> {
+    let open_len = open.prefix_len(text)?;
+    let (sep_at, sep_len) = SEP_MARKER.find(&text[open_len..])?;
```
indrajit96, line 1679: ### [P2] Avoid rescanning the whole incomplete call header

`parse_tag_header` searches from the beginning of the retained header on every `push`. While a native Kimi K3 call header is missing `<|sep|>`, `consume_call_open` repeats this search for every streamed chunk, making total work quadratic.

### Example

With one-character chunks and a deterministic examined-byte counter, increasing an incomplete header from 1,024 to 2,048 bytes increased scan work from 533,027 to 2,114,595 bytes—approximately 4× work for 2× input.

### Suggested fix

Replace the stateless `parse_call_header(&self.buffer)` retry with request-local incremental header state. Once `CALL_OPEN` matches, initialize a cursor after the opener and advance it only across newly appended bytes. The cursor must retain a partial separator match—or rescan at most the longest separator minus one byte—so both `<|sep|>` variants are recognized when split across chunks.

When the separator completes, parse the buffered attributes once and continue through the existing valid- or malformed-header handling. Reset this state whenever the header is consumed or dropped, the parser resynchronizes, or the request is finished/reset.

Add a native counterpart to `guided_incomplete_and_malformed_header_work_scales_linearly`: stream 4,096- and 8,192-byte incomplete headers one character at a time, count examined bytes, and assert approximately 2× rather than 4× growth. Also cover both separator spellings split at every byte boundary to ensure the cursor does not miss a cross-chunk marker.

parsers/v2/src/tool_calling/traits.rs
```diff
@@ -170,6 +170,11 @@ pub trait ToolParser: Send {
         Ok(ToolParseResult::default())
     }
 
+    /// Return the model-emitted identifier for a completed call, when the grammar names one.
+    fn tool_call_id(&self, _tool_index: usize) -> Option<&str> {
```
indrajit96, line 174: ### [P2] Forward model call IDs through every tool-parser layer

The default `None` makes missing delegation silent. `KimiK2ToolStreamParser` does not forward the ID already exposed by its scanner, and `DebugToolParser` does not forward its inner parser’s ID.

### Example

A Kimi K2 call containing `functions.get_weather:7` is emitted successfully, but `tool_call_id(0)` returns `None`. Kimi K3 normally returns `weather:1` for `index="2"`, but the debug wrapper returns `None`. Enabling `DYNAMO_PARSERS_DEBUG` therefore changes Kimi K3 behavior.

### Suggested fix

Keep the trait default for parsers whose grammars do not provide IDs, but explicitly delegate through the two wrappers that already contain an ID-aware parser:

```rust
// KimiK2ToolStreamParser
fn tool_call_id(&self, tool_index: usize) -> Option<&str> {
    self.scanner.tool_call_id(tool_index)
}

// DebugToolParser
fn tool_call_id(&self, tool_index: usize) -> Option<&str> {
    self.inner.tool_call_id(tool_index)
}
```

Add a Kimi K2 regression that parses a native `functions.get_weather:7` call and asserts both that the call is emitted and that `tool_call_id(0)` returns that exact ID. Add a debug-wrapper transparency regression using an ID-producing parser such as Kimi K3, asserting that wrapped and unwrapped parsers return the same ID and that an unknown index still returns `None`.

parsers/v2/src/unified/kimi_k3.rs
keivenchang: Fixed in bd4ae6ae. Native Kimi K3 call-header scanning now retains an incremental separator cursor, covering canonical and spaced separators across chunk boundaries without rescanning the retained header. The new regression covers 4,096/8,192-byte one-character streams, separator splits, EOF, reset, and reuse.

parsers/v2/src/tool_calling/traits.rs
keivenchang: Fixed in bd4ae6ae. KimiK2ToolStreamParser and DebugToolParser now delegate tool_call_id through their wrapped scanner/parser. Regressions cover native IDs, debug-wrapper transparency, unknown indexes, and Kimi K2 reset/reuse.

parsers/v2/src/unified/kimi_k3.rs
keivenchang: Follow-up fix in 220ea5fb. The ID normalizer now preserves zero, negative, and overflowing signed indices as raw IDs instead of producing invalid normalized values or panicking. Added all-split native regressions and retained the committed Kimi K3 capture path.

parsers/v2/src/tool_calling/traits.rs
keivenchang: Follow-up coverage is included in 4b89a6a8. The prior delegation fix remains in bd4ae6ae; the new commit adds the guided incomplete-header EOF/reset regression and leaves the adapter ID behavior unchanged.

parsers/v2/src/unified/mod.rs
```diff
@@ -1468,6 +1486,64 @@ struct GuidedState {
...
+    fn is_empty(&self) -> bool {
+        self.len == 0
+    }
+}
+
+#[cfg(test)]
+std::thread_local! {
+    static GUIDED_APPEND_REPLACEMENTS: std::cell::Cell<usize> = const { std::cell::Cell::new(0) };
+    static GUIDED_APPEND_RETAINED_PREFIX_COMPARISONS: std::cell::Cell<usize> = const { std::cell::Cell::new(0) };
+    static GUIDED_APPEND_COPIED_BYTES: std::cell::Cell<usize> = const { std::cell::Cell::new(0) };
```
rmccorm4: It seems like these 2 counters are never incremented anywhere - are they needed?
- GUIDED_APPEND_RETAINED_PREFIX_COMPARISONS
- GUIDED_APPEND_COPIED_BYTES

parsers/v2/src/tool_calling/dsml.rs
```diff
@@ -1,282 +1,428 @@
...
-    /// while we buffer the body and wait for `</｜DSML｜invoke>`. Nothing is
-    /// emitted until the close arrives; if it never does, the call is dropped.
-    open_invoke: Option<(usize, String)>,
+pub(crate) struct DsmlEmitter;
+
+#[cfg(test)]
+std::thread_local! {
+    static BOUNDARY_EXAMINED_BYTES: std::cell::Cell<usize> = const { std::cell::Cell::new(0) };
+    static BOUNDARY_RETAINED_PREFIX_COMPARISONS: std::cell::Cell<usize> = const { std::cell::Cell::new(0) };
+    static BOUNDARY_COPIED_BYTES: std::cell::Cell<usize> = const { std::cell::Cell::new(0) };
```
rmccorm4: It seems like these 2 counters are never incremented anywhere - are they needed?
- BOUNDARY_RETAINED_PREFIX_COMPARISONS
- BOUNDARY_COPIED_BYTES

parsers/v2/src/unified/mod.rs
```diff
@@ -48,10 +48,11 @@
 //! adopted without a translation layer. Where it diverges from the peer shape, the
 //! divergence is stated at the item.
 
-pub mod deepseek_v41;
```
rmccorm4, line 51: Please restore the Deepseek v4.1 parser code/tests from this PR - this PR shouldn't remove it, probably got removed in merge conflict resolution.

rmccorm4 (approved):
LGTM aside from the Deepseek v4.1 comments and small nits - please make sure to restore Deepseek v4.1 bits before merging

After merge, can you also work with @indrajit96 to run through these new v2 parsers through his e2e test suite and share the results with me on Slack?

parsers/v2/src/unified/mod.rs
keivenchang: Fixed in 024d9151 — removed the unused guided append counters and kept the live replacement-count coverage.

parsers/v2/src/tool_calling/dsml.rs
keivenchang: Fixed in 024d9151 — removed the unused DSML boundary counters and retained the examined-byte scaling test.

parsers/v2/src/unified/mod.rs
keivenchang: Fixed in 024d9151 — restored the DeepSeek V4.1 registry, corpus generation, and a PR-qualified current capture.

PR #222, opened 2026-09-10
feat(mm-preprocessor): [PR 2/N] qwen-vl support

sherlockwu:
Second PR for `dynamo-mm-preprocessor`, on top of the skeleton (#204). The skeleton laid out the modules and public signatures; this PR fills in every `todo!()` except `fetch`, and adds Qwen VL support end to end — the bit-exact image path, HF config resolution, and a golden replay that pins the whole composition byte-for-byte.

Reviewing commit by commit is easiest: primitives, then the family, then the goldens.

## What lands

**Image and layout primitives** (`image/`, `token_layout`, `execution`, `lib`)

| item | notes |
| --- | --- |
| `resize_rgb` | both resamplers: PIL's i32 fixed-point kernels, and ATen's uint8 antialias path with its per-axis i16 weight precision. Not interchangeable at the byte level, which is why the spec selects one |
| `decode_rgb` / `dimensions` | the PIL-accepted formats; >8-bit samples are refused rather than silently rescaled (PIL clips), and the probe reads the header only |
| `apply_layout` | expands the *already tokenized* prompt and validates the family's description while doing it: source covered exactly once in order, each item placed once, each item's `Feature` parts matching the embeddings it actually produced |
| `content_hash_bytes` / `content_hash_canonical_image` | pinned to the values SGLang's BLAKE3 fallback and Dynamo's `canonical_content_hash` produce — a router hashes and an engine looks the key up, so "stable here" is not the contract |
| `execution` | wires the `parallel` feature's rayon dep. Fan-out is inline on the caller until `init_pool` arms the crate-owned pool; two tests guard both halves from the outside (process thread count), so the guarantee survives refactors of the seam |

**Qwen VL** (`models/qwen_vl`) — Qwen2-VL / 2.5-VL / 3-VL / 3.5. `smart_resize` reproduces the Python reference including banker's rounding, and rejects the degenerate 0-side case Python leaves to PIL to blow up on. Normalization is a per-channel LUT folded into patchify, in the arithmetic form of the mirrored processor: the slow path rescales then normalizes, the fast path fuses the rescale into mean/std first, and the two disagree on 128 of the 256 u8 inputs. Token accounting (`num_media_tokens`, from header dimensions alone), placeholder layout, and image-only M-RoPE all derive from the same resolved grid.

**Spec resolution** (`registry::spec_from_hf_configs` / `spec_from_model_dir`) — the `AutoProcessor.from_pretrained` equivalent. `model_type` selects the family; `preprocessor_config.json` knobs are checked against what the pipeline actually models, and anything it cannot honor bit-exactly — a disabled stage, a foreign kernel, a knob this crate has never seen — refuses resolution instead of approximating. Qwen2/2.5 carry the pixel bounds at the top level and Qwen3 under `size.{shortest,longest}_edge`; both resolve, and the tests use trimmed copies of the real hub configs.

## Testing

26 unit tests plus three integration binaries. The golden replay drives the README §2.2 composition (`processor_from_spec` → `decode_rgb` → `process_item` → `layout` → `apply_layout` → `positions`) against fixtures generated by SGLang tooling from the live HF processor and `get_rope_index`, and compares `pixel_values`, `image_grid_thw`, `input_ids`, offsets, feature ranges, hashes and the M-RoPE positions/delta **bitwise** — this repo's CI has no Python or HF, and a systematic skew (wrong filter, wrong normalize rounding, wrong patch order) still yields plausible-looking tensors. Four cases cover both resamplers, both `smart_resize` branches, and multi-image.

Workspace-wide `cargo test`, `clippy -D warnings`, `fmt`, `machete` and `deny` are clean, and `mm-preprocessor` also passes with `--no-default-features`. `deny.toml` gains BSL-1.0 for `xxhash-rust`, the XXH3 crate Dynamo's canonical media identity uses.

## Still open

`fetch` remains signatures only — it is the last stub, and what keeps the crate at `publish = false`. Family coverage grows after that; GLM and Kimi are the validated candidates.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sherlockwu:
Closing in favor of three smaller PRs — this one was too large to review in one piece. Same content (plus extra tests), re-sliced along the seams so each piece stands on its own:

| PR | scope | size |
| --- | --- | --- |
| #225 | pixel primitives: the two bit-exact resamplers, decode + header probe, transforms, the fan-out seam | ~640 |
| #226 | prompt layout expansion, content-hash identity, HF config → spec resolution | ~590 |
| #227 | the qwen-vl family + bitwise golden replay | ~575 + fixtures |

#225 and #226 touch disjoint files and can be reviewed and merged in either order. #227 needs both, so it is a draft carrying their commits to keep CI green until they land.

Two things got stronger in the split: the resamplers are now pinned byte-for-byte against outputs captured from PIL and torchvision directly (on a noise fixture — on a smooth gradient the two quantizations agree, so a gradient would pass even with the wrong weight precision), and the transform primitives picked up unit tests. Previously both were only covered indirectly through the qwen goldens.

PR #223, opened 2026-09-10
fix(parser): preserve XML argument order

### LINKED ISSUE #221: v1 shared XML tool parser returns arguments in nondeterministic key order
## Summary

The shared v1 XML tool parser collects parameters in a `HashMap` and then serializes that map into `function.arguments`. Because `HashMap` iteration order is nondeterministic, the returned JSON keys may not match the order emitted by the model and can vary between processes.

This affects:

- `qwen3_coder`
- `nemotron_nano`, which uses the Qwen3-Coder configuration
- `minimax_m2`

The relevant code is:

https://github.com/ai-dynamo/frontend-crates/blob/2843ba9e3d3890b21ebc82410d9ea07d763b0500/parsers/v1/src/tool_calling/xml/parser.rs#L394-L409

The affected parser registrations are:

https://github.com/ai-dynamo/frontend-crates/blob/2843ba9e3d3890b21ebc82410d9ea07d763b0500/parsers/v1/src/tool_calling/parsers.rs#L63-L81

PR #220 fixes the same problem in the GLM-specific v1 parser, but the shared XML parser remains affected.

## Reproduction

Using this Qwen3-Coder output:

```xml
<tool_call>
<function=get_current_weather>
<parameter=city>
Dallas
</parameter>
<parameter=state>
TX
</parameter>
<parameter=unit>
fahrenheit
</parameter>
</function>
</tool_call>
```

Running the parser in six fresh processes produced four different argument orders, including:

```json
{"city":"Dallas","state":"TX","unit":"fahrenheit"}
{"city":"Dallas","unit":"fahrenheit","state":"TX"}
{"state":"TX","unit":"fahrenheit","city":"Dallas"}
{"unit":"fahrenheit","city":"Dallas","state":"TX"}
```

## Expected behavior

`function.arguments` should preserve the parameter order emitted by the model and produce stable output across executions.

For duplicate parameter names, the existing last-value-wins behavior should remain, while retaining the position of the first occurrence.

## Why this matters

Although JSON object key order does not change its semantic meaning, `function.arguments` is externally visible serialized output. Reordering creates inconsistent results and differs from parsers that preserve the model-emitted order, including the GLM behavior being addressed by #220.

## Suggested fix

Replace the `HashMap`-only collection with an explicitly ordered representation.

One option is:

- a `Vec<(String, ParsedValue)>` for serialization order;
- a `HashMap<String, usize>` mapping each key to its vector position, allowing duplicate values to be updated without an O(n) scan.

Add regression tests for:

- source-order preservation in `qwen3_coder`;
- source-order preservation in `minimax_m2`;
- duplicate parameters retaining their first position while the last value wins.

Testing `qwen3_coder` also covers the shared `nemotron_nano` configuration.

ITSMERNB:
## Summary
- preserve model-emitted parameter order in the shared v1 XML parser instead of serializing a `HashMap`
- keep duplicate-parameter semantics: last value wins without moving the key from its first position
- cover both Qwen3-Coder and MiniMax M2 order, plus duplicate replacement

## Reproduction
On current `main`, the same Qwen3-Coder input produced a fresh-process result such as:

`{"state":"TX","city":"Dallas","unit":"fahrenheit"}`

instead of the emitted order:

`{"city":"Dallas","state":"TX","unit":"fahrenheit"}`

## Validation
- `cargo test -p dynamo-parsers --locked`
- `cargo clippy -p dynamo-parsers --all-targets --all-features --locked -- -D warnings`
- `git diff --check`

Closes #221

### DIFF
--- a/parsers/v1/src/tool_calling/xml/parser.rs
+++ b/parsers/v1/src/tool_calling/xml/parser.rs
@@ -8,6 +8,7 @@ use std::collections::HashMap;
 
 use num_traits::ToPrimitive;
 use regex::Regex;
+use serde::ser::{Serialize, SerializeMap, Serializer};
 use serde_json::Value;
 use uuid::Uuid;
 
@@ -391,7 +392,8 @@ fn parse_tool_call_block(
         let param_config = get_arguments_config(function_name, tools);
 
         // Parse parameters from the function body.
-        let mut parameters: HashMap<String, ParsedValue> = HashMap::new();
+        let mut parameters: Vec<(String, ParsedValue)> = Vec::new();
+        let mut parameter_indices: HashMap<&str, usize> = HashMap::new();
 
         for param_cap in parameter_regex.captures_iter(function_body) {
             let param_name_raw = param_cap.get(1).map(|m| m.as_str().trim()).unwrap_or("");
@@ -401,12 +403,18 @@ fn parse_tool_call_block(
             if !param_name.is_empty() {
                 let parsed_value =
                     convert_param_value(param_value, param_name, &param_config, function_name);
-                parameters.insert(param_name.to_string(), parsed_value);
+                match parameter_indices.get(param_name).copied() {
+                    Some(index) => parameters[index].1 = parsed_value,
+                    None => {
+                        parameter_indices.insert(param_name, parameters.len());
+                        parameters.push((param_name.to_string(), parsed_value));
+                    }
+                }
             }
         }
 
         // Create tool call response.
-        let arguments_json = serde_json::to_string(&parameters)?;
+        let arguments_json = serde_json::to_string(&OrderedArguments(&parameters))?;
 
         let tool_call = ToolCallResponse {
             id: format!("call-{}", Uuid::new_v4()),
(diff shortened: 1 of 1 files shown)


parsers/v1/src/tool_calling/xml/parser.rs
```diff
@@ -423,6 +431,18 @@ fn parse_tool_call_block(
     Ok(results)
 }
 
+/// Serialize parsed parameters as a JSON object in the order emitted by the model.
+struct OrderedArguments<'a>(&'a [(String, ParsedValue)]);
```
rmccorm4: Could we move `OrderedArguments` into `xml/mod.rs` and import it in both parsers? This matches the implementation in `glm47_parser.rs`, so we can share it without adding a new module or exposing a public API.

rmccorm4 (approved):
LGTM other than one comment: https://github.com/ai-dynamo/frontend-crates/pull/223/changes#r4010598019 . Thanks for the contribution @ITSMERNB !

rmccorm4:
CC @indrajit96

parsers/v1/src/tool_calling/xml/parser.rs
ITSMERNB: Done. `OrderedArguments` now lives once in `xml/mod.rs` and both XML parsers import it. `cargo fmt --check`, `cargo test -p dynamo-parsers --locked` (826 passed, 4 ignored), and clippy with `-D warnings` all pass locally. Thanks for the suggestion.

rmccorm4:
/ok to test [0f1b029](https://github.com/ai-dynamo/frontend-crates/pull/223/commits/0f1b0297d22d63be4c9bf8656aeefc7338387953)

PR #230, opened 2026-09-11
feat(tokenizers): expose vocab introspection on the Tokenizer trait

ajcasagrande:
## Summary
- Adds vocab-introspection accessors (`vocab_size`, `token_to_id`, `special_token_ids`, `num_special_tokens_added`) to the `Tokenizer` trait, needed by a downstream consumer for raw-prompt generation and special-token accounting.
- `token_to_id`, `special_token_ids`, and `num_special_tokens_added` return `Result<T>` rather than a bare `Option`/`Vec`/`usize`: those bare types can't distinguish "this backend cannot determine the answer at all" from a genuine empty/miss/zero result (e.g. tiktoken's `token_to_id` always returning `None` vs. HuggingFace's `None` meaning "looked up, not in vocab"). `Err` = unsupported, `Ok(_)` = the real answer. All three default to `Err(unsupported)`, so only backends that can actually answer need to override.
- `vocab_size` stays `Option<usize>` — `Option` already separates `None` (can't determine) from `Some(0)` (empty vocab), so there's no ambiguity to fix there.
- HuggingFace, Baseten, and tiktoken backends implement the accessors they can determine; `FastTokenizer` and `CachedTokenizer` delegate to their inner tokenizer.

## Test plan
- [x] `cargo test -p dynamo-tokenizers` — 77 unit + 8 integration tests, including new coverage for the unsupported-path default (`tiktoken::tests::test_vocab_size_and_token_to_id_are_unsupported`, `cache::tests::unoverridden_vocab_methods_default_to_unsupported`)
- [x] `cargo fmt -p dynamo-tokenizers --check`
- [x] `cargo clippy -p dynamo-tokenizers --all-targets -- -D warnings`
- [x] `cargo build --workspace`

### DIFF
--- a/tokenizers/src/cache/mod.rs
+++ b/tokenizers/src/cache/mod.rs
@@ -271,7 +271,23 @@ impl Decoder for CachedTokenizer {
     }
 }
 
-impl Tokenizer for CachedTokenizer {}
+impl Tokenizer for CachedTokenizer {
+    fn vocab_size(&self) -> Option<usize> {
+        self.inner.vocab_size()
+    }
+
+    fn token_to_id(&self, token: &str) -> Option<TokenIdType> {
+        self.inner.token_to_id(token)
+    }
+
+    fn special_token_ids(&self) -> Vec<TokenIdType> {
+        self.inner.special_token_ids()
+    }
+
+    fn num_special_tokens_added(&self) -> usize {
+        self.inner.num_special_tokens_added()
+    }
+}
 
 #[cfg(test)]
 mod tests {
@@ -315,6 +331,22 @@ mod tests {
         fn validate_prefix_cache(&self) -> Result<()> {
             Ok(())
         }
+
+        fn vocab_size(&self) -> Option<usize> {
+            None
+        }
+
+        fn token_to_id(&self, _token: &str) -> Option<TokenIdType> {
+            None
+        }
+
+        fn special_token_ids(&self) -> Vec<TokenIdType> {
+            Vec::new()
+        }
+
+        fn num_special_tokens_added(&self) -> usize {
+            0
+        }
     }
 
     impl Encoder for FailingTokenizer {
@@ -340,6 +372,22 @@ mod tests {
     impl Tokenizer for FailingTokenizer {
         fn validate_prefix_cache(&self) -> Result<()> {
             Ok(())
+        }
+
+        fn vocab_size(&self) -> Option<usize> {
+            None
+        }
+
+        fn token_to_id(&self, _token: &str) -> Option<TokenIdType> {
+            None
+        }
+
+        fn special_token_ids(&self) -> Vec<TokenIdType> {
+            Vec::new()
+        }
+
+        fn num_special_tokens_added(&self) -> usize {
+            0
         }
     }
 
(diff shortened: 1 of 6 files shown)


tokenizers/src/fastokens.rs
```diff
@@ -76,6 +76,24 @@ impl Tokenizer for FastTokenizer {
...
+    fn token_to_id(&self, token: &str) -> Result<Option<TokenIdType>> {
+        self.hf_decoder.token_to_id(token)
+    }
+
+    fn special_token_ids(&self) -> Result<Vec<TokenIdType>> {
+        self.hf_decoder.special_token_ids()
+    }
+
+    fn num_special_tokens_added(&self) -> Result<usize> {
+        self.hf_decoder.num_special_tokens_added()
```
jthomson04, line 95: [P2] Match FastTokenizer special-token accounting to its encoder

For a tokenizer with a BOS post-processor, such as the TinyLlama fixture, this returns `Ok(1)` even though `FastTokenizer` never inserts BOS. Its `encode` calls upstream `encode`, which disables special-token insertion, and its inherited `with_options` ignores `add_special_tokens: true`. Thus, callers that subtract this count from a token budget reserve space for a token that is not added. Return zero for the current encoder behavior, or implement the option before reporting the post-processor count.

ajcasagrande:
All five review threads are addressed and resolved in c340355.\n\nFocused verification passed: cargo fmt --check plus the full dynamo-tokenizers suite (81 unit, 8 integration). The replacement workspace CI run reached the same unrelated failure reproduced on origin/main: conformance/tests/unified_render.rs requires the missing source-qualified fixture capture dynamo_v2-0.6.1+source.bb44f665182a33c083d64189dad7b366c3f0aff8efd27e06cd7c1ffd29924859. The tokenizer PR changes no conformance, parser, protocol, or fixture files.\n\n@jthomson04, the requested FastTokenizer special-token accounting fix is included and covered; a re-review would unblock the remaining approval gate.

tokenizers/src/cache/mod.rs
```diff
@@ -295,7 +295,23 @@ impl Decoder for CachedTokenizer {
...
+    fn token_to_id(&self, token: &str) -> Result<Option<TokenIdType>> {
+        self.inner.token_to_id(token)
+    }
+
+    fn special_token_ids(&self) -> Result<Vec<TokenIdType>> {
+        self.inner.special_token_ids()
+    }
+
+    fn num_special_tokens_added(&self) -> Result<usize> {
+        self.inner.num_special_tokens_added()
```
jthomson04: [P2] Match CachedTokenizer special-token accounting to its encoder

With the TinyLlama fixture, this returns `Ok(1)`, but `CachedTokenizer` inherits a `with_options` implementation that ignores `add_special_tokens: true`, and its constructor rejects an inner HF tokenizer configured to insert BOS. Encoding therefore adds no BOS, so callers that subtract this count from a token budget reserve space for a token that is never added. Return a count consistent with the wrapper's supported behavior, and check the count against encoding behavior rather than only against the inner accessor.

tokenizers/src/cache/mod.rs
```diff
@@ -647,4 +667,37 @@ mod tests {
         // First call populates, second/third hit.
         assert!(cached.cache_stats().hits >= 2, "expected hits on q2 and q3");
     }
+
+    #[test]
+    fn vocab_introspection_forwards_to_inner() {
+        // CachedTokenizer wraps `inner` behind an opaque cache; without
+        // forwarding, every caller reaching a tokenizer through the cache
+        // wrapper would see None/empty/zero regardless of what `inner`
```
jthomson04: [P3] Remove the remaining test narration

This comment repeats the forwarding assertions and states the wrong defaults: unsupported lookup and special-token methods return `Err`, not empty/zero. Remove it and the similar narration in `fastokens.rs:253-256`, `hf.rs:443-445`, `tiktoken.rs:497-502`, and the new vocabulary test in `basetenkenizer.rs:551-553,576-579,584-587`; the test names, fixtures, and assertions already show that behavior.

jthomson04 (approved):
The five earlier review comments are addressed. Source review completed at c340355bf07f8b2596b41558895200c0b7370c3f; the two remaining findings are attached inline. Tests and CI were not checked in this pass.

ishandhanani (approved):

PR #235, opened 2026-09-15
fix(muse-glimmer): drop truncated channel headers at EOF

indrajit96:
## Summary

  Fix Muse Glimmer responses leaking channel headers when generation ends before `<|message|>` is completed.

  ## Failure

  When generation ended with:

  ```text
  <|start|>assistant to=user
```
  the streaming response incorrectly exposed parser markup:
```
  {
    "content": "assistant to=user",
    "finish_reason": "length"
  }
```
  The EOF flush stripped <|start|> and treated the remaining incomplete header as ordinary text.

  ## Fix

  Detect framed Muse channel-header prefixes before stripping their framing. Incomplete framed headers are dropped at EOF, while unframed prose such as assistant to=user remains unchanged.

  Regression tests cover every truncated prefix for self, user, and tool channels.

   ## Passing response

  The equivalent truncated request now preserves the reasoning and emits no visible content:

  ```json
  {
    "id": "chatcmpl-0e30d95a-e67a-4bc2-bd25-8b22256213a3",
    "choices": [
      {
        "index": 0,
        "message": {
          "content": null,
          "refusal": null,
          "role": "assistant",
          "reasoning_content": "Think carefully and solve: A box has 17 rows of 23 marbles. How many marbles are there? Give the final answer clearly.\n\nA box has 17 rows of 23 marbles. How many marbles are there?\n\n17 rows of 23 marbles. 17 * 23 = ?\n\n17*23 = 17*(20+3)=340+51=391\n\nOr 23*17 = 391.\n\nAnswer: 391 marbles.\n\nProbably simple.\n\nGive final answer clearly.\n\nMaybe they want 391.\n\nCheck: 17 rows of 23 marbles = 391.\n\nYes.\n\nPossibly trick? No.\n\nFinal answer:
          391 marbles.\n\nIf they want explanation.\n\nReturn.\n\nProbably just 391.\n\nEnsure clear.\n\nAnswer: 391 marbles.\n\nDone."
        },
        "finish_reason": "length",
        "logprobs": null
      }
    ],
    "created": 1789509174,
    "model": "meta-models/Muse-Glimmer-30B",
    "object": "chat.completion",
    "usage": {
      "prompt_tokens": 84,
      "completion_tokens": 165,
      "total_tokens": 249,
      "prompt_tokens_details": {
        "cached_tokens": 0
      },
      "completion_tokens_details": {
        "reasoning_tokens": 162
      }
    }
  }
  ```

  Validated with Muse-Glimmer-30B on an H100. The targeted E2E pytest passed.

### CODE REVIEW

File: parsers/v2/src/tool_calling/muse_glimmer.rs [hunk 1]
```diff
@@ -249,6 +249,34 @@ fn valid_header_fragment(after: &str) -> bool {
     rest.is_empty() || (rest.len() < MESSAGE.len() && MESSAGE.starts_with(rest))
 }
 
+/// Whether `buffered` is a framed channel header that EOF cut before its
+/// `<|message|>` terminator arrived.
+///
+/// This must run before [`stripped`] removes `<|start|>`: the framing is the
+/// provenance that distinguishes parser-owned markup from an ordinary unframed
+/// response such as `assistant to=user`, which remains visible prose.
+fn is_incomplete_framed_header(buffered: &str) -> bool {
```
keivenchang, line 258: I think this still leaks parser markup for the supported recipient-less framed header form. `resolve_header` accepts `<|start|><|message|>` as a framed content header, but this helper only recognizes `<|start|>` followed by `assistant` or `to=`, so `<|start|><` and `<|start|><|` fall through to `flush_open_text` and reach the client as visible text. You can either recognize partial `<|message|>` prefixes immediately after `<|start|>` or remove recipient-less framing from the grammar; I think the first option matches the existing parser behavior. Please add EOF regressions for those prefixes and the complete recipient-less header path.

keivenchang (changes requested):
hey Indrajit, I found one remaining EOF case in the framed-header classifier. The supported recipient-less `<|start|><|message|>` form can still leak a partial `<` or `<|` when generation ends before the message marker completes, so this needs one more guard and regression coverage.

parsers/v2/src/tool_calling/muse_glimmer.rs
```diff
@@ -249,6 +249,34 @@ fn valid_header_fragment(after: &str) -> bool {
...
+    let Some(mut rest) = buffered.strip_prefix(START) else {
+        return false;
+    };
+    rest = rest.trim_start_matches(|c: char| c.is_whitespace());
+
+    if "assistant".starts_with(rest) {
```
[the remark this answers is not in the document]
indrajit96: Fixed in `da96310`. The EOF classifier now recognizes partial `<|message|>` prefixes after `<|start|>`. I also added complete and every-prefix coverage for both recipient-less header forms.

parsers/v2/src/unified/muse_glimmer.rs
```diff
@@ -367,6 +367,44 @@ mod tests {
         assert_eq!(out, vec![reasoning("thought")]);
     }
 
+    #[test]
```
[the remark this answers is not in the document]
indrajit96: Agreed. I removed `truncated_framed_answer_header_is_dropped_at_finish` and retained the exhaustive prefix test, which covers the same cutoff along with the new recipient-less cases.

parsers/v2/src/tool_calling/muse_glimmer.rs
indrajit96: Thanks for catching this. Fixed in `da96310`: `<|start|><` and `<|start|><|` are now treated as incomplete parser-owned headers and no longer emitted as content. The tests cover the complete recipient-less path and every truncated prefix for both role-less and `assistant` forms.

Validation: all 60 Muse unified tests and the full V2 suite passed; conformance rendered with 0 red.

keivenchang (approved):
approved — thanks Indrajit, the EOF classifier and exhaustive prefix coverage are in place. nice work.

parsers/v2/src/tool_calling/muse_glimmer.rs
```diff
@@ -1093,6 +1124,12 @@ impl MuseChannelScanner {
             // into framing. Complete markers are stripped; a committed partial
             // special token (`<|sta`) is parser-owned markup and dropped; the
             // ambiguous `<` / `<|` and any `to=`-shaped prose stay visible.
+            State::Idle if is_incomplete_framed_header(&buffered) => {
```
KrishnanPrash, line 1127: The guided reader is a second reader of this grammar and still has the bug this PR fixes. `guided_stray_header` (line 473) consumes only `START.len()` at flush:

```rust
    if flush {
        return haystack.find(START).map(|at| (at, START.len()));
    }
```

The input from the test this commit deleted, run under `tool_choice`:

```
in : " to=self<|message|>Look it up.<|eom|>" + "<|start|>assistant to=user"

native  [Reasoning("Look it up.")]
guided  [Reasoning("Look it up."), Text("assistant to=user")]
```

The `MESSAGE` branch added in `da96310a` widened the gap rather than closing it: `<|start|>assistant<|mes` is now `[]` natively and `Text("assistant<|mes")` guided. Same bytes, two answers by request mode, which is the `I3` split named at `unified/mod.rs:1167`. The comment on line 471 claims this branch makes "the same judgement `flush_open_text` makes for the native path"; that stops being true with this change.

Routing the flush branch through `is_incomplete_framed_header` closes it, and the full suite passes:

```rust
    if flush {
        return haystack.find(START).map(|at| {
            if is_incomplete_framed_header(&haystack[at..]) {
                (at, haystack.len() - at)
            } else {
                (at, START.len())
            }
        });
    }
```

parsers/v2/src/tool_calling/muse_glimmer.rs
```diff
@@ -249,6 +249,37 @@ fn valid_header_fragment(after: &str) -> bool {
     rest.is_empty() || (rest.len() < MESSAGE.len() && MESSAGE.starts_with(rest))
 }
 
+/// Whether `buffered` is a framed channel header that EOF cut before its
+/// `<|message|>` terminator arrived.
+///
+/// This must run before [`stripped`] removes `<|start|>`: the framing is the
+/// provenance that distinguishes parser-owned markup from an ordinary unframed
+/// response such as `assistant to=user`, which remains visible prose.
+fn is_incomplete_framed_header(buffered: &str) -> bool {
```
KrishnanPrash, line 258: The prompt ends with `<|start|>assistant`, so the turn's first header arrives unframed and this predicate never sees it:

```
fresh parser
in  : " to=self<|mess"
out : [Text(" to=self")]
```

`allow_bare_header` is the provenance signal that makes these bytes a header, and `drain_idle_prose` holds them for exactly that reason, but `take_stream_state()` resets the latch before `flush` can read it. Same shape in the missing-terminator recovery: `" to=self<|message|>thinking to=user"` yields `Reasoning("thinking to=user")` while the complete sibling yields `Reasoning("thinking ")`.

parsers/v2/src/unified/muse_glimmer.rs
```diff
@@ -367,6 +371,37 @@ mod tests {
...
+                    "header {header:?} cut at byte {at}"
+                );
+            }
+        }
+    }
+
+    #[test]
+    fn unframed_header_like_prose_is_kept_at_finish() {
+        for (input, expected) in [
+            ("to=user", "to=user"),
```
KrishnanPrash, line 397: `to=user` on a fresh parser is the one state where `allow_bare_header` is true, so these bytes are a header rather than prose. This case asserts the leak above is correct.

Spending the latch first, with a completed message ahead of the fragment, tests what this means to test:

```rust
("<|start|>assistant to=self<|message|>x<|eom|>to=user", "to=user"),
```

parsers/v2/src/tool_calling/muse_glimmer.rs
```diff
@@ -249,6 +249,37 @@ fn valid_header_fragment(after: &str) -> bool {
...
+    let Some(mut rest) = buffered.strip_prefix(START) else {
```
KrishnanPrash, line 259: `strip_prefix` anchors at byte 0, but `drain_idle_prose` holds the buffer from the *first* `<|start|>`. When stray framing sits in front, the real truncated header never reaches this check:

```
in  : "<|start|>noise<|start|>assistant to=user"
out : [Text("noiseassistant to=user")]
```

Scanning from the last `<|start|>` closes it. Plain prose is already fine (`"prose <|start|>assistant to=user"` drains correctly); it is prose *behind* framing that breaks.

parsers/v2/src/tool_calling/muse_glimmer.rs
```diff
@@ -249,6 +249,37 @@ fn valid_header_fragment(after: &str) -> bool {
...
+    if rest.len() < MESSAGE.len() && MESSAGE.starts_with(rest) {
+        return true;
+    }
+    if "to=".starts_with(rest) {
+        return true;
+    }
+    let Some(after_to) = rest.strip_prefix("to=") else {
+        return false;
+    };
+    valid_header_fragment(after_to)
```
KrishnanPrash, line 280: One whitespace character after the recipient re-opens the reported bug:

```
"<|start|>assistant to=user"   -> []
"<|start|>assistant to=user "  -> [Text("assistant to=user ")]
"<|start|>assistant to=user\n" -> [Text("assistant to=user\n")]
"<|start|>assistant<|mes "     -> [Text("assistant<|mes ")]
```

`valid_header_fragment` asks whether the bytes could still grow into a valid header, while the doc comment on line 255 says the framing is the provenance. The two rules disagree the moment whitespace follows the recipient, and the framing is still there.

parsers/v2/src/unified/muse_glimmer.rs
```diff
@@ -367,6 +371,37 @@ mod tests {
         assert_eq!(out, vec![reasoning("thought")]);
     }
 
+    #[test]
+    fn every_truncated_framed_header_prefix_is_dropped_at_finish() {
```
KrishnanPrash, line 375: `da96310a` deleted `truncated_framed_answer_header_is_dropped_at_finish`, which was the only multi-chunk case here and the only one reproducing the scenario in the PR description: a reasoning message completes, then a header is cut.

```rust
    let out = events(&[
        " to=self<|message|>Look it up.<|eom|>",
        "<|start|>assistant to=user",
    ]);
    assert_eq!(out, vec![reasoning("Look it up.")]);
```

The single-chunk sweep below does not cover it, because it never exercises a prior message or a chunk boundary. That input is also the one still failing in guided mode. Worth restoring?

parsers/v2/src/unified/muse_glimmer.rs
```diff
@@ -367,6 +371,37 @@ mod tests {
...
+        for header in [
```
KrishnanPrash, line 376: This loop reads as exhaustive, and the two header shapes added in `da96310a` widen it usefully. Every case in it still holds four axes fixed: native mode, byte-0 start, a single chunk, and no trailing whitespace.

Each of those is a live leak today:

```
guided mode      "<|start|>assistant to=user"              -> Text("assistant to=user")
offset > 0       "<|start|>noise<|start|>assistant to=user" -> Text("noiseassistant to=user")
bare header      " to=self<|mess"                           -> Text(" to=self")
trailing space   "<|start|>assistant to=user "              -> Text("assistant to=user ")
```

Adding those four rows here would have caught all of them.

PR #236, opened 2026-09-16
fix(gpt-oss): recover the final message a malformed harmony header strands

### LINKED ISSUE #237: gpt-oss: final message dropped when the model omits <|message|> after <|channel|>final
## Summary

`GptOssReasoningParser` silently drops a gpt-oss model's entire final message when the
model emits a malformed harmony header. Downstream this surfaces as `content: null` with
`finish_reason: "stop"` and HTTP 200 — indistinguishable from a legitimately empty answer.

`finish_reasoning_stream()` flushes only the held `pending_text` suffix. It never calls
`StreamableParser::process_eos()`; there is no `process_eos` call anywhere in the crate.

## Two distinct failure modes

These look identical from outside but fail for different reasons, which matters for the fix:

**1. Stray token where `<|message|>` belongs.** `StreamableParser::new(enc, Some(Role::Assistant))`
uses `ParseOptions::default()`, and `openai-harmony` 0.0.8 defines that default as
`strict: true`. In `StreamState::Header`, harmony's recovery arm is
`Some(token) if !self.options.strict && self.stop_tokens.contains(&token)` — disabled under
strict — so a stray token falls through to `header_tokens.push(token)` and is swallowed.
The parser never leaves `Header`, `last_content_delta()` never fires, and
`current_content()` returns `""` for any non-`Content` state. `process_eos()` would
`bail!("Unexpected EOS while waiting for message header to complete")` — but it is never called.

**2. Mistyped channel name** (e.g. `<|channel|>finalx<|message|>`). This one is *not* stranded.
Harmony reaches `Content`, parses the header, and hands over content deltas perfectly
normally; `process_eos()` succeeds. The text is lost purely because the streaming match arms
accept only `analysis`, `commentary` and `final`, so no arm claims the delta and it is dropped.

A `process_eos()`-based fix alone therefore recovers mode 1 but **not** mode 2.

## Measured behaviour

Four synthetic fixtures, run against this crate at `main` (`aa4d464`), at chunk sizes 1, 20 and
whole-string, plus `detect_and_parse_reasoning` for batch parity. `aggregator_content_would_be_null`
means `normal_text.trim().is_empty()`.

| Case | Input shape | `normal_text` | null? |
|---|---|---|---|
| A | `<\|channel\|>final<\|message\|>Hello!` (control) | `Hello!` | no |
| B | `<\|channel\|>final` + stray token + `Hello!` | `""` | **yes — bug** |
| C | `<\|channel\|>finalx<\|message\|>Hello!` | `""` | **yes — bug** |
| D | analysis block only, no final message | `""` | yes (correct) |

Identical at all three chunk sizes and in batch mode. Case D is the contrast case: empty
content there is the right answer, and a fix must not "recover" anything.

## Why no version bump helps

`gpt_oss_parser.rs` is **byte-identical** across `dynamo-parsers` 7.0.1 → 8.0.0 → 8.1.0 →
8.2.2 → 9.0.0 and to `frontend-crates@main`:

```
sha256 352bb8db98568cfa0f0737e886768478cebdaebb707ba0a69c2e69c4f6bc752d
```

`openai-harmony` 0.0.8 is the newest release. There is no version to upgrade to.

## Reference implementation

vLLM's `vllm/parser/harmony.py::flush()` handles exactly this and is the behaviour worth
matching:

```python
try:
    self._harmony_parser.process_eos()
    msg = self._poll_completed_message()
except HarmonyError:
    logger.warning("Harmony parser ended in a non-terminal state; "
                   "returning the recovered raw output.")
    text = self.model_tokenizer.decode(self._current_message_tokens)
    ...
```

## Question

Would you prefer `process_eos()`-based recovery, or `ParseOptions { strict: false }` plus
feeding the stop token through so harmony's own non-strict `Header` arm can fire?

**Corrected 2026-09-17.** This section previously claimed the stop token never reaches the
parser. That was wrong — a token capture of a real production null shows `200002`
(`<|return|>`) inside the stranded `header_tokens`, and `encode_with_special_tokens`
reproduces it. I had inferred it from the stop-token-hiding code without checking a capture.

The real reason not to take the `strict: false` route is a bug in `openai-harmony` 0.0.8, and
it is a stronger one. The non-strict `Header` arm is guarded on `next_role`
(`encoding.rs:1173`), and `next_role` is cleared at `encoding.rs:1162` on the first message's
`Header` → `Content` transition. So the arm can only ever recover the **first** message of a
stream — and gpt-oss always emits `analysis` first:

| Input, `strict: false` | Recovered |
|---|---|
| single malformed message (`next_role` still set from ctor) | `final__HANDOFF__` |
| `analysis`, then malformed `final` (`next_role` cleared) | **nothing** |

Worse for this use case: under `strict: false` the arm still fires, discarding `header_tokens`
and moving to `ExpectStart`, so `process_eos()` afterwards returns `Ok` instead of `Err`. The
stranded content is gone with no error signal, which would remove the very trigger a
`process_eos()`-based recovery depends on.

So `strict: false` is not a milder alternative — it is incompatible with the recovery in #236.
The `next_role` guard looks like a separate bug worth reporting against `openai-harmony`.

## Proposed fix

PR #236 implements `process_eos()`-based recovery for mode 1 and an unattributed-delta buffer
for mode 2.

Note that PR alone does **not** change observable behaviour: the dynamo preprocessor invokes
`finish_reasoning_stream()` only for `force_nonempty_content` requests, so for gpt-oss the
finalizer is never called. The companion gate widening is ai-dynamo/dynamo#14934, and the
underlying general defect is ai-dynamo/dynamo#14935.

## Related

ai-dynamo/dynamo#7154 reports the same user-visible symptom (gpt-oss, `content` missing while
`reasoning_content` is present, intermittent) on **TensorRT-LLM** rather than vLLM. It was
closed by the stale bot without a root cause. That it reproduces on a different backend is
consistent with the defect living here, in the shared parser, rather than in any engine.

---
*Filed by an AI agent (Claude Opus 5) working from a source audit plus a local reproduction.
All fixtures are synthetic; no user or deployment data is included.*

### LINKED ISSUE #14935: [BUG]: Reasoning parsers are never finalized at EOF unless force_nonempty_content is set
### Describe the Bug

Reasoning parsers are never finalized at end of stream unless the request sets
`force_nonempty_content`. Any `ReasoningParser` still holding state when the upstream stream
ends loses it silently.

The streaming reasoning transform in `lib/llm/src/preprocessor.rs` has two flush sites, and
both were gated on `defer_reasoning_for_nonempty_content`:

- the terminal-chunk drain, gated on `defer_reasoning && choice.finish_reason.is_some() && ...`
- the end-of-stream fallback, whose guard reads
  `} else if !state.defer_reasoning_for_nonempty_content || state.saw_terminal_error {` → `None`

The code states the consequence plainly in its own comment on that branch:

```rust
// Only the force_nonempty_content path reaches this branch; every
// other parser took the `None` branch above and keeps its original
// no-flush EOF behavior.
```

`ReasoningParser::finish_reasoning_stream` exists precisely because incremental parsing buffers
an undecided partial delimiter rather than emitting it — the next chunk might complete a
`<think>` or `</think>`. At EOF no next chunk is coming, so the parser must flush. For every
parser except the deferring one, that call never happened and the buffer was dropped with the
parser state.

This is a **general** defect. Every parser that overrides `finish_reasoning_stream` to flush a
buffer is affected — in `dynamo-parsers` that is `base`, `gemma4`, `granite`, `inkling` and
`gpt_oss`. Since `base_parser` backs the `<think>`-style parsers, the reachable surface is wide.

gpt-oss is the worst instance because there the lost state can be the *entire answer* rather
than a few characters of partial delimiter — see ai-dynamo/frontend-crates#236, where a
malformed harmony header strands the whole final message in the parser.

### Steps to Reproduce

Drive `parse_reasoning_content_from_stream_inner` with a reasoning parser and a stream whose
last content chunk ends mid-delimiter, without `force_nonempty_content`:

```
parser: "deepseek_r1", defer_reasoning_for_nonempty_content: false

chunk 1: content = "<think>thought</think>Answer"
chunk 2: content = "<thi"            # partial opener, held by the parser
chunk 3: finish_reason = "stop"
```

`<thi` is a partial `<think>` opener, so `BasicReasoningParser` holds it in `_buffer` rather
than emitting it. No chunk completes the marker.

### Expected Behavior

The held bytes are flushed at EOF and reach the client, attributed by the parser's own split —
recovered reasoning as `reasoning_content`, recovered text as `content`.

### Actual Behavior

The held bytes are dropped. `finish_reasoning_stream()` is never called, and the parser is
discarded with `_buffer` still populated.

For gpt-oss the same gap loses the whole final message, producing a response with
`content: null` and `finish_reason: "stop"` at HTTP 200 — which nothing downstream can
distinguish from a legitimately empty answer.

### Environment

Present on `main` at `c76d3208f32b`, and on the `v1.5.0` tag (`bfcf71343acf`). Backend-agnostic:
this is in the shared Rust preprocessor, so it reproduces regardless of engine. `grep -c process_eos`
in the embedded `dynamo-parsers` reasoning parsers returns 0 for every published version from
7.0.1 through 9.0.0.

### Additional Context

**Proposed fix:** #14934 widens both gates so every parser is finalized, and passes the deferral
flag explicitly into the drain so attribution stays correct per path:

- *without* deferral nothing was ever held back, so the parser's own split is authoritative;
- *with* deferral the buffered bytes are ambiguous by construction, and the existing
  `force_nonempty_content` rule still decides where they land (unchanged).

An error remains terminal — `saw_terminal_error` still suppresses the flush, so a failed
generation cannot emit recovered text as if it had succeeded.

That PR also makes the end-of-stream envelope a capture-once rather than a per-chunk clone,
which removes the hot-path cost that widening the flush would otherwise have introduced. It is
a net reduction: the per-token clone previously ran for the entire deferral path.

**Open question worth a maintainer's view:** is the same EOF gap reachable for the other
reasoning parsers via a different malformed shape? A general fix is easier to justify than a
gpt-oss special case, which is why #14934 widens the gate for all of them rather than
special-casing harmony.

**Prior report:** #7154 describes the same user-visible symptom (gpt-oss, `content` missing
while `reasoning_content` is present, intermittent) on TensorRT-LLM. It was closed by the stale
bot without a root cause. That it reproduces on a different backend is consistent with the
defect living in the shared preprocessor rather than in any engine.

---
*Filed by an AI agent (Claude Opus 5) from a source audit plus local reproduction and tests.
No user or deployment data is included.*

hnt2601:
Closes #237. Companion to ai-dynamo/dynamo#14934 (which closes ai-dynamo/dynamo#14935).

## Summary

A gpt-oss model that emits a malformed harmony header loses its entire final
message. The response comes back with `content: null`, `finish_reason: "stop"`
and HTTP 200, so nothing downstream can tell it apart from a legitimately empty
answer.

Two distinct shapes cause it, and they fail for different reasons:

**1. Stray token where `<|message|>` belongs** — e.g. `<|channel|>final` followed
by anything other than `<|message|>`. `StreamableParser` never leaves
`StreamState::Header`: every later token is pushed onto `header_tokens` instead
of becoming a content delta. `process_eos()` — which this crate never called —
fails with *"Unexpected EOS while waiting for message header to complete"*, and
the whole message is discarded with the parser state.

`finish_reasoning_stream()` now calls `process_eos()` and, on error, decodes the
stranded header tokens and surfaces them.

**2. Mistyped channel name** — e.g. `<|channel|>finalx<|message|>`. This one is
*not* stranded: harmony reaches `Content`, parses the header, and hands over
content deltas perfectly normally. `process_eos()` succeeds. The text is lost
purely because the streaming match arms only accept `analysis`, `commentary` and
`final`, so no arm claims the delta and it is dropped.

Such deltas are now held in `unattributed_text` and surfaced at EOF. Directed
`analysis`/`commentary` tool-call payloads keep being suppressed — they are
reconstructed as whole envelopes on the `<|call|>` chunk, and that path is
untouched.

The reference behaviour is vLLM's `vllm/parser/harmony.py::flush()`, which
catches `HarmonyError` from `process_eos()` and returns the recovered raw
output. Recovered text is attributed to the `final` channel to match it, so the
same generation is reported identically whichever frontend served it.

## Design notes

- **No double-emit.** Recovery decodes *only* the header tokens, read back from
  `state_json()` — never the cumulative `tokens()` buffer, which also holds every
  message already streamed to the caller. Header tokens have by definition never
  been emitted. On the `Ok` path nothing is collected at all: `process_eos()`
  there merely closes a message whose content was already streamed token by
  token, so re-reading `messages()` would duplicate text the caller already has.
- **No false recovery.** A turn that genuinely produced only an analysis block
  still yields empty content. The recovery never invents an answer.
- **No markup leak.** Recovered text is stripped of harmony control tokens
  before it can reach assistant `content`.
- **Idempotent.** `finish_reasoning_stream()` is now safe to call twice and
  documented as such. `process_eos()` cannot be repeated, and the dynamo
  preprocessor is growing more paths that finalize.

Also adds `ReasoningParser::has_unflushed_state`, defaulting to `false` and
overridden by every parser that buffers (`base`, `gemma4`, `granite`,
`inkling`, `gpt_oss`). It lets a streaming caller decide whether an
end-of-stream flush envelope is worth retaining without paying a per-chunk clone
on the hot path.

## Correction: why `ParseOptions { strict: false }` is not the fix

**An earlier revision of this description gave the wrong reason for rejecting the
`strict: false` alternative, and I want to correct it rather than quietly edit it away.**
It claimed the stop token never reaches the parser. That is false and falsifiable in one
command — a token-level capture of a real production null shows `200002` (`<|return|>`)
sitting inside harmony's stranded `header_tokens`:

```
[173781, 200005, 17196, 2516, 157262, 41265, 771, 200002]
 assistant <|channel|> final   __     HAND   OFF   __   <|return|>
```

and `encode_with_special_tokens("<|channel|>final__HANDOFF__<|return|>")` does emit `200002`.
The stop token reaches the parser.

The real reason is a bug in `openai-harmony` 0.0.8, and it is a **stronger** argument for the
`process_eos()` route. The non-strict `Header` recovery arm is guarded on `next_role`:

```rust
Some(token) if !self.options.strict && self.stop_tokens.contains(&token) => {
    if let Some(role) = next_role_clone {        // <-- guard
        if !header_tokens.is_empty() { ...push recovered message... }
    }
    self.state = StreamState::ExpectStart;
    self.next_role = None;
}
```

`next_role` is cleared at `encoding.rs:1162`, on the first message's `Header` → `Content`
transition. So the arm can only ever recover the **first** message of a stream — and gpt-oss
always emits `analysis` first. Measured both ways:

| Input (`strict: false`) | Result |
|---|---|
| single malformed message (`next_role` still set from ctor) | recovers `final__HANDOFF__` |
| `analysis`, then malformed `final` (`next_role` cleared) | **no final message** |

There is a second consequence that matters more for this PR: under `strict: false` the arm still
*fires* — it discards `header_tokens` and moves the state to `ExpectStart` — so `process_eos()`
afterwards returns **`Ok`**. The stranded content is gone with no error signal. Measured:

| | `process_eos()` after `analysis` + malformed `final` |
|---|---|
| `strict: true` (this PR) | **`Err`** — recovery fires, message recovered |
| `strict: false` | `Ok` — header tokens already discarded, nothing to recover |

So flipping to `strict: false` would not merely fail to help, it would **defeat** the recovery in
this PR by removing the error that triggers it. That is why this PR keeps `strict: true` and
recovers from the `process_eos()` error instead.

I think the `next_role` guard is worth a separate issue against `openai-harmony`; happy to file it
if you agree it reads as a bug rather than intent.

## Important: this PR alone does not change observable behaviour

The dynamo preprocessor only invokes `finish_reasoning_stream()` for
`force_nonempty_content` requests, so for gpt-oss the finalizer is never called
and this recovery never runs. The companion change widening that gate is
ai-dynamo/dynamo#14934. **Both are required**; neither fixes the bug on
its own.

## Production evidence

The fixtures below are synthetic, so here are measured numbers from a real gpt-oss-120b
deployment.

**Impact before the fix:** 29 nulls in 150 sequential identical requests — **19.3%**, 0 HTTP
errors, every null carrying `finish_reason: "stop"`. This is not a rare edge case.

**Recovery, replayed through this branch.** Captured completion token ids were decoded to text and
fed exactly the way the dynamo preprocessor feeds the parser —
`parse_reasoning_streaming_incremental(text, &[])`, text only, empty `token_ids` — at chunk sizes
1 char, 20 chars and whole-string, then finalized:

| Set | nulls recovered | good responses changed |
|---|---|---|
| 60 mixed (11 null / 49 good) | **11/11** | **0/49** — byte-identical to what the live API returned |
| 21 nulls | **21/21** | — |
| **total** | **32/32** | **0/49** |

No case was chunk-boundary dependent. The 49-good column is the regression argument: on every
response that was already fine, the patched parser reproduces the live API's output exactly.

## Known imperfection in recovered text

4 of the 32 recovered as `*)__HANDOFF__` rather than `__HANDOFF__`. The `*)` comes from the
malformed token itself — the model emitted `164797` (`*)__`) where `<|message|>` belonged — so
there is no general way to tell a junk header token from the start of content. Recovery therefore
surfaces it rather than guessing.

Two stray characters is a better outcome than a silently lost answer, but it is not a clean
recovery and the PR should not imply otherwise. If you would prefer a heuristic here (for example
trimming a leading non-alphanumeric run), say so — I left it verbatim on the principle that
inventing a trim rule risks eating real content.

## Validation

Reproduced against this crate at `main` before the change and re-run after, at
chunk sizes 1, 20 and whole-string, plus batch parity. All fixtures are
synthetic; the malformed shape imitates an observed stray token in place of
`<|message|>`.

| Case | Before | After |
|---|---|---|
| A — well-formed `<|channel|>final<|message|>` | `Hello!` | `Hello!` (unchanged, no duplication) |
| B — stray token instead of `<|message|>` | `""` ← bug | recovered |
| C — mistyped channel `finalx` | `""` ← bug | recovered |
| D — analysis only, no final message | `""` | `""` (correct; no false recovery) |

- `cargo test -p dynamo-parsers` — **830 passed, 0 failed** (7 new)
- `cargo clippy -p dynamo-parsers --all-targets` — clean
- `cargo fmt` — clean
- The pre-existing regression guard for the `pending_text` behaviour this PR
  touches (`test_gpt_oss_reasoning_parser_streaming_flushes_pending_suffix_on_finish`)
  passes unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hnt2601:
**Correcting a claim I made in this PR's description, and adding production evidence.** Flagging it
as a comment rather than only editing the body, since the original wording has been up for a while.

### What was wrong

The description previously argued against the `ParseOptions { strict: false }` alternative like
this:

> the serving pipeline hides EOS ids from the decoded text and calls the parser with text only and
> empty `token_ids`, so no stop token ever reaches the parser and the non-strict arm has nothing to
> fire on.

**That is false.** A token capture of a real production null shows `200002` (`<|return|>`) inside
harmony's stranded `header_tokens`, and `encode_with_special_tokens` reproduces it:

```
encode("<|channel|>final__HANDOFF__<|return|>") = [200005, 17196, 771, 157262, 41265, 771, 200002]
                                                                                          ^ <|return|>
```

The stop token does reach the parser. I had reasoned that from the stop-token-hiding code without
checking it against a real capture.

### The actual reason, which is a better argument

`openai-harmony` 0.0.8 guards its non-strict `Header` recovery arm on `next_role`
(`encoding.rs:1173`), and clears `next_role` at `encoding.rs:1162` on the first message's
`Header` → `Content` transition. The arm can therefore only ever recover the **first** message of a
stream — and gpt-oss always emits `analysis` first. Measured:

| Input, `strict: false` | Messages recovered |
|---|---|
| single malformed message (`next_role` still set from ctor) | `final__HANDOFF__` |
| `analysis`, then malformed `final` (`next_role` cleared) | **none** |

And the part that matters most here — under `strict: false` the arm still *fires*, discarding
`header_tokens` and moving to `ExpectStart`, so `process_eos()` then returns **`Ok`**:

| | `process_eos()` after `analysis` + malformed `final` |
|---|---|
| `strict: true` (this PR) | **`Err`** → recovery fires, message recovered |
| `strict: false` | `Ok` → header tokens already discarded, nothing left to recover |

So `strict: false` would not merely fail to help — it would **defeat** this PR's recovery by
removing the error that triggers it. That is now the stated reason in the description.

Happy to file the `next_role` guard as an issue against `openai-harmony` if you read it as a bug
rather than intent.

### Production evidence added

The description now also carries measured numbers rather than only synthetic fixtures:

- **19.3% null rate** before the fix — 29 nulls in 150 sequential identical requests against a
  gpt-oss-120b deployment, 0 HTTP errors, every null with `finish_reason: "stop"`.
- **32/32 captured production nulls recovered** when replayed through this branch, with
  **0/49 good responses changed** — byte-identical to what the live API returned.

### One imperfection now disclosed

4 of the 32 recover as `*)__HANDOFF__` rather than `__HANDOFF__`. The `*)` is part of the malformed
token itself (`164797` = `*)__`), so there is no general way to separate a junk header token from
the start of content. I left it verbatim rather than inventing a trim rule that could eat real
content — but it is not a clean recovery and the description now says so.

### Also worth recording: this PR is load-bearing, and was being tested without being present

A runtime image built from the companion dynamo branch still produced nulls, which initially looked
like a defect in one of the PRs. It was not. The dynamo workspace consumes `dynamo-parsers` from
crates.io with no `[patch.crates-io]`, so rebuilds linked the *published, unpatched* crate — the
finalizer was there with nothing to call into.

With this PR patched into a dynamo build and driven through dynamo's real reasoning pipeline, at
chunk sizes 1, 20 and whole-string:

| Build | `content` |
|---|---|
| dynamo finalizer + this PR | `"__HANDOFF__"` |
| dynamo finalizer, this PR reverted | `""` — the production null, reasoning intact |

Which is direct evidence for the "neither PR changes observable behaviour alone" note above.

Issue #159, opened 2026-07-28
Accept Kimi-style system tools and assistant partial fields


michaelfeil:
## Summary

Kimi-style chat templates may send dynamic tool metadata on a system message without ordinary system text, and may use an assistant `partial` flag for prefill-style assistant turns. Dynamo currently relies on `async-openai` chat request types for some of this shape, which are too strict for these model-specific but OpenAI-compatible frontend payloads.

Baseten currently carries local protocol type overrides for:

- `ChatCompletionRequestSystemMessage.content` being optional.
- A system-message `tools` passthrough field.
- `ChatCompletionRequestAssistantMessage.partial`.

## Desired behavior

The protocol layer should accept a synthetic request like:

```json
{
  "model": "dummy-kimi-model",
  "messages": [
    {
      "role": "system",
      "tools": [
        {
          "name": "lookup",
          "description": "dummy lookup tool",
          "parameters": {
            "type": "object",
            "properties": {
              "query": { "type": "string" }
            }
          }
        }
      ]
    },
    {
      "role": "assistant",
      "content": "synthetic prefill",
      "partial": true
    },
    {
      "role": "user",
      "content": "continue"
    }
  ]
}
```

The `tools` field should be preserved for downstream chat-template rendering. The `partial` flag should be optional and omitted from serialization when absent.

PR #240, opened 2026-09-17
fix(renderer): reject unsupported extensions in DeepSeek V4.1

luckyq:
## Summary

- reject unsupported `assistant.partial: true` before DeepSeek V4.1 prompt rendering
- reject non-empty `system.tools` before top-level tool injection
- preserve ordinary DeepSeek V4.1 requests and standard top-level tools

## Why

DeepSeek V4.1 was the only native formatter that skipped the shared guards used by the other non-Kimi formatters. As a result, Kimi-style message extensions could be accepted but rendered with the wrong semantics: a partial assistant turn was closed, a bare dynamic tool produced an empty schema section, and top-level tools could overwrite message-level declarations.

The guards now run on the original messages before normalization or injection, so unsupported shapes fail explicitly as `PromptRenderError::InvalidRequest` instead of silently producing a mismatched prompt.

## Tests

- added a regression test for `assistant.partial: true`
- added a regression test combining a bare `system.tools` declaration with a valid top-level tool
- `git diff --check`
- rendered the canonical conformance report: 81 model/tab pairs, 0 red

Per the requested workflow, compilation and Rust tests are delegated to PR CI rather than run locally.

## Related

- Follow-up to the DeepSeek V4.1 finding in [Dynamo PR #14755](https://github.com/ai-dynamo/dynamo/pull/14755#discussion_r4022493250)
- Related to [frontend-crates issue #159](https://github.com/ai-dynamo/frontend-crates/issues/159)


renderer/src/deepseek/v41.rs
```diff
@@ -160,6 +160,12 @@ impl crate::OAIPromptFormatter for DeepSeekV41Formatter {
     }
 
     fn render(&self, req: &dyn crate::OAIChatLikeRequest) -> Result<String> {
+        let messages_value = req.messages();
+        let messages_json =
+            serde_json::to_value(&messages_value).context("Failed to convert messages to JSON")?;
+        crate::reject_unsupported_partial_assistant(&messages_json)?;
+        crate::reject_unsupported_message_tools(&messages_json, &["developer"])?;
```
indrajit96, line 167: ### [P2] Preserve developer tools during top-level injection

Passing `["developer"]` declares developer-level tools supported, but V4.1 later assigns top-level tools directly to `messages[0]["tools"]`. If the first message is a developer message with its own tools, that assignment silently replaces them.

### Example

A raw `OAIChatLikeRequest` containing `developer_tool` on the first developer message and `top_level_tool` at request level renders only `top_level_tool`. V4 and V3.2 preserve both declarations.

The standard typed chat request currently rejects `developer.tools`, so this primarily affects custom implementations of the public `OAIChatLikeRequest` interface.

### Suggested fix

When the first developer message already declares tools, preserve both sets—either inject the top-level tools into a separate leading system message or merge them without replacement. Add the V4/V3.2 parity regression asserting both tool names appear.

indrajit96 (approved):
LGTM!
Added a small comment.
https://github.com/ai-dynamo/frontend-crates/pull/240#discussion_r4040441231

renderer/src/deepseek/v41.rs
luckyq: Fixed in c499c59. When effective top-level tools would collide with a non-empty leading developer.tools declaration, V4.1 now injects them into a separate leading system message, preserving both declarations. Added a raw OAIChatLikeRequest parity regression that asserts both developer_tool and top_level_tool are rendered. The signed commit is verified; Rust CI is running.

luckyq: CI is now green, including the full Rust test suite, doctests, Clippy, and crate publishability verification. Resolving this thread.

PR #243, opened 2026-09-18
fix(tokenizers): disable unused fastokens Hub downloads

lagoon0o0:
## Summary

`dynamo-tokenizers` loads Fastokens from local files, but Fastokens' default `hf-hub` feature also enables unused downloads through HF Hub 0.4.3 and `ureq`.

Disable that default feature and remove its two lockfile dependency edges. All locked package versions remain unchanged. The demo retains its existing HF Hub 1.0 downloader; Tokenizers' optional TLS forwarding and public tokenizer APIs are unchanged.

## Validation

Rebased onto upstream `98c7c29` in `7669021`: 78 tokenizer tests, tokenizer Clippy/doctests, formatting, and macOS/Linux dependency-graph checks pass. All 589 upstream locked package versions, sources and checksums are preserved; only the two Fastokens download dependency edges are removed.

The broader validation below was performed on the original change (`3850385`, based on `a6f6c65`); the full workspace suite was not rerun after the rebase.

- Rust 1.96.1/macOS: 78 tokenizer tests pass before and after the change.
- macOS and Linux dependency graphs: tokenizer-only builds have no active HF Hub; workspace default/all-features retain the demo's HF Hub 1.0 dependency, with 0.4.3 inactive.
- Packaged crate preserves `default-features = false`; an independent consumer compiles and matches local loading, token IDs and decoded text on English/Chinese inputs.
- Baseline/candidate demo responses match for `/health`, `/v1/tokenize` and `/v1/detokenize`.
- SGLang/Rust 1.92: both variants pass server checks and 44 existing tokenizer-manager tests; loading, encoding and full/streaming decoding match.
- Formatting, workspace check/doctests/Clippy, `cargo machete`, dependency/license checks and workspace publish dry run pass.
- Full workspace tests: **2147 passed, 1 failed, 7 ignored**. The failure is detailed below.

## Known validation limitation

`committed_dynamo_capture_matches_the_live_parsers` fails because its expected capture directory is absent. The exact upstream baseline (`a6f6c65`) also fails this test, expecting `dynamo_v2-0.6.1`. This change updates the source fingerprint because it includes both root Cargo files, so the candidate expects a new source-qualified capture, which is also absent. Both fail before comparing parser output. No parser behavior or capture records are changed in this PR.

Linux build/test CI still needs maintainer approval for this external contribution; Linux dependency graphs were checked locally.

## Dependency scope

Inactive optional dependencies can remain in `Cargo.lock`. Downstream applications that explicitly enable Fastokens Hub downloads or Tokenizers HTTP support can reintroduce the older Hub dependency. This change does not address SGLang's separate cache-lookup behavior.


rmccorm4:
/ok to test [17bf170](https://github.com/ai-dynamo/frontend-crates/pull/243/commits/17bf170a8ced443274e5af80032dc58fd634b224)

PR #244, opened 2026-09-18
perf(tokenizers): reuse prefix hashes on cache hits and misses

jthomson04:
## Summary

Tokenizer cache lookup already computes the BLAKE3 digest at every special-token boundary. Reuse the deepest digest when extending a partial hit, and retain the boundary/hash vector on a miss so cache population does not scan or hash the input again.

Keep the public `L1Cache` signatures through compatibility wrappers. A private lookup result carries either the matched prefix or the existing hash vector. Both public population calls and saved-hash misses use one population loop. Cache keys, boundary selection, token IDs, token-vector allocation policy, counters, and error handling are unchanged. Preserve the overlap-safety checks and incremental miss insertion: earlier completed prefixes remain cached if a later segment fails.

Add debug-only checks for input/digest mismatches, remove the unreachable empty-suffix branch, and cover saved entries with both ASCII and Unicode inputs. The optional digest remains only to support the existing public extension method without a separate implementation.

## Validation

Current follow-up on base `23b402787dab`:

- `cargo fmt -p dynamo-tokenizers -- --check` passed.
- `cargo test -p dynamo-tokenizers --lib cache:: --locked`: 31 passed.
- `cargo test -p dynamo-tokenizers --test cache_correctness --locked`: 8 passed.
- `cargo clippy -p dynamo-tokenizers --all-targets --locked -- -D warnings` passed.
- Tests compare saved-hash and public miss population, exact tokens and cached entries, accounting, empty/no-boundary inputs, trailing special tokens, partial failures, and debug rejection of mismatched inputs. Existing concurrent, eviction, observer, and extension-disabled tests pass.
- Source inspection confirms that miss population uses the saved iterator and that its validation hash state and updates are guarded by `cfg(debug_assertions)`. Hit validation uses `debug_assert_eq!`.

### Earlier partial-hit measurements

**The measurements below apply only to the earlier partial-hit implementation on 1.8.1. This review follow-up has correctness validation only; miss-path savings have not been measured.** No new benchmark or frontend performance claim is included.

- The measured 1.8.1 variant passed 26 cache unit tests and 8 integration tests locally and on Grace. Both benchmark variants matched the HF correctness oracle for 128 AgentX-derived base inputs, 128 continuations, and the 775,168-token input.
- Grace comparison used one shared CachedTokenizer, 72 caller threads on NUMA0, a 4 GiB prefix cache, jemalloc, identical locked dependencies, and identical per-caller request sequences and quotas. No tokenizer or Rayon thread overrides. Input construction was outside encode timing.

| Microbenchmark | Baseline | Hash reuse | Change |
|---|---:|---:|---:|
| Encode CPU/request, 2,500 requests/s | 0.952 ms | 0.644 ms | **32.4% lower** |
| Encode CPU/request, 10,000 requests/s | 0.955 ms | 0.644 ms | **32.6% lower** |
| Unpaced fixed-work throughput, 72 callers | 53.0k requests/s | 71.1k requests/s | **34.0% higher** |

Short-input and extension-disabled controls were effectively unchanged (CPU/request differences below 1%). Allocations and allocated bytes per request were identical. Matching 499 Hz profiles showed approximately 32.7% fewer encode cycles/request: hash work outside the named lookup frame fell from about 956k cycles/request to approximately zero, while lookup-hash cycles/request changed by only 0.3%. Both accepted profiles had zero lost records and less than 0.1% cycle weight with unknown frames.

These measurements compare the patch against dynamo-tokenizers 1.8.1 at `ffa953fc76b0e03731b1c5d94f9fcb2e29e0ab37`; the PR is ported onto current main with focused tests rerun. This is a warm-prefix CachedTokenizer microbenchmark, not a full frontend benchmark or exact AgentX trace replay. Base inputs span 2,439–386,789 tokens; measured requests average about 141k tokens with 98.35% token reuse. The 775,168-token case is correctness coverage only: no dedicated ultra-long-context performance sweep or sequence-length scaling claim is included.


tokenizers/src/cache/l1.rs
```diff
@@ -387,12 +434,13 @@ impl L1Cache {
...
 
-        // Key is blake3 of input[0..deepest]. Built with the same streaming idiom as
-        // `longest_prefix_match`/`insert_at_boundaries` so the digest is byte-for-byte
-        // identical to the incremental one a future lookup computes for this prefix.
-        let mut hasher = blake3::Hasher::new();
-        hasher.update(&input.as_bytes()[..deepest]);
-        let hash_bytes: Blake3Hash = *hasher.finalize().as_bytes();
+        // CachedTokenizer already computed this key during lookup. Preserve the
+        // existing public extension method for callers that do not supply a digest.
+        let hash_bytes = deepest_hash.unwrap_or_else(|| {
```
jthomson04, line 466: **The reused digest is accepted on trust — nothing ties `deepest_hash` to the `input`/`deepest` pair it is keyed against.**

Before this change the key was always derived from the actual `input` inside this function, so a desync between "the string we tokenized" and "the string the key names" was structurally impossible. Now `input` and `matched` are independent parameters, and `hash_bytes` is inserted without ever being checked against `input[..deepest]`.

Failure scenario: call `longest_prefix_match_with_hash(a)` and then `extend_after_match_with_hash(b, matched_a, tok)` — or, more realistically, make a future edit where `deepest` is anything other than `matched.deepest_boundary` (clamping it, or selecting a shallower boundary). The cache then holds an entry keyed by `blake3(a[0..d])` whose value is `prefix_tokens(a) + encode(b[prefix_len..d])`. A later lookup of `a` hits that key and returns token ids for a different string: silently wrong tokenization, no error, no counter, no failing test.

Suggested fix: bind the input into the match — `PrefixMatch<'a> { input: &'a str, .. }` — which also removes the now-redundant `input` parameter, and/or add
```rust
debug_assert_eq!(hash_bytes, *blake3::hash(&input.as_bytes()[..deepest]).as_bytes());
```

tokenizers/src/cache/mod.rs
```diff
@@ -220,31 +220,27 @@ impl Encoder for CachedTokenizer {
             return self.inner.encode(input);
         }
 
-        if let Some((prefix_tokens, prefix_len, deepest_boundary)) =
-            self.l1.longest_prefix_match(input)
-        {
-            let cached_tokens = prefix_tokens.len();
-            let suffix = &input[prefix_len..];
+        if let Some(matched) = self.l1.longest_prefix_match_with_hash(input) {
```
jthomson04: **The miss path still performs the exact duplicated work this PR removes from the hit path.** (The affected call is `populate_and_encode` at mod.rs:256, just below the last diff hunk.)

By the time control reaches the miss path, this lookup has already run `self.boundaries(input)` (a full Aho-Corasick pass) and incrementally hashed every prefix up to the deepest boundary. `populate_and_encode` then calls `self.boundaries(input)` again, and `populate_boundaries` re-hashes every prefix from scratch. For the ~500 KB inputs benchmarked in the PR description that is one wasted AC pass plus one wasted full blake3 pass per miss — the same redundancy the PR measures as ~956k cycles/request on the hit path.

Suggested fix: thread the boundaries and the per-boundary digests already computed in `longest_prefix_match_with_hash` into `populate_and_encode`, the same way the deepest digest is now threaded into `extend_after_match_with_hash`. That fixes the underlying mechanism rather than just the hit path.

tokenizers/src/cache/mod.rs
```diff
@@ -220,31 +220,27 @@ impl Encoder for CachedTokenizer {
...
+            let cached_tokens = matched.tokens.len();
+            let suffix = &input[matched.prefix_len..];
             let encoding = if suffix.is_empty() {
```
jthomson04: **This `suffix.is_empty()` branch is unreachable dead code.**

`boundaries_with` filters with `.filter(|&end| end < text.len())`, so every boundary — and therefore every `prefix_len` a match can return — is strictly less than `input.len()`. `suffix` is never empty, so `Encoding::Sp(matched.tokens.to_vec())` on the next line can never execute.

It reads as a live fast path, and as the only place in the hit path where a whole cached prefix (potentially ~500 KB) is deep-copied, which invites future readers to reason about a state that cannot occur.

Suggested fix: delete the branch, or keep it with a comment / `debug_assert!` recording why it is unreachable.

tokenizers/src/cache/l1.rs
```diff
@@ -48,6 +48,15 @@ type PrefixHasher = BuildHasherDefault<FxHasher>;
 /// Weighted W-TinyLFU cache mapping a prefix's blake3 digest to its cumulative tokens.
 type PrefixCache = Cache<Blake3Hash, Arc<[TokenIdType]>, PrefixHasher>;
 
+/// Request-local lookup result. The deepest digest can differ from the matched key.
+pub(super) struct PrefixMatch {
+    pub(super) tokens: Arc<[TokenIdType]>,
+    pub(super) prefix_len: usize,
+    deepest_boundary: usize,
+    // Absent only for callers of the existing public extension method.
+    deepest_hash: Option<Blake3Hash>,
```
jthomson04, line 57: **`Option<Blake3Hash>` makes the PR's entire win silently optional.**

A `None` here falls back to a full re-hash of the prefix with no assert, no counter, and no test. Any future `PrefixMatch` construction site — or a refactor of `longest_prefix_match_with_hash` — that leaves this as `None` reverts the measured 32% encode-CPU saving with zero signal: all 37 tests still pass, clippy stays clean, and the only symptom is production latency.

Suggested fix: `PrefixMatch` is module-private, so this can be a plain `Blake3Hash`, with the legacy `extend_after_match` wrapper keeping its own recompute path (or computing the digest itself before constructing the match). Then the optimized path cannot degrade unnoticed.

tokenizers/src/cache/l1.rs
```diff
@@ -355,6 +378,30 @@ impl L1Cache {
         deepest_boundary: usize,
         tokenizer: &E,
     ) -> anyhow::Result<Vec<TokenIdType>> {
+        self.extend_after_match_with_hash(
```
jthomson04, line 405: **Altitude: this adds two methods plus an optional-digest field on top of the existing API instead of changing the underlying one, leaving four methods where two would do.** (The compat wrapper starts at l1.rs:373.)

After this change, `longest_prefix_match` and `extend_after_match` have zero non-test callers anywhere in the repo — grep finds them only in this file's `mod tests`. Production goes exclusively through the `_with_hash` pair. The tuple API survives purely for external consumers of the published `dynamo-tokenizers` crate, and it costs a duplicated method pair, the `Option` flagged on `deepest_hash`, and the digest/input trust gap flagged at l1.rs:439.

Suggested fix: make `PrefixMatch` public and replace the tuple API (a semver-visible change release-plz can carry). That collapses four methods to two and removes both of those problems at the source.

tokenizers/src/cache/l1.rs
```diff
@@ -197,6 +206,11 @@ impl L1Cache {
     /// `deepest_boundary` is the deepest special-token boundary in `input` (end-exclusive),
     /// handed back so [`extend_after_match`] need not rescan the input for it.
     pub fn longest_prefix_match(&self, input: &str) -> Option<(Arc<[TokenIdType]>, usize, usize)> {
+        self.longest_prefix_match_with_hash(input)
+            .map(|matched| (matched.tokens, matched.prefix_len, matched.deepest_boundary))
+    }
+
+    pub(super) fn longest_prefix_match_with_hash(&self, input: &str) -> Option<PrefixMatch> {
```
jthomson04: **The two methods that now carry all the real logic have no doc comments, while the rationale docs stay on the vestigial wrappers.**

In a file where every other method has a multi-line "why" doc, `longest_prefix_match_with_hash` (here) and `extend_after_match_with_hash` (l1.rs:393) have none. Meanwhile the surviving doc on `longest_prefix_match` (l1.rs:203-207) still says the deepest boundary is "handed back so [`extend_after_match`] need not rescan", with no mention of the digest that is now the whole point — and `CachedTokenizer::extend_on_hit` (mod.rs:95) still links `L1Cache::extend_after_match`, which production no longer calls.

Net effect: a reader tracing the hot path lands on undocumented functions, with the docs pointing at the dead one.

Suggested fix: move the rationale docs onto the `_with_hash` methods, leave a one-line "compatibility wrapper" note on the public pair, and update the mod.rs:95 link.

jthomson04 (commented):
Review of the prefix-hash reuse change.

**Verified green** on `af99992` in a scratch worktree off `main`:
- `cargo fmt -p dynamo-tokenizers -- --check` — clean
- `cargo clippy -p dynamo-tokenizers --all-targets --locked` — clean, no warnings
- `cargo test -p dynamo-tokenizers --lib cache::` — 29 passed
- `cargo test -p dynamo-tokenizers --test cache_correctness` — 8 passed

**The core optimization is correct.** The digest carried out of `longest_prefix_match_with_hash` is the last entry of `prefix_hashes`, i.e. blake3 of `input[0..boundaries.last()]`, which is byte-identical to what `extend_after_match` used to recompute from `input.as_bytes()[..deepest]` (blake3 streaming and one-shot agree). And `deepest` in the extend path is `(deepest_boundary > prefix_len).then_some(deepest_boundary)`, so whenever an insert happens the reused digest corresponds to exactly the prefix being inserted. I found no hard correctness bug.

The 9 inline comments below are therefore **fragility, efficiency, and cleanup — not blockers**. The two worth a decision before merge are the first two: the digest is now accepted on trust with no structural or runtime link to the `input` it is keyed against (silent cache poisoning if that ever desyncs), and the miss path in `CachedTokenizer::encode` still performs the exact duplicated boundary scan + full blake3 pass that this PR removes from the hit path.

tokenizers/src/cache/l1.rs
```diff
@@ -853,6 +908,62 @@ mod tests {
         );
     }
 
+    #[test]
+    fn hash_reuse_without_deeper_boundary_does_not_insert() {
```
jthomson04: **This test never exercises hash reuse — both inputs take the early-return branch where `deepest_hash` is dropped unread.**

For `"<s>世界"` the boundaries are `[3]`, and for `"<s>世界</s>"` they are also `[3]` (the trailing `</s>` ends at `len` and is filtered out by `boundaries_with`). So in both cases `deepest_boundary == prefix_len`, `deepest` is `None`, and `extend_after_match_with_hash` returns before ever touching `hash_bytes`. The test would pass unchanged if `deepest_hash` were set to garbage.

It re-covers pre-existing no-insert behavior under a name that implies it guards the new code path; the reuse-specific assertions live only in `extend_after_match_persists_correct_deepest_entry`.

Suggested fix: either rename it to reflect what it actually covers, or add a case with a deeper boundary where the resulting insert key is asserted.

tokenizers/src/cache/l1.rs
```diff
@@ -798,24 +845,32 @@ mod tests {
         // reuses a correct prefix. Also proves the deepest-only invariant: extend
         // persists exactly one new entry.
         let tok = load_tokenizer();
-        let turns = growing_chat_turns(3);
+        let turns: Vec<_> = growing_chat_turns(3)
+            .into_iter()
+            .map(|t| t.replace("system", "system 世界 🦀"))
```
jthomson04: **Mutating the existing regression test's input rather than adding a case drops the ASCII coverage of the persisted-deepest-entry invariant.**

`extend_after_match_persists_correct_deepest_entry` is the only test that verifies the *saved* entry equals `encode(input[0..deepest])` and is retrievable by a later lookup. After this `.map(...)` it only ever runs against the mutated multibyte conversation, so a regression that manifests only on the plain ASCII prefix shape — the shape the integration tests and the rest of the l1 suite use — would no longer be caught here.

Suggested fix: parameterize over both variants (e.g. loop over `[turns, turns_multibyte]`) instead of replacing the input.

tokenizers/src/cache/l1.rs
```diff
@@ -223,6 +237,10 @@ impl L1Cache {
             last_pos = boundary_pos;
         }
 
+        // Reuse this digest if extension later inserts at the deepest boundary. It is
+        // not necessarily the digest of the shallower entry that satisfies the lookup.
+        let deepest_hash = prefix_hashes.last().expect("prefix hashes is non-empty").1;
```
jthomson04: **Simplification: the same "non-empty" fact is now `expect`-guarded twice in one function, and the deepest digest is materialized on paths that never use it.**

`deepest_boundary` comes from `boundaries.last().expect("boundaries is non-empty here")` at l1.rs:227, and `deepest_hash` from `prefix_hashes.last().expect("prefix hashes is non-empty")` here — two panic sites asserting the identical invariant, when `prefix_hashes.last()` alone yields both `(boundary_pos, hash)`.

The copy also runs before the search loop, so it executes on every full miss and on every hit where no deeper boundary exists, where the value is then discarded.

Suggested fix: take both values from a single `last()` after the loop, removing the redundant `expect` and the dead copy.

tokenizers/src/cache/l1.rs
jthomson04: Added the debug-only digest/input check in f07fc0f. Hit extension checks the deepest digest before insertion; miss population checks saved digests incrementally in debug builds. A negative test uses different inputs with matching boundary offsets and confirms that the incorrect entry is not inserted. This keeps the request-local types free of input lifetimes and adds no release validation hashing.

tokenizers/src/cache/mod.rs
jthomson04: Addressed in f07fc0f. The private lookup result now retains the existing boundary/hash vector on a miss. CachedTokenizer moves it into the shared population loop, avoiding both the second boundary scan and the second BLAKE3 pass. Public callers use the same loop with lazily computed hashes. Tests cover token/entry parity, accounting, and partial failures. Miss-path savings have not been benchmarked.

tokenizers/src/cache/mod.rs
jthomson04: Removed the unreachable empty-suffix branch in f07fc0f. Lookup boundaries remain strictly before the end of the input, so a successful prefix lookup always leaves a nonempty suffix.

tokenizers/src/cache/l1.rs
jthomson04: Keeping the optional digest for the existing public extension wrapper. It supplies None so hashing remains lazy and is skipped when there is no deeper boundary to insert. Internal lookup always supplies Some, and the saved-entry regression test explicitly checks the exact deepest digest; returning None there would fail that test. This preserves the public behavior with one extension implementation.

tokenizers/src/cache/l1.rs
jthomson04: Keeping the public signatures in this PR. L1Cache is part of a published crate, so the absence of in-repo production callers does not establish that changing the tuple API is safe for external callers. The private lookup result now supports both hit and miss reuse, while the public methods remain thin compatibility wrappers. A public API change is outside this patch.

tokenizers/src/cache/l1.rs
jthomson04: Added documentation to the internal lookup, population, and extension helpers in f07fc0f, including the same-input digest contract and the public-wrapper fallback. Updated the private extend_on_hit field link to the internal extension method. Retained the useful contracts on the public methods for external callers.

tokenizers/src/cache/l1.rs
jthomson04: Renamed the test to extend_without_deeper_boundary_does_not_insert in f07fc0f. The saved-entry test remains the check for actual digest reuse and insertion, so there is no duplicate insertion test.

tokenizers/src/cache/l1.rs
jthomson04: The saved-entry test now runs both the original ASCII conversation and the Unicode variant, with a fresh cache for each. Both cases check the deepest digest, one new entry, and exact tokens retrieved from the saved prefix. Included in f07fc0f.

tokenizers/src/cache/l1.rs
jthomson04: Addressed in f07fc0f. The deepest boundary and digest now come from one prefix_hashes.last() call after a cache hit is found. Full misses retain and move the existing vector to population, without extracting or copying a separate deepest digest.

rmccorm4 (approved):

jthomson04:
/ok to test [beb5103](https://github.com/ai-dynamo/frontend-crates/commit/beb51032b3e3a08d8dbb32f40325bb9d4dbbee8c)

rmccorm4:
/ok to test [7ff17f8](https://github.com/ai-dynamo/frontend-crates/pull/244/commits/7ff17f8986e95f3b87cbb27ffdd3decbdaaf5aaa)

rmccorm4:
/ok to test [b6a0794](https://github.com/ai-dynamo/frontend-crates/pull/244/commits/b6a0794b71c4dd7e3eb03443f7d07492f88452c1)

PR #246, opened 2026-09-19
refactor: migrate Unified fixtures to YAML history

keivenchang:
#### Overview:

This PR moves Unified conformance history from binary `tar.gz` shards to sparse YAML. It changes fixture storage and conformance tooling only; no parser or production Rust behavior changes.

Each family has one `inputs_and_golden.yaml` file. Each implementation/runtime version has its own `<implementation>-<runtime_version>.yaml` file with provenance, a parent capture, and only changed observations. An unchanged version records an empty delta, so results are not duplicated; materialization follows the parent chain.

#### Why this reduces conflicts:

Separate text files keep captures from different families and implementations out of one binary merge hotspot. A new capture changes only the affected family and implementation/version file. The manifest still pins the published snapshot, and the old archives remain as migration evidence for later duplicate cleanup.

#### Capturing and checking:

Run the documented one-command gate from the repository root:

```bash
bash conformance/utils/regenerate_unified.sh
```

It generates GOLDEN data, captures and packages the current results, extracts the snapshot, renders the JSON/HTML report, and runs the parser, schema, parity, render, Python, archive, and empty/red-cell checks.

#### Changing cases or columns:

- Add or change a case in the family’s authored input/golden source, then recapture the affected current columns.
- Add a new implementation/runtime by capturing it and adding `<implementation>-<runtime_version>.yaml`; the filename identifies the column and the file records provenance and its parent.
- Recapture affected columns when the request, tool schema, chunk schedule, or expected result changes. Do not reuse old output for a changed request based only on the case ID.

Where should the reviewer start?

`conformance/utils/src/unified_history.py`, `conformance/utils/src/package_fixtures.py`, `conformance/fixtures/unified/README.md`, and one family under `conformance/fixtures-unified-v2/families/`

/coderabbit profile chill

Related: [DIS-2944](https://linear.app/nvidia/issue/DIS-2944)


rmccorm4 (commented):
It looks like each of these large 30k+ yamls stores ~10 previous versions of captures. Can we keep it simple to N-1/N-2 captures, and just rely on git history if you want look back and compare any further? This should significantly reduce the amount of lines we need to store/update/etc. - and give us a simple policy to follow going forward. Let me know what you think.

keivenchang:
> It looks like each of these large 30k+ yamls stores ~10 previous versions of captures. Can we keep it simple to N-1/N-2 captures, and just rely on git history if you want look back and compare any further? This should significantly reduce the amount of lines we need to store/update/etc. - and give us a simple policy to follow going forward. Let me know what you think.

Thought about that, unfortunately, we are constantly adding new cols and have to backfill old captures. Backfilling needs the retained capture and provenance.

Also, if a version changes and an output does not, we just add a new capture node with a parent pointer. The unchanged output does not get copied again.

But going forward, the YAML should be mostly additive from here, so I would rather improve its physical layout if parsing becomes an issue than drop the history.

rmccorm4 (approved):

PR #247, opened 2026-09-19
fix(parsers): preserve DeepSeek V4 string whitespace

michaelfeil:
## Summary

Preserve the exact value of DeepSeek V4 DSML parameters marked `string="true"`.
The current decoder trims every parameter before inspecting its type, silently
removing indentation, trailing newlines, and whitespace-only string values.

Only move trimming into the non-string branch. JSON parsing and malformed
non-string fallback behavior remain unchanged. No routing, request-policy,
tokenization, dependency, or performance change is claimed.

## Reproducer

```text
<｜DSML｜tool_calls><｜DSML｜invoke name="edit"><｜DSML｜parameter name="text" string="true">  café
</｜DSML｜parameter><｜DSML｜parameter name="count" string="false"> 42 </｜DSML｜parameter></｜DSML｜invoke></｜DSML｜tool_calls>
```

Expected arguments: `{"text":"  café\n","count":42}`.
Before this fix: `{"text":"café","count":42}`.

The Rust regression exercises the public stream parser with boundary whitespace,
whitespace-only text, an empty string, and literal `"null"`. Each case includes
a whitespace-padded numeric parameter to protect existing non-string behavior.
The regression failed on the unchanged base and passed after the fix.

## Peer evidence

Current [vLLM's DSML converter](https://github.com/vllm-project/vllm/blob/729ebac4983e1510e035bc579142b0fc210a49a3/vllm/parser/deepseek_v4.py)
preserves string values. Added the equivalent parameterized test locally to its
existing `tests/parser/engine/test_deepseek_v4.py::TestArgConverter` suite and
executed it against that pinned source: **4 passed**, 103 deselected.

```python
@pytest.mark.parametrize("value", ["  café\n", "\t\r\n ", "", "null"])
def test_string_preserves_boundary_whitespace(self, value):
    raw = self._raw(("text", "true", value), ("count", "false", " 42 "))
    result = json.loads(_dsml_arg_converter(raw, partial=False))
    assert result == {"text": value, "count": 42}
```

The peer run uses the repository's test module and parser fixtures with
`--confcutdir` at `tests/parser/engine` to exclude unrelated GPU-serving fixtures.
It is not a vLLM full-suite or model-inference claim. The peer test is local
comparison evidence, not a separately submitted vLLM PR. Cases are authored
synthetic examples, not captured customer requests.

## Validation

- Rust regression failed on the unchanged base, then passed after the fix.
- `cargo test -p dynamo-parsers-v2 --lib`: **634 passed**.
- Five integration binaries (`gemma4_guided`, `gemma4_native_recovery`,
  `preserve_special_tokens_parity`, `vendor_parser_example`, `vendor_registry`):
  **28 passed**.
- `cargo clippy -p dynamo-parsers-v2 --all-targets -- -D warnings`: passed.
- `cargo fmt --all --check` and `git diff --check`: passed.
- Equivalent vLLM repository-native parameterized regression: **4 passed**.
- Canonical conformance report rendered. This uses existing recorded snapshots,
  not a fresh model capture; generated artifacts are not committed.

Rust checks ran in plain Docker with repository-pinned Rust 1.96.1.


rmccorm4:
/ok to test [d4e4c8c](https://github.com/ai-dynamo/frontend-crates/pull/247/commits/d4e4c8c4259de40e1bff27cf8fbe4220d9b7ed7d)

rmccorm4 (approved):

michaelfeil:
Actually let my codex write a test here. This test does not exist, and vllm/sglang just happen to have this behaviour, while this one did not. 

Provenance clarification: `test_string_preserves_boundary_whitespace` is a **locally authored test**, not an existing upstream vLLM test. It extends the existing [`TestArgConverter`](https://github.com/vllm-project/vllm/blob/729ebac4983e1510e035bc579142b0fc210a49a3/tests/parser/engine/test_deepseek_v4.py) helper:

```python
@pytest.mark.parametrize("value", ["  café\n", "\t\r\n ", "", "null"])
def test_string_preserves_boundary_whitespace(self, value):
    raw = self._raw(("text", "true", value), ("count", "false", " 42 "))
    result = json.loads(_dsml_arg_converter(raw, partial=False))
    assert result == {"text": value, "count": 42}
```

Re-ran against the unmodified pinned vLLM parser: **4 passed**. Its [converter](https://github.com/vllm-project/vllm/blob/729ebac4983e1510e035bc579142b0fc210a49a3/vllm/parser/deepseek_v4.py#L89-L99) directly assigns `value` when `string="true"`; it does not trim it. These are synthetic regression cases, not recorded model outputs or an upstream vLLM PR.

PR #251, opened 2026-09-19
fix(parsers): preserve non-nullable Qwen string arguments containing null

michaelfeil:
## Summary

Preserve the literal argument `null` as a string when the XML parameter schema
admits strings but excludes null. Uses the existing schema-type lookup and retains explicit nullable behavior.
No legacy-v1 change.

For a Qwen3-Coder tool with `location: {type: string}`:

```text
<tool_call><function=get_weather><parameter=location>null</parameter></function></tool_call>
```

Expected: `{"location":"null"}`.

Before this fix, one whole update produces `{"location":null}`. Small updates
start emitting a string, but final coercion changes its type. The consistency
guard then suppresses completion, leaving incomplete argument JSON and no
assembled call. This is an argument-fidelity and chunk-invariance bug, not just
a different recovery policy.

OpenAPI `type: string, nullable: true` had the same incomplete-stream failure.
It now uses the existing buffered path for null-admitting unions instead of
committing early to a string. Its final JSON-null policy is unchanged; ordinary
non-nullable strings retain early streaming.

Sibling type constraints now intersect with the alternatives from
`anyOf`/`oneOf`. For example, `type: string` beside `anyOf: [string, null]`
excludes null and must emit the string `"null"` consistently across chunks.
The existing regression also covers these compositions and an unconstrained
branch. This remains type-hint handling, not full JSON Schema validation:
no `allOf`, `$ref`, `const`/`enum`, or value-constraint resolution is added.

The composed-schema case above was supplied in review, not taken from an
existing vLLM test or production capture. A JSON Schema validator accepts the
string `"null"` and rejects JSON null for it. vLLM also emits JSON null for this
case; this follow-up is not a claim of peer parity. Dynamo additionally lost
the completed call with one-character updates before the fix.

## Existing peer regression

Found by replaying vLLM's existing
[`TestSchemaCoercionBoolNumberNull.test_null_stays_string_without_null_schema`](https://github.com/vllm-project/vllm/blob/729ebac4983e1510e035bc579142b0fc210a49a3/tests/parser/engine/test_qwen3.py#L947)
against the real Dynamo binding. That test passes unchanged in vLLM; the full
vLLM file passes all 51 tests. The native replay uses an API-shape adapter,
unchanged test inputs/assertions, and actual native parsing, not a parser mock.

Added one native regression covering whole-input and one-character updates,
non-nullable string alternatives, nullable type arrays and OpenAPI `nullable`.
The added OpenAPI case also failed before its stream-eligibility fix. The test
fails on the unchanged base (no completed call) and passes after the fix.

## Validation

- 121 Qwen-focused tests passed.
- Strict all-target Clippy passed.
- Full Rust suite: **634 library tests, 1 binary test and 28 integration tests passed**.
- Canonical conformance report rendered from existing recorded snapshots;
  generated files are not committed and this is not a fresh model capture.
- Plain Docker, Rust 1.96.1. No performance improvement claimed.


rmccorm4:
/ok to test [04d313f](https://github.com/ai-dynamo/frontend-crates/pull/251/commits/04d313f23a8fdbdd207340f71e36f4f511247efa)

rmccorm4 (approved):

michaelfeil:
ok, i think implementation and pr look fine to me

michaelfeil:
Addressed the composed-schema completion gap in 5cd3ab8. Sibling `type` constraints now intersect with existing `anyOf`/`oneOf` type hints; no new schema keywords or general validator were added (12 net production lines, 28 test lines).

For `{"type":"string","anyOf":[{"type":"string"},{"type":"null"}]}`, literal `null` now completes as the string `"null"` with both whole-input and one-character updates. The regression failed before the change. A small numeric-intersection assertion protects integer/number compatibility.

Provenance: this case was supplied in review, not copied from an existing upstream test or production capture. JSON Schema validation excludes JSON null here, but vLLM also emits null for this case, so this follow-up is **not** a vLLM-parity claim. Dynamo additionally lost completion under small chunks.

Validation: all 663 Rust tests passed; formatting and strict all-target Clippy passed in plain Docker. Canonical conformance report regenerated from existing snapshots, not new model captures. No performance claim.

parsers/v2/src/tool_calling/v1core/xml/parser.rs
```diff
@@ -774,16 +779,28 @@ fn collect_allowed_types_into(schema: &Value, out: &mut HashSet<SchemaType>) {
...
+    let mut constraints = Vec::new();
+    if !out.is_empty() {
+        constraints.push(out);
+    }
     for key in ["anyOf", "oneOf"] {
         if let Some(options) = schema.get(key).and_then(Value::as_array) {
-            for option in options {
-                collect_allowed_types_into(option, out);
+            let branches = options.iter().map(collect_type_constraints);
+            if let Some(alternatives) = branches.collect::<Option<Vec<_>>>() {
```
rmccorm4, line 801: **[P2] Preserve typed alternatives when a sibling uses `const`/`enum` without `type`**

`branches.collect::<Option<Vec<_>>>()` discards the entire `anyOf`/`oneOf` constraint when any branch lacks an explicit `type`, but `const`/`enum` branches can still constrain valid values. For example:

```json
{"anyOf": [{"const": "auto"}, {"type": "integer"}]}
```

A Qwen argument containing `42` previously produced `{"x":42}`; this change produces `{"x":"42"}`, which satisfies neither alternative. The same regression occurs with `oneOf: [{"enum":["auto"]},{"type":"integer"}]`, with both whole-input and single-byte chunking. Please preserve valid typed alternatives or account for `const`/`enum` constraints instead of treating these branches as unconstrained.

michaelfeil:
@codex review, addressed the enum case.

parsers/v2/src/tool_calling/v1core/xml/parser.rs
michaelfeil: ok, taking a look

parsers/v2/src/tool_calling/v1core/xml/parser.rs
```diff
@@ -774,16 +779,34 @@ fn collect_allowed_types_into(schema: &Value, out: &mut HashSet<SchemaType>) {
...
+    }
+    let mut constraints = Vec::new();
+    if !out.is_empty() {
+        constraints.push(out);
+    }
+    if let Some(value) = schema.get("const") {
+        constraints.push(HashSet::from([value_category(value)]));
+    }
+    if let Some(values) = schema.get("enum").and_then(Value::as_array) {
+        constraints.push(values.iter().map(value_category).collect());
```
rmccorm4, line 796: **[P2] Preserve integer-valued `const`/`enum` numbers**

The new literal constraints use `value_category`, which classifies `42.0` as `Number`. For a parameter schema such as:

```json
{"type": ["integer", "null"], "enum": [42.0]}
```

Intersecting `{Number}` with `{Integer, Null}` leaves no allowed types, so a Qwen argument containing `42` now produces `{"x":"42"}` instead of the parent's `{"x":42}`. The string violates the schema; [JSON Schema considers numbers with a zero fractional part integers](https://json-schema.org/understanding-json-schema/reference/numeric).

Also reproduced with `const: 42.0`, using both whole-input and single-byte chunking, against parent `af920fb`. Please account for integer-valued numeric literals before intersecting the `const`/`enum` type constraints.

michaelfeil:
@rmccorm4 this might be able to get merged / ok to test.

parsers/v2/src/tool_calling/v1core/xml/parser.rs
```diff
@@ -774,25 +776,43 @@ fn collect_allowed_types_into(schema: &Value, out: &mut HashSet<SchemaType>) {
...
```
rmccorm4: **Preserve decimal/exponent arguments when literal constraints narrow to integer**

For a parameter schema such as:

```json
{"type": ["number", "null"], "const": 42.0}
```

Model output `<parameter=x>42.0</parameter>` now produces `{"x":"42.0"}` instead of the base's `{"x":42}`. The literal constraint narrows the allowed types to `Integer`, but `coerce_union_value` only accepts digit-form integer literals in that branch, so valid integral decimal/exponent spellings fall through to strings.

Also reproduced with raw `4.2e1`, `const:42`, and `enum:[42]` / `enum:[42.0]`, in both Qwen3-Coder and MiniMax-M2, with whole-input and character streaming. The same 72-case probe passes on base `8cd55a1` and fails 48 cases on head `aff7f52`.

Please align integer coercion with the semantic integer classification and add coverage using a decimal/exponent-form **raw argument**, not only a floating-point schema literal. These inputs already satisfy the schema; this does not require general value-constraint validation.

rmccorm4: This is describing the 42.0 / 4.2e1 rows below:

| Model-emitted value | Dynamo before #251 | Dynamo after #251 | vLLM | SGLang |
| ------------------- | ------------------ | ----------------- | ---- | ------ |
| `42`                | `42`               | `42`              | `42` | `42`   |
| `42.0`              | `42`               | `"42.0"`          | `42` | `42.0` |
| `4.2e1`             | `42`               | `"4.2e1"`         | `42` | `42.0` |

keivenchang:
Could you rebase this PR onto the latest main when you get a chance? Thanks.

keivenchang:
@michaelfeil, is there any way I can push directly to michaelfeil:fix/qwen-string-null for things like rebases? It would speed up the back-and-forth.

PR #250, opened 2026-09-19
fix(parsers): accept DeepSeek V4.1 JSON invocation bodies

michaelfeil:
## Summary

Accept a JSON object inside DeepSeek V4.1's spaced DSML invocation, alongside
the existing XML-parameter body. This is a concrete interoperability gap:
[SGLang's V4.1 structural-tag builder](https://github.com/sgl-project/sglang/blob/10b0bcfd18161d4c57bb1098257265e2df6b48cd/python/sglang/srt/function_call/deepseekv41_detector.py)
deliberately generates JSON bodies because XGrammar's `deepseek_xml` style
hardcodes unspaced parameter names. Its detector accepts that output, but
Dynamo currently raises `invalid DeepSeek V4.1 parameter header`.

```text
<｜DSML｜ calls>
<｜DSML｜ invoke name="inspect">{"value":"café 🐈"}</｜DSML｜ invoke>
</｜DSML｜ calls>
```

The invocation boundary reuses the existing incremental JSON-prefix state so
literal DSML closing markers inside JSON strings remain data. Only newly
appended JSON characters are scanned. The emitter validates the complete object
with serde_json. Existing XML parsing and guided-output routing remain intact.

## Evidence and scope

- A public native-stream regression failed on the unchanged base and passes
  with this change. It covers nested values, Unicode, escaped quotes, literal
  closing markers, and consecutive calls at every valid UTF-8 split.
- Existing malformed/EOF tests cover malformed JSON, trailing junk, and
  incomplete bodies; an incremental-work test guards against rescanning a
  16-KiB string with one-character updates.
- A local repository-native SGLang test compiled the actual strict-tool
  structural tag through XGrammar, accepted the text and EOS, and recovered
  the original arguments with 1-character, 7-character and whole-input updates:
  **1 test, 3 subtests passed**. SGLang can release an inter-tag newline as
  normal text at fine partitions; this evidence asserts argument correctness,
  not identical whitespace emission.
- This is authored grammar-valid data, not a captured model generation. It
  adds interoperability with that existing serving grammar, not a claim about
  all model-produced V4.1 output. No performance improvement is claimed.

The SGLang run used its real detector/protocol/helpers, bypassing only the root
package initializer's unrelated GPU imports. No SGLang PR was submitted.
AI assistance was used to investigate and implement this change.

## Validation

- 17 focused DeepSeek V4.1 tests passed.
- Strict all-target Clippy passed; formatting and diff checks passed.
[...]

PR #252, opened 2026-09-19
feat(parsers): expose DeepSeek V4.1 through the tool-only API

michaelfeil:
## Summary

Register `deepseek_v41` with the tool-only factory by projecting its existing
UnifiedParser events. There is no new scanner or model-specific state machine.

This lets callers use the same tool-parser selector as other families while
keeping their own reasoning parser. Initialization uses Response mode so
`<think>`/`</think>` remain ordinary text, including when the prompt supplied
the opening reasoning marker. Native tool framing is still parsed normally.

No guided-decoding builder or Python integration change. The independent JSON
invocation-body fix in #250 is not included; this adapter inherits the existing
unified parser's behavior and future fixes.

## Evidence

The installed binding currently accepts V4.1 through UnifiedParserStream but
rejects `ToolCallStream("deepseek_v41")`. A boundary comparison using vLLM's
existing [V4.1 parallel-call fixture](https://github.com/vllm-project/vllm/blob/729ebac4983e1510e035bc579142b0fc210a49a3/tests/parser/engine/test_deepseek_v41.py)
passed 36/36 replays through Baseten Python and native Response-mode parsing:
three reasoning/plain-text prefixes and six update sizes. Exact text and typed
arguments matched. This is text-parser evidence, not live model serving.

Two native integration tests cover the public factory, exact reasoning-marker
preservation, parallel calls with Unicode and numeric values, every UTF-8 split,
empty updates, and finalization. Both failed before registration and pass with
the adapter. Parser-family correctness remains in the existing unified tests.

## Validation

- Focused integration tests and strict all-target Clippy passed.
- Full Rust suite: 633 library tests, one binary test, and 30 integration tests passed.
- Canonical report rendered from existing recorded snapshots; generated files
  are not committed. This is not a new family conversion or fresh model capture.
- Plain Docker; no performance improvement claimed.


rmccorm4:
/ok to test [c9ed381](https://github.com/ai-dynamo/frontend-crates/pull/252/commits/c9ed381622ca18da7ef8880199e03ce1a3795476)

parsers/v2/src/tool_calling/deepseek_v41.rs
rmccorm4, line 1: > This lets callers use the same tool-parser selector as other families while
keeping their own reasoning parser.

The change LGTM, but wanted to pick your brain on why you may use a different reasoning parser in this case. 

1. Are there open issues that need fixing with the current reasoning parser?
2. How are you currently introducing the reasoning parser? Through a maintained patch, or some kind of plugin style?

michaelfeil: Thanks! This is an incremental-adoption boundary, rather than a workaround for a known defect in the V4.1 unified reasoning parser.

1. We have not identified a V4.1 reasoning bug motivating this PR. Our text-path replay passes the 14 applicable cases from the existing vLLM V4.1 suite; token-ID accounting tests are outside that comparison. We want to replace tool extraction first without simultaneously changing reasoning policy, prefilled reasoning handling, or answer-only stop placement.
2. Reasoning currently stays in our existing Python frontend, outside frontend-crates. It is not a patched Rust reasoning parser or a plugin injected into this crate. Python calls the native tool parser, then passes the remaining text through its existing reasoning parser. The tool-only projection therefore preserves the reasoning delimiters instead of consuming them.

The implementation still uses the existing V4.1 UnifiedParser underneath, initialized in Response mode; there is no second tool-parsing state machine. Moving to the full unified interface can be a separate migration once the surrounding response contracts are checked.

michaelfeil: Sorry, codex went a bit wild here!

michaelfeil:
We need to support the V4.1 wire format, but we do **not** require a `deepseek_v41` parser name or separate tool-only registry entry. Matching SGLang/vLLM's naming is not a requirement for this API.

There is a real syntax difference: DeepSeek's [reference encoder](https://huggingface.co/deepseek-ai/DeepSeek-V4.1-Flash/blob/dba1be0a40aa45a94ad051997016db3960a90277/encoding/encoding.py) uses `<｜DSML｜ calls>` and spaced `invoke`/`parameter` tags, while V4 uses `<｜DSML｜tool_calls>` and unspaced tags. I replayed the official V4.1 fixture through the existing V4 parser across 263 chunkings: it produced zero tool calls and returned the markup as normal text. So the existing implementation needs to recognize the additional syntax; simply reusing it unchanged would not work.

Could we revise this approach to extend the existing DeepSeek/DSML parser entry point to support both dialects, instead of adding another named parser and wrapper?

Suggested approach:

- Factor structural markers into a shared DSML dialect configuration and reuse the boundary/emitter machinery where the semantics agree.
- Recognize the V4 and V4.1 structural openers through the existing entry point, retaining the appropriate dialect while parsing each invocation, including supported bare-invocation recovery.
- Preserve argument contents exactly. Do not normalize the input with global string replacements: marker-like strings inside argument values must remain data.
- Reuse the existing tool-only projection and coordinate reasoning-marker preservation with #253. V4.1's existing UnifiedParser implementation can supply behavior and tests to consolidate; another projection class is unnecessary.
- Test both wire formats through the **same existing selector**, including whole-input and split updates, parallel calls, literal markers in strings, reasoning delimiters, and truncated input. Include JSON-body support from #250 when consolidating that path.

If the broader consolidation belongs in a follow-up, we can defer this PR: V4.1 is already available through UnifiedParser. The missing `ToolCallStream("deepseek_v41")` selector alone is not a feature gap we need to fill.

The current registration also causes the conformance sweep to fail because the new family has zero stream fixtures. Cover the added dialect through the existing entry point's conformance cases rather than adding a registry name merely to match other projects.

michaelfeil:
Closing in favor of a replacement that supports both V4 and V4.1 DSML dialects through the existing DeepSeek tool-only entry point, using a shared adapter rather than introducing a separate deepseek_v41 selector. The replacement should preserve argument contents and reasoning delimiters and cover both dialects in streaming and conformance tests.

michaelfeil:
Replacement: #255 supports both DSML dialects through the existing deepseek_v4 tool-only selector, sharing one scanner and delegating to the existing dialect boundary parsers and decoders. No deepseek_v41 tool-only registration is added. Both native golden corpora and the stream split sweep pass.

parsers/v2/src/tool_calling/deepseek_v41.rs
michaelfeil: actually was checking. maybe there is no need to have a separate parser for the deepseev v4.1 archiecture altogether. [https://github.com/ai-dynamo/frontend-crates/pull/255)

For reasoning parser, we still use a legacy self rolled parser at the moment.

PR #253, opened 2026-09-19
fix: preserve reasoning markers in DeepSeek V4 tool projection

michaelfeil:
## Summary

Keep `<think>` / `</think>` in `DeepSeekV4ToolStreamParser`'s normal text so a
caller can perform its own reasoning extraction. Initialize the existing
UnifiedParser in Response mode; no new scanner or change to the unified factory.

Before this change, tool-only input `reason</think>\n\n` returns `reason\n\n`.
The caller cannot tell where reasoning ended. A recorded frontend replay exposed
this as answer whitespace being attributed to reasoning and a changed reasoning
token count. The rebuilt binding restores the original split and count.

The old projection test only compared with the same reasoning-enabled unified
implementation. Replace it with explicit expected text and tool assertions for
both a complete reasoning block and a prompt-prefilled reasoning close, at every
UTF-8 split. It fails before the fix and passes after it.

This deliberately changes legacy tool-only projection semantics. Callers that
want reasoning extracted should use UnifiedParser; ToolParseResult has no
reasoning channel. Compact argument JSON and argument streaming behavior are
unchanged.

## Verification

- Fail-before/pass-after native regression.
- Formatting and strict all-target Clippy pass.
- Real rebuilt Python binding preserves the marker.
- Unchanged recorded HTTP/SSE cases restore content/reasoning attribution and
  reasoning usage. Five exact-output failures remain from compact JSON versus
  the Python parser's spaced serialization; no expectations were rewritten.
- Full crate suite: 633 library + 1 binary + 28 integration tests pass (662 total).
- Required canonical report rendered from existing snapshots; not a fresh capture.


rmccorm4:
/ok to test [2e997ea](https://github.com/ai-dynamo/frontend-crates/pull/253/commits/2e997ead127f7ccb5488bef5a4d69b53e84b039a)

rmccorm4 (approved):
Seems reasonable to me - should only affect when tool parser is used in isolation, and not alter the behavior of unified parser when both tool and reasoning parser are used in combination

michaelfeil:
@rmccorm4 Fully agree, not sure if this PR helps a lot. Maybe if reasoning parser not activated, but leaking the <think> or end of think tag used to be a common issue of inference providers. Ok with not merging, but also no regression.

PR #255, opened 2026-09-19
feat(parsers): support both DSML dialects in DeepSeek tool parser

michaelfeil:
The existing `deepseek_v4` tool-only selector misses DeepSeek V4.1's spaced DSML markers. Accept both dialects through the same scanner, delegating invocation boundaries and argument decoding to the existing dialect implementations. No new tool-only selector or parser state machine is introduced, and argument text is not rewritten.

The adapter supports wrapped and bare invocations, including switching dialects between calls. Explicit unified selectors retain their existing wiring. Reasoning behavior is unchanged: #253 remains independent, so this PR does not need to be stacked on it. V4.1 JSON invocation bodies remain a separate decoder change in #250.

This replaces #252. The simplified diff is 185 additions / 42 deletions across five files, including regression tests.

Validation on this revision:
- DSML unit tests: 14 passed, including mixed dialects, bare calls, literal markers/entities, Unicode, every UTF-8 split, and character-by-character input.
- Unified parity tests: 14 passed, including 84 native DeepSeek golden cases across five chunking strategies through the existing tool selector.
- Tool-calling streaming sweep: 9 passed.
- Strict Clippy for both parser and conformance crates, all targets: passed.
- Formatting and diff whitespace checks: passed.

The canonical conformance report was refreshed from existing snapshots; this is not a fresh live capture, and does not establish that the broader conformance table is green.

AI-assisted implementation and validation.


keivenchang:
Hey Michael — could you rebase PR #255 onto current `main`? I tested PR #255 locally by applying its two commits on top of PR #257's merged result. The rebase completed cleanly because #257 changed the conformance fixture/history infrastructure, while #255 only touches the DeepSeek parser, its README, and the parser parity test, so there is no overlapping file-level change to resolve.

I also rebuilt the workspace and ran the same CI checks locally: `cargo fmt --all -- --check`, `cargo machete`, `cargo deny --all-features check bans licenses`, the locked workspace check/tests/doctests/clippy/publish dry run, and `PATH="/tmp/pvenv-pr255/bin:$PATH" conformance/utils/check.sh ci`. They passed locally, including the conformance render with `0` red cells and `519 passed, 2 skipped` Python tests. Once the branch is rebased and pushed, I can submit it to CI again.

rmccorm4 (approved):

furionw:
@keivenchang Rebased onto current `main` (`26095c9dc9d673cef2fa12af899a5ea367c9d1f5`) as requested in https://github.com/ai-dynamo/frontend-crates/pull/255#issuecomment-5769782782. The new head is `028ef880457fba1c10797ba3048a8dd671174685` and is 0 commits behind; I am submitting this exact SHA to CI now.

